Issue #6, opened 2015-08-04
Is it possible to access original props after applying decorator?


danmartinez101:
I'm wondering if I'm doing something wrong with this decorator; when I apply it to a component I can no longer access things like actions passed via props. 

In the beginning of the render method, I logged `this.prop` and can see the actions that I passed, but as soon as I add the decorator the only props available are the ones from redux-form. Am I making a mistake here?

PR #7, opened 2015-08-04
include original props on when rendering decorated component

danmartinez101:
I'm not sure if the removal of the original props was an intentional design decision, but in case it wasn't: This small change "fixed" #6 for me so that I can now access actions passed as props from a container component.


erikras:
Thanks. I had just fixed this locally a slightly different way.

PR #8, opened 2015-08-04
Fixed arguments documentation

johanneslumpe:
Just two two tiny fixes.


erikras:
Thanks! I did that api change hastily. :+1:

PR #9, opened 2015-08-04
No need for a 2nd destructuring and `delete`

johanneslumpe:
Since we're already using destructuring here, we can skip the 2nd destructuring step and the `delete` (in order to avoid possible deoptimization of the function)


erikras:
Very nice!!

johanneslumpe:
argh

johanneslumpe:
it has a typo

PR #13, opened 2015-08-07
Use separate blur method to set visited status

whodidthis:
https://github.com/erikras/react-redux-universal-hot-example/issues/101

Adding a separate blur action for marking `visited` to make validation a bit more friendly

### DIFF
--- a/src/actionTypes.js
+++ b/src/actionTypes.js
@@ -1,3 +1,4 @@
 export const CHANGE = 'REDUX_FORM_CHANGE';
+export const BLUR = 'REDUX_FORM_BLUR';
 export const SHOW_ALL = 'REDUX_FORM_SHOW_ALL';
 export const RESET = 'REDUX_FORM_RESET';
--- a/src/actions.js
+++ b/src/actions.js
@@ -1,4 +1,4 @@
-import { CHANGE, SHOW_ALL, RESET } from './actionTypes';
+import { CHANGE, BLUR, SHOW_ALL, RESET } from './actionTypes';
 
 export function change(form, field, value) {
   return {
@@ -6,6 +6,14 @@ export function change(form, field, value) {
     form: form,
     field: field,
     value: value
+  };
+}
+
+export function blur(form, field) {
+  return {
+    type: BLUR,
+    form: form,
+    field: field
   };
 }
 
--- a/src/createFormReducer.js
+++ b/src/createFormReducer.js
@@ -1,4 +1,4 @@
-import { CHANGE, SHOW_ALL, RESET } from './actionTypes';
+import { CHANGE, BLUR, SHOW_ALL, RESET } from './actionTypes';
 
 /**
  * Creates a state structure like:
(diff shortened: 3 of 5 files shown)


erikras:
I considered this, and even began implementing it, but later decided against it. It's a matter of taste, I guess. I prefer the immediate reaction of the validation. 

And, unless I'm mistaken, they can't really coexist that well. Either you mark visited on change or not.

Let me think about it a bit more. And see if you can think of an elegant way for onlyMarkVisitedOnBlur to be an option.

whodidthis:
I think it could make sense to add a `changed` tree in addition to `visited`.

So in react-redux-universal-hot-example's case recreating the immediate behavior would be
```js
emailError && (emailVisited || emailChanged) && <div className="text-danger">{emailError}</div>
```

AngularJS has good support for really pinpointing how to display validation (taken from https://docs.angularjs.org/guide/forms):
> - ng-valid: the model is valid
- ng-invalid: the model is invalid
- ng-valid-[key]: for each valid key added by $setValidity
- ng-invalid-[key]: for each invalid key added by $setValidity
- ng-pristine: the control hasn't been interacted with yet
- ng-dirty: the control has been interacted with
- ng-touched: the control has been blurred
- ng-untouched: the control hasn't been blurred
- ng-pending: any $asyncValidators are unfulfilled

erikras:
How can something "change" without being "visited"?

I think you might be onto something with the idea of copying those flags from angular.

whodidthis:
Yeah, might be better to use similar flags with angular. I at least saw a couple of react form libs using `pristine/dirty`.

I could update pull request changing `visited` to `touched` and instead of `changed` adding `dirty`, if you want.

dzannotti:
is there a way to have validation called only on submit or pre-submit or something like that?

erikras:
Many of the ideas discussed here just got implemented in the v0.0.7 release. @whodidthis, I'd appreciate your feedback. This discussion was a big influence.

whodidthis:
Awesome :sunglasses:

ericclemmons:
Wandered on this thread from Reactiflux & I wanted to point out that, in the case of autocomplete, you'll definitely have lots of change events that the user may never have touched (or even seen, if it's hidden).

erikras:
Having thought about it for only 30 seconds after reading your comment, @ericclemmons, I would think that autocomplete would be handled by its own component, which would use its own reducer, and only call `handleChange` when a value was selected.

ericclemmons:
@erikras It's...quite a bit more complex than that :)  Unfortunately, the browser's autocomplete doesn't have an event to listen to.

Similar to the "autosuggest" on phones, all you'll get is an `onChange` fired all over the place :)

Enjoy! Muwhahahahahahahahaha

whodidthis:
Probably just means don't disable submit buttons because of any validation errors and remember to `touchAll` on submit and you should be good to go? Or at least to the extent easily possible.

https://github.com/erikras/redux-form#-touchall--function

PR #24, opened 2015-08-13
minor fix for angle brackets

apucacao:


erikras:
Ha! Putting it in back quotes must have broken that. Thanks! 👍🏻

Issue #23, opened 2015-08-13
How to dynamically add input fields


haohcraft:
Say I have a job application form and there are fields for your previous working experience.

Since you have multiple jobs before, I need a way to dynamically add those field.

Any suggestion for this case? Thanks.

erikras:
So your "field" (property on an object in your store) is an Array of strings (or objects of each job has more than one field). You'd have to have some sort of component that called `handleChange('jobs')(modifiedArray)`. Keeping each of those inputs in redux might be tough, but perhaps the component that did the addition could also use `reduxForm('newJobForm')` or the like. Does that help?

haohcraft:
>  perhaps the component that did the addition could also use reduxForm('newJobForm')

In this case, the newly added *newJobForm* will use the same sliceName `newJobForm` as the existing one does. Does this mean that the two job forms would reflect the same data within `this.state.newJobForm` ?

haohcraft:
All right, wrote a higher lever component: 
```
/**
*
* This Component is to handle Nested inputs
*
**/

"use strict";
import BaseComponent from'BaseComponent';
import React from 'react/addons';
import cx from 'classnames';
import _ from 'lodash';


export default class NestedInput extends BaseComponent {
	
	static propTypes = {
        customClasses: React.PropTypes.string,
        onChangeHandler: React.PropTypes.func.isRequired,
        optionsType: React.PropTypes.string.isRequired,
        data: React.PropTypes.array,
        label: React.PropTypes.string,
        parseFn: React.PropTypes.func
    }

	constructor(){
        super();
        this._bind('_onChangeHandler');
    }


    render() {
    	const {customClasses, onChangeHandler, optionsType, data, label} = this.props;
    	let classnames = cx('NestedInput', customClasses);
        let _this = this;
        let _data = data ? data : [];
    	return (
    		<div className={classnames}>
                <label className="label-control">{label}</label>
    			{React.Children.map(this.props.children, function(child, index) {
            		return React.cloneElement(child, {
            			data: _data[index] || {},
                        handleChange: _this._onChangeHandler(index),
                        handleBlur: _this._onChangeHandler(index)
                    });
            	})}
    		</div>
    	);
    }

    _onChangeHandler(index) {
    	return (field) => (e) => {
    		const {data, onChangeHandler} = this.props;
	        let _data = data ? data : [];
	        let _index = index;
	        let _newData = {};
	        _newData[field] = e.target.value;

[...]

PR #25, opened 2015-08-14
make `handleChange` and `handleBlur` 's input value more flexible

haohcraft:
A use case 

https://github.com/erikras/redux-form/issues/23


erikras:
Good call. 👍🏻

PR #26, opened 2015-08-17
Properly detect async validation result

johanneslumpe:
In case we don't have an async validation function `asyncErrors.valid` will be `undefined`, and `valid` will never be `true`. This fixes that by checking if we actually have a value for `asyncErrors.valid`.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -24,7 +24,8 @@ export default function reduxForm(sliceName, ...args) {
   function runValidation(form) {
     const syncErrors = validate(form.data);
     const asyncErrors = {...form.asyncErrors};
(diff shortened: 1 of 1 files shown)


erikras:
Seems like changing
```javascript
const asyncErrors = {...form.asyncErrors};
```
to
```javascript
const asyncErrors = {valid: true, ...form.asyncErrors};
```
would also accomplish this. Do you agree?

johanneslumpe:
I use `void 0` because it's shorter ;) I agree, using a default value there is more readable!

erikras:
Shortening is for bundlers. :-) I like how `false` is too long, so they all shorten it to `!1`.

johanneslumpe:
Hahah :D I just got used to using it :) Originally the use case for `void 0` was that `undefined` could be overwritten but `void 0` would always really return `undefined`.

PR #40, opened 2015-08-31
null handling in isPristine

heydiplo:
`isPristine` is throwing when value is changing from `null` to object. This is a fix.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -7,7 +7,7 @@ export function isPristine(initial, data) {
     return true;
   }
   if (typeof initial === 'object') {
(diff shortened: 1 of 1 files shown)


erikras:
I believe you that this is a real problem, but this code can't be right. What is `type`?

heydiplo:
Oh. Fixed.

erikras:
Why don't we make it `if (!data || typeof data !== 'object')` to catch `undefined`, too?

heydiplo:
Why not =)

erikras:
Third time's the charm!

PR #51, opened 2015-09-03
FIX the implementation guide, tell the user he *have to* use the key 'form' in his state

thomasguillory:
In the documentation it's stated that it's "recommended" to use "form" as key in the state.

I think this is not a recommendation but an obligation, otherwise connectReduxForm will fail
https://github.com/erikras/redux-form/blob/master/src/connectReduxForm.js#L5

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ import { createStore, combineReducers } from 'redux';
 import { reducer as formReducer } from 'redux-form';
 const reducers = {
   // ... your other reducers here ...
-  form: formReducer   // it is recommended that you use the key 'form'
+  form: formReducer   // you have to use the key 'form'
 }
 const reducer = combineReducers(reducers);
 const store = createStore(reducer);


erikras:
It's not an obligation.

You can use any key as long as you pass it into the `form` prop to `reduxForm`. But if you use the `connectReduxForm` decorator, it will force the key to be `form`, since it's doing the `connect()`ing for you, too.

thomasguillory:
Ok, I got your point. Thanks for your quick answer.

In that case a simple remark: in the "Implementation Guide", it's unclear that connectReduxForm() is optional. In STEP 1 it's stated "it's recommended to use the key 'form'" (but it's not an obligation) and in STEP 2 "Wrap your form component with connectReduxForm()".
I would say that in the context of this implementation guide, either you tell explicitly that the first step is an obligation if you don't do the connect()-ing by yourself, either you don't tell anything and let the user discover later in the doc how he can customize that key.

Otherwise, (dumb) people like me will just fail your implementation guide by using another key ;)

Thanks,

erikras:
Thanks for your input. The more clarity the better.

gaearon:
>tell the user he *have to* 

he or she.

erikras:
Thankfully docs can be written in the second person.

PR #58, opened 2015-09-08
fix DynamicForm example, fields is an object, not an array

vincentfretin:


erikras:
Thanks again, @vincentfretin.

PR #59, opened 2015-09-08
Fix initialize in example

vincentfretin:
Hi, me again, not tested, but I think it should be 'contact' instead of 'contactForm' here right?


erikras:
Yep. You are correct. :+1:

Issue #55, opened 2015-09-06
Support for nested fields.


asaf:
Hey,

I was wondering how hard it would be to support nested forms,

Something like react-form's `children` section:

http://prometheusresearch.github.io/react-forms/examples/family-editor.html

Thanks.

erikras:
The answer I've been giving to this question is that your form value _could_ be an array of other information, and you could handle mutating that array yourself before calling `handleBlur` or `handleChange`. But you're not going to get all the `dirty`/`visited`/`touched` flags for each of those sub-form inputs, so it remains a little hacky.

Another alternative would be to make each subform a separate form, e.g. `form: 'profile.children', formKey: '0'`, and merge all the form data later on submit.

If there's an elegant way to do arbitrary nesting, I haven't thought of it yet.

vincentfretin:
I'm trying to implement dynamic nested forms. So far, the solution I took is to have a separate form for each subform with formKey being:

    formKey={ formKey + '[' + field.name + '][' + String(idx) + ']' }

have my own getValues that normalize the form from this:

    form:
      _submitting: true
      Lieu:
        value: 'Lille'
    form.etablissements.0:
      _active: false
      organisme:
        value: 3

To:

    form[Lieu]: 'Lille'
    form[etablissements][0][organisme]: 3

and then using toObj function from form-data-to-object package (come initially from formsy-react) to have a nested structure like this:

    form:
      Lieu: 'Lille',
      etablissements: [{
          organisme: 3}]

Here is some code I'm currently using:
```javascript
    import { toObj } from 'form-data-to-object'

    function createStoreShape() {  // there is no way to import it from redux
      return PropTypes.shape({
        subscribe: PropTypes.func.isRequired,
        dispatch: PropTypes.func.isRequired,
        getState: PropTypes.func.isRequired
      })
    }

    const storeShape = createStoreShape(PropTypes)

    function getValues(form, fieldsInfo) {
      const values = Object.keys(form).reduce((accumulator, formKey) =>
        ({ ...accumulator,
           ...Object.keys(form[formKey]).reduce((acc, fieldName) => {
             return fieldName[0] === '_' || (fieldsInfo[fieldName] && fieldsInfo[fieldName].type === 'List') ? acc : {
               ...acc,
               [formKey + '[' + fieldName + ']']: form[formKey][fieldName].value
             }
           }, {})
         }
        ), {})
      return values
    }

    class Form extends Component {
      static propTypes = {
[...]

PR #68, opened 2015-09-15
--DO NOT MERGE YET--

asaf:
Support for nested fields.

Hey,
This is far from a complete version but it exposes a new approach for handling form fields in a nested way,

see: #55 

Thoughts?

### DIFF
--- a/src/getSet.js
+++ b/src/getSet.js
@@ -1,0 +1,17 @@
+export function getIn(obj, path) {
+  return (path.length === 0 || obj === undefined) ? obj : getIn(obj[path[0]], path.slice(1));
+};
+
+
+export function setIn(obj, path, val) {
+  if (path.length === 0)
+    return val;
+  else {
+    const sub = (obj === null) ? null : obj[path[0]];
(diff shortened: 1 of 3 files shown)


erikras:
I'd like some tests for that `getSet.js`. That's the kind of function that both a) can be easily tested, and b) really needs to work 100% correct.

erikras:
Oops, sorry about that.  I didn't see the getSet test.

asaf:
@erikras btw u can close it for now if the (1) bothers u, it needs more work,
The question if this is a good direction or not, your call.

erikras:
No, no, @asaf, I still think this is something to look into, and there is enough demand for it, but I haven't yet found the time to actually design a "compound" form to test its viability.

asaf:
@erikras Look at my concept, it doesn't break anything yet provides a powerful nested form support,

The most annoying thing is that all other corresponding states to fields (touch, errors, etc) must conform the same structure... so it takes a bit time to move everything but it shouldn't be that hard.

erikras:
Implemented.

PR #76, opened 2015-09-18
Fix async validation checking on submit to fix #75

### LINKED ISSUE #75: Async validation requires a 'valid' attribute?
I was trying to use async validation, but I'm not able to submit the form due to unknown reason.

It turns out that returning an empty object isn't accepted as a valid form.
```js
function validateContactAsync(data, dispatch) {
  // This doesn't validate the form!
  return Promise.resolve({});
}
```

I looked at the redux-form's code and found that isAsyncValid() function only looks at the 'valid' attribute of returned object, unlike isValid() function.
Which means, this works:
```js
function validateContactAsync(data, dispatch) {
  // This does validate the form.
  return Promise.resolve({valid: true});
}
```

Is this an expected behavior? I think sync / async form validation acting differently is wrong.

yoo2001818:
This fixes issue #75.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -54,10 +54,7 @@ function isAsyncValid(errors) {
   if (!errors) {
     return true;
   }
-  if (errors.valid === undefined) {
-    return !Object.keys(errors);
-  }
-  return !!errors.valid;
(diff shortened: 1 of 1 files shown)


erikras:
Ahh! My line was supposed to be `return !Object.keys(errors).length;` (was missing `.length`), but yours is probably better.

PR #78, opened 2015-09-18
Support for React Native TextInput

### LINKED ISSUE #77: Add react-native TextInput support
In order to add react-native TextInput support I added this snippet here

https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L32

  if(!(event.nativeEvent === undefined)){
    return event.nativeEvent.text;
  }

erikras:
To fix #77.

Like this, @mattiamanzati?


mattiamanzati:
Yep, that's right!

PR #87, opened 2015-09-24
Fix for missing initialState in reducer.normalize

jimpick:
Attempts to use reducer.normalize as documented would result in the following error:
    
    Uncaught TypeError: Cannot read property 'myField' of undefined
    
I've some unit tests and a fix that sets initialState in reducer.normalize

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -173,7 +173,7 @@ function decorate(target) {
           const formResult = {
             ...initialState,
             ...result[form]
-          }
+          };
           return {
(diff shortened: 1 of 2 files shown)


erikras:
I :heart: PRs with tests and with a lower number than the issue they fix. Excellent work, @jimpick. :+1:

erikras:
[v1.6.8](https://github.com/erikras/redux-form/releases/tag/v1.6.8)

PR #116, opened 2015-10-06
fix wrong "=" for object definition in readme.md

StanBoyet:


erikras:
No, the `=` is correct.

Issue #108, opened 2015-09-30
calling handleChange('fieldName', newValue) does not fire change event


bdefore:
According to the redux-form docs you can call `handleChange` passing in the field name and new value rather than have it be in an `onChange` event of the UI component sending a SyntheticEvent. But the action never fires if you call `this.props.handleChange('fieldName', newValue)`.

A workaround for now is to call `this.props.handleChange('fieldName', newValue)()`

This is on version 0.7.

erikras:
Also true on v2.0.0.

erikras:
Fixed in v2.2.4.

PR #129, opened 2015-10-14
Update handleChange to dispatch immediately when passed a value (#108)

Nicktho:
This patch:

• Updates a form's `handleChange()` prop to dispatch if passed both a field and value as stated in documentation. Related issue [#108](https://github.com/erikras/redux-form/issues/108)


erikras:
Thanks for that. Oversight on my part. :+1:

PR #132, opened 2015-10-14
Populate `initialValues` on first render

### LINKED ISSUE #97: InitialValues first render
Howdy, i found this issue when trying to render a form on the server via Reacts `React.renderToStaticMarkup`.

When the form is called with `initialValues` the `componentWillMount`-method will successfully invoke the `dispatch` with `initialize`.
However the `render` function will be called with the old reference to `form` which is not yet updated with the initial values and thus renders an empty form without the values.
In the browser this is almost invisible as the change in the store immediatelly calls a second render and the initial values show up.

I thought about a couple of possible solutions but didnt really come up with anything worth considering. Any thoughts on that?

SpainTrain:
To support server-side rendering, populate `initialValues` using `defaultValue` and `defaultChecked`

I was not ambitious enough to add component tests to add a regression test for this issue.  Hopefully that is okay for now :grin:.

Closes #97

[Review on Reviewable](https://reviewable.io/reviews/erikras/redux-form/132)

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -286,6 +286,7 @@ export default function createReduxForm(isReactNative, React) {
             const pristine = isPristine(field.value, field.initial);
             const error = syncErrors[name] || field.asyncError || field.submitError;
             const valid = isValid(error);
+            const initialValue = passableProps.initialValues && passableProps.initialValues[name];
             if (!valid) {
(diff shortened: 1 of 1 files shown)


wmertens:
:+1:

p0wl:
:+1:

PR #136, opened 2015-10-15
Add redux as peer dependency

### LINKED ISSUE #135: Define redux and react-redux as peer dependencies
Locking `redux` and `react-redux` is really annoying since it prevents using newer versions (or may duplicate packages). This is a problem since react 0.13 will use `react-redux@3.*` and 0.14 `react-redux@4.*`.

ooflorent:
Fix #135 

Now you also have to include `redux` and `react-redux` explicitly when using `redux-form`. This is not an issue since almost every developer will include them. I've also change version requirements to carets since `redux` is strictly following semver.

This PR does not change `redux` or `react` compatibility :wink:

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,9 @@
     "url": "https://github.com/erikras/redux-form/issues"
   },
   "homepage": "https://github.com/erikras/redux-form",
+  "dependencies": {
+    "react-lazy-cache": "^3.0.0"
+  },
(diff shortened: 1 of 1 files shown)


ooflorent:
Tests OK using:

- `react@0.13.3` and `react-redux@3.1.0`
- `react@0.14.0` and `react-redux@4.0.0`

PR #139, opened 2015-10-15
Allow removal of the whole form object from the reducer

bbenezech:
for proper memory management with apps that generate a significant amount of dynamic forms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -995,6 +995,10 @@ for most of your needs.**
 
 > Resets the 'touched' flag for all the fields passed in.
 
+#### -`destroy(form:String)`
+
+> Destroys the form, removing all its state.
+
 ---
   
 ## Working Demo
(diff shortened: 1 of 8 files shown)


erikras:
Excellent PR, @bbenezech!

erikras:
Even added the docs to the readme! :+1: :+1:

bbenezech:
Thanks. The PR is pretty much all copy pasta. I couldn't remove the hash key while keeping the super-duper reducer generator pattern you used. If you have any idea...

erikras:
Can you be more specific?

erikras:
Oh, I see what you mean.

bbenezech:
Yes. I just replaced the Object with undefined, I did not remove the key.

On a separated topic, I missed createAll.js

Not sure what it does and why my patch worked without it.

I'm writing a commit.

erikras:
I just noticed that you missed `createAll` as I was reviewing.

bbenezech:
annnnnd native.js... Really sorry, strike two.

Issue #45, opened 2015-09-02
Validation state in store


vkrol:
Hello!

Please, can you help me? I want to store the form validation state in the form reducer. How I can do this?

erikras:
That's an anti-pattern. The general rule of thumb for Redux, and React in general, is:

##### If you can calculate it from the state, don't store it in the state.

The less state the better.

If you want to do it anyway, you'll need your own Redux reducer, [much like how this form keeps `saveError` separate](https://github.com/erikras/react-redux-universal-hot-example/blob/master/src/components/WidgetForm.js#L8-L19), to keep it. This library isn't going to support keeping validation state in the store.

vkrol:
Thanks. But what if I need to show the form validation state (valid/invalid) in an another component?

erikras:
Two options:

* `connect()` to the form data and `import` and run your validator in the other component.
* Just `connectReduxForm()` in the other component and read the props. It doesn't matter how many components connect to the same form.

erikras:
#49 will help with this when it gets implemented.

vkrol:
Thanks. So if I have async validation and i will connect two components to the same form, then the asynchronous validation will be performed twice?

erikras:
Good question. No, the async validation will only be fired when the `BLUR` action is sent, so as long as you are only sending it from one of the components (which is necessarily the case), the async validation will only happen once, but will be displayed (available to) both components. Make sense?

vkrol:
Thanks!

anatoliyarkhipov:
Sorry, I still not understand.

Following the discussion, if I have two components which want to know about the validation state, then I should connect them both to the form and should pass the same validation function to `connectReduxForm`. Is it correct?

If it is, then it means, that the validation function will be called twice, during rendering of the first and the second components. How to avoid this?

erikras:
@anatoliyarkhipov Your validation function should be quite fast, but if it isn't, then [memoize it](https://github.com/erikras/lru-memoize).

PR #142, opened 2015-10-15
Global dirty

bbenezech:
Push dirty and pristine handling in the reducer instead of the component.

I know it sounds like duplication of state (since originals and current values are already in the store), but I think it really belongs there. It makes it really easier to read the status of a form from the outside, which can be a blessing with dynamic forms.

Is it open to discussion?

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -1,6 +1,5 @@
 import * as formActions from './actions';
 import getDisplayName from './getDisplayName';
-import isPristine from './isPristine';
 import isValid from './isValid';
 import bindActionData from './bindActionData';
 import {initialState} from './reducer';
@@ -233,7 +232,6 @@ export default function createReduxForm(isReactNative, React) {
 
           // Calculate calculable state
           let allValid = true;
-          let allPristine = true;
 
           const handleSubmit = submitOrEvent => {
             const createEventHandler = submit => event => {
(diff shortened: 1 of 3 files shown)


erikras:
Hmm.. I'm [suspicious](https://github.com/erikras/redux-form/issues/45#issuecomment-137027739).

bbenezech:
Sure, it makes sense.

I used reselect with the same algorithm, pasted here for future reference:

```javascript
import { createSelector } from 'reselect';
import isPristine from 'redux-form/lib/isPristine';

export const dirtyFormKeysSelector = forms => ({
  dirtyFormKeys: Object.keys(forms).filter(key =>
    Object.keys(forms[key]).reduce((accumulator, fieldName) =>
      accumulator || (fieldName[0] !== '_' && !isPristine(forms[key][fieldName].value, forms[key][fieldName].initial))
      , false)
  )
});

export const dirtyTodoKeysSelector = createSelector(
  state => (state.form.todo || {}),
  dirtyFormKeysSelector
);

// Todos.js
import { dirtyTodoKeysSelector } from '../selectors/todos';

@connect(dirtyTodoKeysSelector)
export default class Todos extends Component {
  static propTypes = {
    dirtyFormKeys: PropTypes.arrayOf(String).isRequired
  }
  // ...
}
```

bbenezech:
Regarding dynamic forms and redux-form,

The only thing I think is missing regarding to dynamic forms is the actions:

```javascript
import { destroy } from 'redux-form';
// ...

this.props.dispatch(destroy('todo', 'some-todo-key'));
// => {type: 'redux-form/DESTROY', form: 'todo'}
// key is missing
```

It seems something may be missing here: https://github.com/erikras/redux-form/blob/master/src/createAll.js#L14

I am not confortable with JS yet, sorry I can't provide a patch.
Of course I can use my own action creators for now, but it looks like the last missing piece to me.

erikras:
Good find. Try v2.4.1.

bbenezech:
@erikras Shouldn't we add `reset` and the other field-centric actions?

erikras:
yes, of course. duh.

It's just `initialize`, `reset` and `destroy`, right?

bbenezech:
I guessed reset was the only one remaining (unless initializeWithKey is a totally different thing).
But otherwise, don't you want to do the same for field actions (change/focus/touch)?

Actions creators should all be usable in all contexts.

PR #155, opened 2015-10-21
fixed link in README

ramdog:
Was linking to how-it-works, which appears to have been removed some time ago.  I think it should be linking to the "simple form component" example instead.


erikras:
Thanks.

PR #158, opened 2015-10-21
isPristine should return false when the number of keys is different

lenstr:
Hey guys! I have a case when initial value of a form's field is an empty array.

After updating the field via `this.props.fields['someField'].handleChange([1, 2, 3])`, `this.props.pristine` remain equal to `true`.

This pull request should fix the problem.


erikras:
Great PR. Thanks! :+1:

erikras:
Released in `v2.4.2`.

PR #160, opened 2015-10-22
Fix issue where empty err object fails validation

### LINKED ISSUE #159: Submit handler / hasErrors validation bug
I think a bug was introduced in 1ad1a71 in relation to #156  and the `hasErrors` function here https://github.com/erikras/redux-form/blob/1ad1a71b1e0a9da6c218b28a52f3700960cce73f/src/createReduxForm.js#L27

My submitHandler was not triggering at all, so after tracing the code to find where it stopped I ended up here: https://github.com/erikras/redux-form/blob/1ad1a71b1e0a9da6c218b28a52f3700960cce73f/src/createReduxForm.js#L246

For me `syncErrors` is an empty object `{}` which suggests no errors, but the `hasErrors` function returns true for it, where I assume it should return false?

Reverting to 2.4.2 fixes the issue

willmcclellan:
Fixes #159 

Another approach would be change the initial value of the reduce function to false, but I wasn't sure how that would sit with the rest of the code

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -24,7 +24,18 @@ function silenceEvents(fn) {
 }
 
 function hasErrors(errors) {
-  return !errors || Object.keys(errors).reduce((valid, error) => valid && isValid(errors[error]), true);
+  if (!errors) {
+    return false;
+  }
+
+  const errorsKeys = Object.keys(errors);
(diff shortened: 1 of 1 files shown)


erikras:
Late night coding sloppiness on my part. Thanks.

Issue #161, opened 2015-10-23
Potential issue with asynchronous validation


skeary-immt:
Hi Erik,

Thanks for a great library. Using v2.4.4 of the library we also came across the problem outlined in https://github.com/erikras/redux-form/issues/159. Moving to v2.4.5 seems to address that problem but we're still having an issue with a form that was previously working.

The new issue appears to be related to the asynchronous validation the form uses as it (now) passes the synchronous validation. Looking through the redux-form code I think there may be an issue with the changes made in v2.4.2 (commit  1ad1a71). The code that seems to be the problem is in src/createReduxForm.js:

```
const handleErrors = asyncErrors => {
             dispatch(actions.stopAsyncValidation(asyncErrors));
-            return isAsyncValid(asyncErrors);
+            return hasErrors(asyncErrors);
           };
```

Before this routine used to return *true* if the form was valid and *false* if not. Now it seems to return *false* if the form is valid and *true* if not. It doesn't look like the calling code has changed to reflect this.

if you like I can submit a PR. I haven't yet because I'm not sure if I'm missing something. Let me know if you want me to.

Cheers,
Simon

erikras:
[Upgrade to v3](https://github.com/erikras/redux-form/releases/tag/v3.0.0).

PR #164, opened 2015-10-24
fixed not submitting if there are no async errors

yoo2001818:
Currently `handleSubmit` routine checks whether if there are async errors by doing this:
```js
return this.runAsyncValidation(actions, values).then(asyncValid => {
  if (allValid && asyncValid) {
    ...
  }
}
```

But `runAsyncValidation` function returns true **if there are errors**:
```js
return hasErrors(asyncErrors);
```

Since this function returns true if there are errors, form can be only submitted if there are async errors in the form.

This pull request fixes this issue by changing `runAsyncValidation` to return an object of errors (thus acting same as `runSyncValidation` to give consistency).

```js
return this.runAsyncValidation(actions, values).then(asyncErrors => {
  if (allValid && !hasErrors(asyncErrors)) {
    ...
  }
}
```

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -195,7 +195,7 @@ export default function createReduxForm(isReactNative, React) {
           }
           const handleErrors = asyncErrors => {
             dispatch(actions.stopAsyncValidation(asyncErrors));
-            return hasErrors(asyncErrors);
+            return asyncErrors;
           };
           return promise.then(handleErrors, handleErrors);
         }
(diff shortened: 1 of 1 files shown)


yoo2001818:
Seems like this is related to #161.

yoo2001818:
It'd be good if there were unit tests for submit logic, but since submit logic is too tied to React component, it'd be hard to test it. I think seperating them is good idea.

erikras:
v3, hopefully released next week, will have extensive unit tests.

erikras:
Fixed in v3.

PR #178, opened 2015-10-27
Support nested fields

yesmeck:
I didn't touch any validation related codes. Validation can be wrote as before.

Example: 

https://github.com/yesmeck/redux-form/blob/gh-pages/src/examples/HighOrderNestedForm.js
https://github.com/yesmeck/redux-form/blob/gh-pages/src/examples/NestedForm.js

### DIFF
--- a/src/readFields.js
+++ b/src/readFields.js
@@ -7,6 +7,8 @@ import isPristine from './isPristine';
 import isValid from './isValid';
 import getValues from './getValues';
 import silencePromise from './silencePromise';
+import set from 'lodash/object/set';
+import get from 'lodash/object/get';
 
 /**
  * Reads props and generates (or updates) field structure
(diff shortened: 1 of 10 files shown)


src/readFields.js
```diff
@@ -7,6 +7,7 @@ import isPristine from './isPristine';
 import isValid from './isValid';
 import getValues from './getValues';
 import silencePromise from './silencePromise';
+import {get} from 'lodash';
```
vincentfretin: If you want to reduce the size of the generated bundle, use:
import get from 'lodash/object/get';

yesmeck: Thanks, fixed.

yesmeck:
@erikras A little thing, the action `initialize` need a extra argument `fields now, it breaks the compatibility.

rogeriochaves:
@yesmeck @erikras any updates on this? It looks great, I'm really needing it

yesmeck:
As @erikras said he will [implement](https://github.com/erikras/redux-form/issues/55#issuecomment-154722417)  something on this, I'm closing this hacky implementing.

PR #213, opened 2015-11-02
allow initialValues to be in the `mapPropsToValues` form

iamdustan:
This allows for the `initialValues` property to be in the `mapPropsToValues` form, matching an idiom already familiar to the redux world. This also provides an entry-point in taking advantage of react props in conjuction with a reduxForm.

```
const Form = reduxForm({
  initialValues: (props) => ({query: props.query})
});

const RenderedForm => (
  <Form field={location.query.query} />
)
```

### DIFF
--- a/src/createHigherOrderComponent.js
+++ b/src/createHigherOrderComponent.js
@@ -68,7 +68,7 @@ const createHigherOrderComponent = (config, isReactNative, React, WrappedCompone
       componentWillMount() {
         const {initialize, initialValues} = this.props;
         if (initialValues) {
-          initialize(initialValues);
+          initialize(typeof initialValues === 'function' ? initialValues(this.props) : initialValues);
         }
       }
 


iamdustan:
Thanks Erik!

erikras:
@iamdustan Presumably you want this because you are `connect()`ing your form to some other reducer and want to initialize form values based on a value elsewhere in the state? How would you feel about the redux-form decorated component initializing itself any time that the `initialValues` prop changes? That would accomplish the same thing, as you could just `mapStateToProps` to the `initialValues` prop.

iamdustan:
Are you saying that initialValues can be provided like the following?

```
const MyForm = reduxForm(options)(DumbForm);

export connect((state) => ({initialValues: state.initialValues})(MyForm);
```

From the docs it appears as though only the following is supported:

```
const options = {
  initialValues: {staticList: 'ofValues'}
};

const MyForm = reduxForm(options)(DumbForm);

export MyForm;
```

the general case is I want initialValues to be derived data and not static data; ideally without having to create my own instantiated component that generates a new `reduxForm(options)(DumbForm)` HOC in lifecycle events.

erikras:
Yes. I'm currently implementing so that the following will work:
```js
const MyForm = reduxForm(
  config, // same config as always
  state => ({initialValues: state.yourReducer.anyData}) // mapStateToProps just like connect() provides
)(DumbForm);
```
It sounds like that will suit your needs.

iamdustan:
Cool. I’m fairly certain that would work for us.

erikras:
[Released in v3](http://erikras.github.io/redux-form/#/examples/initializing-from-state).

PR #240, opened 2015-11-06
add a jsnext:main so we can distribute ES2015 modules

Whoaa512:
See https://github.com/rollup/rollup#can-i-distribute-my-package-as-an-es6-module

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "3.0.5",
   "description": "A higher order component decorator for forms using Redux and React",
   "main": "./lib/index.js",
+  "jsnext:main": "./src/index.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/erikras/redux-form"


erikras:
Should the webpack build also generate a separate module that is transpiled to es6, not es5?

erikras:
Oh, I see. You're pointing it at *src*! Interesting.

taion:
This is incorrect, BTW: https://github.com/rollup/rollup/wiki/jsnext:main

`jsnext:main` should be a transpiled build with everything but the ES modules transform.

See e.g. how Redux and React Router do it.

swernerx:
That @taion is saying is right. The current `jsnext:main` entry breaks webpack 2 with babel as one typically do not transpile content from the node_modules folder.

swernerx:
See also: https://github.com/reactjs/redux/issues/1042

taion:
Ahh, this was actually fixed on the beta releases.

erikras:
@taion, so is this something I need to look into or not?

taion:
@erikras 

Probably not. It was fixed in the `v6` branch in https://github.com/erikras/redux-form/commit/56a3af112b0cc2c51cbb4752a9bc7d41e57e5a2d.

I was going off a comment elsewhere and didn't realize that your development code was not on the `master` branch.

PR #253, opened 2015-11-10
Update GettingStarted.md example

InSuperposition:
changed `props` to `this.props` in ContactForm's component example


erikras:
_This_ is why I made all the examples use real code. Because writing flawless pseudocode is impossible. Thanks!

PR #258, opened 2015-11-11
Fix README link

BobWhitelock:
This commit fixes the README link to the redux-form Github pages docs; this wasn't linking correctly as it seemed to be a relative rather than absolute link.


erikras:
Thanks.

PR #274, opened 2015-11-13
Protect IE8, disallow es7 class properties

mattydoincode:
Hi friends,

This PR protects ie8 support by disallowing es7 class properties (which are still experimental). These in particular break ie8 because babel must use Object.defineProperty in order to support them. They can be easily handled by the reformatting in this PR. Let me know if there is anything else I can do to match the standard of redux-form with regards to this PR!


mattydoincode:
Also it seems like I need to do some merging...

erikras:
Why is this an issue? The [actual lib that you get from NPM](https://npmcdn.com/redux-form@3.0.7/lib/createHigherOrderComponent.js) is transpiled down to ES5.

mattydoincode:
Some context can be found here: https://github.com/rackt/react-redux/issues/133. ES5 isn't 100% supported even with polyfills. So yes it gets transpiled to ES5, but this particular syntax sugar is a breaker. Here is my related PR on react-router: https://github.com/rackt/react-router/pull/2540

erikras:
Hmm... Seems like the problem is really with Babel, not the libraries that use it, but it takes less force to accelerate less mass, so I guess I'm fine with this, once you merge in the changes.

erikras:
Sorry for not merging this pull request, but I had already changed some of these lines when I got around to this PR, so I've just implemented your changes verbatim in my modified repo. Thanks for this!

mattydoincode:
Thanks! No worries. I'm just sorry to be the bringer of ie8 drama. Rock on!

erikras:
Yeah. My favorite thing about [`uglify`](https://github.com/mishoo/UglifyJS2) is the `--screw-ie8` parameter. :smiley:

PR #289, opened 2015-11-17
Document File Input Quirks

solomonhawk:
Hi,

I ran into some confusion working with file inputs and `redux-form` and I thought it would be useful to document this within the project itself. I definitely think this could be cleaned up and improved but wanted to get some :eyes: on it nonetheless.

Questions:
* Does this belong in the `redux-form` examples?
* Is there a current best practice for handling file inputs in client-side apps?
* Should I re-do my commit messages to align with the project style?

Thanks!


src/pages/App.scss
```diff
@@ -74,6 +74,10 @@ a code {
   text-decoration: underline;
 }
 
+pre code {
```
solomonhawk: This ought to be vetted by someone more familiar with the project.

I wanted to use triple backticks to surround code blocks inside the markdown file for my example but the text was unreadable since bootstrap changes the color of `pre code` elements.

src/pages/examples/File.md
```diff
@@ -0,0 +1,43 @@
...
+
+* FireFox: `SecurityError: The operation is insecure.`
+
+* Safari: `Error: InvalidStateError: DOM Exception 11`
+
+Inputs with `type="file"` do not support setting the `value` property programmatically<sup>[1](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#File_inputs)</sup>.
+
+To avoid this warning, remove the `value` property from the associated file field present in the `fields` prop that `redux-form` provides to your wrapped component.
+
+As seen in the [Simple](#/examples/simple) example, for most input types it is sufficient to spread the properties passed in by `redux-form`:
```
solomonhawk: Trying to balance context and terseness here. Perhaps there are improvements to be made here by way of using more consistent vocabulary?

src/examples/FileInputForm.js
```diff
@@ -0,0 +1,37 @@
...
+    const {
+      fields: {avatar},
+      handleSubmit,
+      resetForm
+      } = this.props;
+    return (<form className="form-horizontal" onSubmit={handleSubmit}>
+        <div className="form-group">
+          <label className="col-xs-4 control-label">Avatar</label>
+          <div className="col-xs-8">
+            <input type="file" className="form-control" {...avatar} value={ null } />
```
solomonhawk: Perhaps it would be better to demonstrate the alternative which is pulling `value` out of `avatar` and then spreading the rest of the properties?

src/pages/App.scss
erikras: It's expecting you to have a `javascript` immediately following your triple-backtick. That enables the code coloring.

src/examples/FileInputForm.js
erikras: Can you give that example? (just here in a comment) I'm not sure I understand.

erikras:
This is _great_!

Is the submitted value supposed to be:
```js
{
  "avatar": {
    "0": {}
  }
}
```
?? That's what I'm seeing when I run it.

solomonhawk:
Thanks!

That's what I see as well. I believe this is just a representation of the `FileList` object returned as `files` from the input. That's what I see when I actually inspect the state that `redux-form` manages.

Would it be too much to also provide a warning whenever `redux-form` sees a `event.files` in `getValue` that references this example in the docs?

src/pages/App.scss
solomonhawk: Understood, willfix.

src/examples/FileInputForm.js
solomonhawk: It is mentioned at the end of the markdown file. Alternative looks like:

```
render() {
  const { fields: { avatar } } = this.props;
  // pull out value so it isn't passed into the file input
  let { value, ...avatarProps } = avatar;

  return (
    ...
      <input type="file" { ...avatarProps } />
    ...
  );
}
```

Downside is an additional unused binding (`let value = avatar.value`). Perhaps this is easier to understand than setting `value={ null }` on the input itself.

erikras:
Perhaps the thing that is rendering the json in the examples could be smarter about `FileList`. But it looks good when you inspect it in the debugger?

solomonhawk:
@erikras I changed some of the text and added a custom `replacer` to use with `JSON.stringify` in the `FormValues` component. Works like a charm.

screen shot 2015-11-20 at 9 38 54 am
screen shot 2015-11-20 at 9 38 39 am

Issue #294, opened 2015-11-18
onSubmit prop not working with React.cloneElement?


hunterc:
I have a form component, `SampleForm`, that I am passing an `onSubmit` prop to.
I am passing my form as a child to a wrapper `Card` component. The wrapper component clones the form and overrides the `onSubmit` prop with a composed version that closes the card.

Issue I'm running into is that for some reason the `onSubmit` prop isn't making it to the form. If I change the prop name to anything else it shows up.

Example:
```
<Card>
  <div>...</div>
  <SampleForm
    onSubmit={() => console.log('submit')}
  />
</Card>
```

Card.jsx bits of interest
```
let form = this.props.children[1];
let clonedForm = React.cloneElement(form, {
  onSubmit: (::this.composeSubmit(form.props))
});
```

```
composeSubmit({ onSubmit }) {
  return (data) => {
    ::this.closeCard();
    onSubmit(data);
  };
}
```

When I check the props of my form component the `onSubmit` prop is undefined.

erikras:
What does this print out?
```js
let form = this.props.children[1];
console.log(form.props);
```
Is there an `onSubmit` in there?

hunterc:
![screen shot 2015-11-18 at 10 00 42 am](https://cloud.githubusercontent.com/assets/2365571/11244366/475fcb2e-8ddb-11e5-949f-6f2dd5f3c472.png)
The chrome collapsed preview doesn't show onSubmit but when I expand it shows up.

erikras:
I'm not understanding the problem. Can you rephrase it?

hunterc:
Sure sorry for any confusion. The chrome bit was just an observation.

The issue I'm having is that while the props are showing up on the original form, this.props.children[1], when I clone the element with `React.cloneElement`, the `onSubmit` prop is missing.

erikras:
I have never used `React.cloneElement`, so I am speaking from a position of ignorance, but maybe, just maybe, `React.cloneElement` is cloning the virtual dom (what React Components output), not the _React Components_ themselves, so you're losing all of the logic built into the `ReduxForm` component? Is that a possibility?

hunterc:
Pretty sure it just operates on the component itself. [source](https://github.com/facebook/react/blob/0b29035484f428cb56e7e1c04a88f66ac020d1d4/src/isomorphic/classic/element/ReactElement.js#L215)

Thought onSubmit might be a reserved word for the clone but doesn't appear to be.

[...]

PR #297, opened 2015-11-18
Add onSubmit prop to WrappedComponent

hunterc:
The `WrappedComponent` current loses its reference to `onSubmit` because it is pulled out of `this.props` and not passed back to`WrappedComponent`.

This restores that reference and seems to solve #294.


erikras:
I still don't think this is the correct solution, for the reasons discussed in #294.

PR #302, opened 2015-11-19
Remove spread operators

yesmeck:
Spread operators is very slow when there are many fields (I have a 200+ fields huge form).

Compare using spread operators and not: http://jsbin.com/tibiji/edit?js,console


erikras:
This makes me sad, since I love the spread operator so much, but I can't argue with the numbers.

Thank you for including the demo.

PR #317, opened 2015-11-20
Allow fields to be a function

joearasin:
I'm not sure if this is the ideal API, but we chatted about this on discord a bit.

This enables fields to be expressed as a function from the previous fields object to an array.


joearasin:
Closing this -- The current approach to nested fields seems to handle this better, and my experience giving this API a spin is that it's awkward. Pushing something like this into the reducer might be a better approach.

erikras:
I agree. Something like this might warrant implementation in the future, but it's less important after the nested field thing.

PR #318, opened 2015-11-20
Document Fields as a function

joearasin:
Example that goes along with #317


joearasin:
Closing. See: https://github.com/erikras/redux-form/pull/317

PR #327, opened 2015-11-24
Add ability to pass in arguments to resolve method

agirton-twitter:
If user passes in an argument to the resolve method the library shouldn't treat this as an error. This may not be the best practice but just in case someone needs to wait till that promise has been resolved
to fire an action from there, this update will handle it appropriately.

### DIFF
--- a/src/__tests__/asyncValidation.spec.js
+++ b/src/__tests__/asyncValidation.spec.js
@@ -19,8 +19,22 @@ describe('asyncValidation', () => {
     expect(isPromise(asyncValidation(fn, start, stop))).toBe(true);
   });
 
-  it('should call start, fn, and stop on promise resolve', () => {
+  it('should call start, fn, and stop on promise resolve with no arguments', () => {
     const fn = createSpy().andReturn(Promise.resolve());
+    const start = createSpy();
+    const stop = createSpy();
+    const promise = asyncValidation(fn, start, stop);
+    expect(fn).toHaveBeenCalled();
+    expect(start).toHaveBeenCalled();
+    return promise.then(() => {
+      expect(stop).toHaveBeenCalled();
+    }, () => {
+      expect(false).toBe(true); // should not get into reject branch
+    });
+  });
+
+  it('should call start, fn, and stop on promise resolve with arguments', () => {
+    const fn = createSpy().andReturn(Promise.resolve({foo: 'success'}));
     const start = createSpy();
     const stop = createSpy();
     const promise = asyncValidation(fn, start, stop);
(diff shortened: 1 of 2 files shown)


erikras:
Hmmm... This is how `<v2.x` did it, but it seemed like a better practice to force `reject` for errors.

You may not have noticed that there is a [`returnRejectedSubmitPromise` config option](http://erikras.github.io/redux-form/#/api/reduxForm) that will let you run code _after_ an error has been returned. Don't you think this is sufficient?

Excellent PR, btw; I'm just not sure it's the direction the library should go. But feel free to convince me otherwise.

agirton-twitter:
Thanks for the review @erikras! I can only see two scenarios where this would be helpful. Debugging and using the value after the promise has been resolved using the `then` method. 

While getting up to speed with the library I wasn't aware that if I passed a valid object to resolve that it would treat it as an error. Since I'm already handling the action, the success response still fired, but the rest of the form validation didn't. Once I realized this caused it, everything else worked as expected. 

The second use case is probably not necessary in such a niche part of the validation flow since the user should only be firing one for the piece they want to validate. But just in case they did want to, they could here.

To give you a bit more context this is what my code looked like before and after.

```diff
const handleTwitterError = (result, resolve, reject) => {
  if (result.error && result.error.data) {
    return reject({username: 'Sorry, that user does not exist'});
  }
-return resolve(result);
+return resolve();
};

const handleEmailError = (data, resolve, reject) => {
  if (data) {
    return reject({email: 'Sorry, that email has already been used'});
  }
-return resolve(data);
+return resolve();
};
```

erikras:
OH!! I didn't notice that this was just on `asyncValidate`.

So the main problem is:

1. It makes sense to force errors to come via promise rejection on _submit_, because your submit action might want to return a value, such as the canonical record from the database.
2. Your reasoning _also_ makes sense for async to allow errors on promise resolution, since it doesn't make any sense to ever pass anything other than `{}` back from a successful async validation.
3. It feels a little inconsistent to have those behave differently.

I suspect that # 2 is strong enough to outweigh # 3.

Please add the test cases to `handleSubmit.spec.js` to make sure that the submit will not go through if `asyncValidate` returns a promise that resolves to anything other than `{}`.

agirton-twitter:
Hey @erikras thanks again for helping think through the solution. Just so we are on the same path, my intention with this was not to resolve errors but to handle success responses. My functions could be named better to describe that it handles errors but also if the response is successful it would resolve those. Essentially handling either success or failure. 

Users of the library probably shouldn't do anything with the successful response since that should be an action from the API call. But figured it would be best for the library to not stop async validation if they resolve with successful data.

erikras:
I'm going to close this, because whatever comes of the discussion on #602 will probably supersede this.

PR #328, opened 2015-11-24
Tests for submit function being called

stevoland:
...when passed directly to `handleSubmit`.

Cheers, Erik - enjoying setting up redux-form on a new project.

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -1,6 +1,6 @@
 /* eslint react/no-multi-comp:0*/
 import expect from 'expect';
-import React, {Component} from 'react';
+import React, {Component, PropTypes} from 'react';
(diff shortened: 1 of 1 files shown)


erikras:
1. Awesome PR, thanks!
2. I'm such a testing n00b that I didn't know about the `done()` function. Nice!
3. You needn't call [`e.preventDefault()`](https://github.com/stevoland/redux-form/blob/submit/src/__tests__/createReduxForm.spec.js#L642); `redux-form` does that for you. You can just do `<form onSubmit={this.props.handleSubmit(submit)}/>`

If you make that change, I'll be happy to merge.

stevoland:
Aha! so I was using the API wrong. Thanks! Here's 2 working tests then.

erikras:
I like working tests better. :smile: Cheers. :+1:

PR #356, opened 2015-12-01
Set dirty when array field changes

mdesrosiers:
`readField`'s return value can be an array so in order to determine if a form is dirty, I handle this case by iterating through all the elements.

### DIFF
--- a/src/readFields.js
+++ b/src/readFields.js
@@ -16,14 +16,20 @@ const readFields = (props, myFields, asyncValidate, isReactNative) => {
   const fieldObjects = {...myFields};
   fields.forEach(name => {
     const result = readField(form, name, undefined, fieldObjects, syncErrors, asyncValidate, isReactNative, props);
-    if (result.invalid) {
-      allValid = false;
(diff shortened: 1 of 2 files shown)


src/readFields.js
```diff
@@ -16,14 +16,20 @@ const readFields = (props, myFields, asyncValidate, isReactNative) => {
...
-    }
-    if (result.dirty) {
-      allPristine = false;
-    }
-    if (result.error) {
-      errors = write(name, result.error, errors);
+    if (Array.isArray(result)) {
+      allValid = result.every(res => !res.invalid);
+      allPristine = result.every(res => !res.dirty);
+      errors = result.every(res => write(name, res.error, errors));
```
mdesrosiers: I'm not sure if I'm setting `errors` correctly here.

erikras: Not quite. I added a few more tests, too, to confirm that everything is working.

erikras:
Thanks for this!! :+1:

PR #402, opened 2015-12-09
return undefined if an object is not passed to getValuesFromState before Object.keys is executed

Plummat:
In some cases, the boolean "true" gets passed to getValuesFromState. While this returns an empty array in Chrome and Edge, it returns an error in IE11. This added check with prevent the error(and thusly, the form exploding) in IE11.

### DIFF
--- a/src/getValuesFromState.js
+++ b/src/getValuesFromState.js
@@ -5,6 +5,10 @@ const getValuesFromState = state => {
   if (!state) {
     return state;
   }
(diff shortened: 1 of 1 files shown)


Plummat:
Hey Erik, thanks for pushing out redux-form 4, have you had a chance to review this? I need to use a forked version until this gets pulled in to support IE11. :). Thanks! .

Issue #370, opened 2015-12-02
initialValues overwrites existing form value


qnm:
My form uses `destroyOnUnmount: false` to persist values across route changes but, when a user leaves the page and eventually returns to this form, `initialValues` will overwrite the existing user data.

Is this by design or a bug? `initialValues` suggests to me the values will only be used initially and ignored after user input.

If the latter, I'm happy to code up a test case and solution, otherwise I'll guess I'll have to find a work-around.

erikras:
This is an interesting edge case. There are two things going on.

1. The `INITIALIZE` action sets both the `value` and `initial` keys in the store. I think this is the correct behavior.
2. All the `initialValues` prop does is dispatch an `INITIALIZE` action on `componentWillMount`, which is also think is correct _for almost all cases_.

However, perhaps `initialValues` should add an additional `noOverwrite` boolean to the `INITIALIZE` action that prevents it from overwriting the `value` if it finds one already there. What I can't decide is whether or not that should be the default behavior, or if you should have to pass a `noOverwriteOnInitialize={true}` prop.

Thoughts?

qnm:
> However, perhaps initialValues should add an additional noOverwrite boolean to the INITIALIZE action that prevents it from overwriting the value if it finds one already there.

Yes - that makes sense. We might need to re-consider the name through, `noOverwrite: false` needs a little thought to understand whether or not an overwrite may take place!

> What I can't decide is whether or not that should be the default behavior, or if you should have to pass a noOverwriteOnInitialize={true} prop.

As you've stated, this is an edge case, and only applied to forms using `destroyOnUnmount: false`. 

If folks are using `destroyOnUnmount: false` to support multi-step forms, I'd assume that most multi-step forms allow users to move backwards in the process.

From this, my assumption is that most people using `destroyOnUnmount: false` in a multi-step form would want `noOverwrite` behaviour by default.

Additionally, the name `initialValues`suggests it's only used initially!

My vote is for `noOverwite` to be the default behaviour.

webmasterkai:
I'm having the same issue and found the `initialValues ` name counter-intuitive. I too thought it was only used initially. I was wanting to save the values if the user happens to navigate away from the page and later comes back to edit/save. Why would you **ever** want `initialValues` to overwrite the previously set `value` keys? I could maybe understand overwriting the `initial` keys, but not `value`.

[...]

PR #423, opened 2015-12-15
Issue 370

bangn:
This PR is used to fix issue 370: https://github.com/erikras/redux-form/issues/370

Summary: 
If the form has already been initialized, the `defaultValue` of form's fields will be set. In `componentWillMount`, only call `initialize` if no field has been set( `defaultValue === undefined`)

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -127,6 +127,7 @@ describe('createReduxForm', () => {
       error: undefined,
       touched: false,
       visited: false,
+      isInitialized: true,
       readonly: false
     });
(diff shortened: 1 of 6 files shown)


squallstar:
+1

erikras:
Both of these tests pass with the current `v4` code base, and the test cases are covered already. Thanks, though.

erikras:
Rather than trying to figure out if a form had been initialized from the `defaultValue`s, I've changed it to just flip an `_initialized` flag in the redux state, so we know for sure. Released as [`v4.1.0`](https://github.com/erikras/redux-form/releases/tag/v4.1.0)

PR #411, opened 2015-12-11
Fix initialising a nested array value with an object.

latentflip:
Fixes: #377

Hope this makes sense. Passes all existing tests, and the new one I added.

I know you're looking at reworking the init stuff anyway as per the comments in #377 but it'd be great to get a fix in for this for what we're doing :)

nb: also fixed the title of a test that looks like it's wrong.

<3

PR #425, opened 2015-12-15
Fix initialising a nested array value with an object in v4

### LINKED ISSUE #377: Initial value for deeply nested forms when adding a field.
Is there any way I could add an initial value for nested form but only when I want to add it?
````js
export const fields = [
  "content",
  "children[].firstname",
  "children[].lastname"
];
````
And when I call `children.addField()` I want firstname to have an initial value. Looking into the code I see that addField receives 2 optional arguments and one of them is `value`, but if I try to use it

````js
children.addField({firstname: 'Tom', lastname: 'Whatever'})
````

it doesn't seem to work. I'm not sure I'm doing this right. 
I'm using v3.1.7.

lukekarrys:
This works off #411 and passes the `fields` to `initializeState` as required by v4.

This makes the test from #411 pass now, which I **think** is all thats required to fix #377. I also tested this fix in an app that I just upgraded to v4 and it fixes the behavior I was seeing.

### DIFF
--- a/src/__tests__/reducer.spec.js
+++ b/src/__tests__/reducer.spec.js
@@ -261,6 +261,216 @@ describe('reducer', () => {
       });
   });
 
+  it('should push a deep array value which is a nested object', () => {
+    const state = reducer({
+      testForm: {
+        myField: [
+          {
+            foo: {
+              initial: {a: 'foo-a1', b: 'foo-b1'},
+              value: {a: 'foo-a1', b: 'foo-b1'},
+            },
+            bar: {
+              initial: {a: 'bar-a1', b: 'bar-b1'},
+              value: {a: 'bar-a1', b: 'bar-b1'},
+            }
+          },
+          {
+            foo: {
+              initial: {a: 'foo-a2', b: 'foo-b2'},
+              value: {a: 'foo-a2', b: 'foo-b2'},
+            },
+            bar: {
+              initial: {a: 'bar-a2', b: 'bar-b2'},
+              value: {a: 'bar-a2', b: 'bar-b2'},
+            }
+          },
+        ],
+        _active: undefined,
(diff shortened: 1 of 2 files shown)


erikras:
No, you can't just pass `Object.keys()` as the `fields` list. That works for flat objects, but ideally we'd support arbitrarily deep structures.

lukekarrys:
@erikras I added a few more tests for what I thought you meant by arbitrarily deep structures, but I must be confusing something since I couldn't get them to fail. :smile:

Do you have an example about what you mean? I'd love to contribute a test and fix to get this working the right way.

erikras:
Dammit. I was trying to push a test to this PR to show you what I mean, and accidentally merged it to `master`.

erikras:
Test is currently failing for the "you can't just pass `Object.keys()` as the `fields` list" reason I mentioned above.

erikras:
The solution is going to require [the thing that creates the `addField()` function in `readField()`](https://github.com/erikras/redux-form/blob/c18736569ff7b9db8cea1968e6e0c340e9e8ff59/src/readField.js#L29-L31) to pass along the subset of fields to know the shape of the array items.

lukekarrys:
Thanks for the clarification @erikras. I'll work on getting that test passing and submit a new PR.

lukekarrys:
I couldn't get the test to pass by modifying anything in `readField()`. I think the failing test only touches `reducer()` which doesn't use `readField()` directly?

I could get it to pass by taking the nested values object and turning it into an array of keys like `['name', 'address.street', 'address.postalCode']` for the fields in that specific test. However doing that made my added `should push a deep array value which is a nested object` test fail.

flaub:
@erikras I believe I've implemented passing the subset of fields to the action with PR #617

PR #438, opened 2015-12-17
Add asyncValidating prop in case of async validation

KateKate:
I would like to offer asyncValidation prop for all fields, that are processed ajax validation. So it allows to do the following:
```
if (field.asyncValidating) {
    return <i className="fa fa-spinner fa-spin fa-2x"></i>;
  }
```
Please have a look. One note. It doesn't work when state is array w/o field name.
The main reason of this propose that form _asyncValidating prop hardly helps when you have several fields width async validation. It makes you invent some tricks.
Please let me know your thoughts. Thanks in advance!

### DIFF
--- a/src/__tests__/actions.spec.js
+++ b/src/__tests__/actions.spec.js
@@ -93,7 +93,10 @@ describe('actions', () => {
   });
 
   it('should create startAsyncValidation action', () => {
-    expect(startAsyncValidation()).toEqual({type: START_ASYNC_VALIDATION});
+    expect(startAsyncValidation('myField')).toEqual({
+      type: START_ASYNC_VALIDATION,
+      field: 'myField'
+    });
   });
 
   it('should create startSubmit action', () => {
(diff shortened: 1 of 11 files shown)


erikras:
This is close to excellent. Changes I'd propose:

Why not just save the currently async validating field name in the root of the state, like `_active` does on `FOCUS`? In fact, it can replace the current `_asyncValidating` boolean. That way `stopAsyncValidation()` doesn't even need to pass the field; it can just clear that key from the state. Also, by putting it in the root of the state, `setErrors()` needn't change.

KateKate:
Thank you for the feedback. I'm considering to have asyncValidating as field prop to reduce props passing to the field component. But you offer looks better in terms of the redux-form architecture. So I made it and send new PR to keep clear commits' history.

erikras:
Other PR mentioned in previous comment was merged. Closing this one.

PR #445, opened 2015-12-18
Save field name as asyncValidating prop

KateKate:
Based on your comments I've changed the code. Please have a look. To show loading icon you could write:
```
asyncValidating === field.name ? <i className="fa fa-spinner fa-spin fa-2x"></i> : null;
```

### DIFF
--- a/src/__tests__/actions.spec.js
+++ b/src/__tests__/actions.spec.js
@@ -93,7 +93,10 @@ describe('actions', () => {
   });
 
   it('should create startAsyncValidation action', () => {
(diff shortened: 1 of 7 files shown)


KateKate:
It's strange that one test was failed. I run it locally and everything is good.

KateKate:
@erikras please let me know your feedback about my PR. Do you plan to integrate that feature? Thanks in advance!

erikras:
Awesome work. Thanks, @KateKate! :+1:

Issue #446, opened 2015-12-18
Submit validation not working anymore ?


yleclanche:
Hello,

The submit validation example page (http://erikras.github.io/redux-form/#/examples/submit-validation) doesn't seems to work anymore.
On my local dev, I get "Uncaught (in promise) TypeError: Cannot read property 'submitError' of null" when I submit my form.
Not sure if It's related though.

hnordt:
Same error here, even the API returns 200 OK and promise is resolved.

hnordt:
Looks like the commit that introduced this bug is ee7a64f20a39bd147c80623d2dd0e682827f39dc

VinSpee:
Also experiencing this issue. Getting a touch  but no validation :-(

hnordt:
@VinSpee use `v4.0.1` for now. It's working.

box-turtle:
I get same error in both 4.0.1 and 4.0.3 - "Uncaught (in promise) TypeError: Cannot read property 'submitError' of null"

erikras:
Thanks guys. :+1: I will investigate. :mag:

erikthedeveloper:
Ahhh thanks so much @hnordt I just spent 2+ hours wrestling this (new to `redux-form` and trying to dive straight into server side validation and hitting a wall with this error) and I was beginning to think I was just crazy. Downgraded my version to `v4.0.1` and now it is working exactly as expected :+1: 

Lesson learned: Consult GitHub Issues **before** Google. Thanks!

box-turtle:
I believe the issue is the dreaded "null is an object" javascript quirk. 

If I change line 11 of setErrors.js  in v4.0.3 from 

```    if (typeof state === 'object') {  ```

to 

```    if (state && typeof state === 'object') {  ```

then the error goes away for me and validation works as expected. Not sure if that's the best way to do the null check because I'm not sure I fully understand the recursion logic that's going on here.

erikras:
Excellent catch, @box-turtle! I've been bitten by that quirk more than once.

erikras:
What puzzles me is how [these tests](https://github.com/erikras/redux-form/blob/master/src/__tests__/createReduxForm.spec.js#L852-L925) are passing, even with this broken.

morgante:
@erikras It doesn't look like those tests actually check about asynchronous validation on form submission?

erikras:
Fixed published. Problem in example was (at least partially) an actual bug in the example.

[...]

PR #452, opened 2015-12-21
Adds check for null state within setErrors

erikthedeveloper:
Previously was throwing error "TypeError: Cannot convert undefined or null to object"

This was being encountered in relation to async form submission [as seen in the submit-validation demo](http://erikras.github.io/redux-form/#/examples/submit-validation?_k=5igft0)

Credit: Implemented fix suggested by @box-turtle here https://github.com/erikras/redux-form/issues/446#issuecomment-166014354


erikras:
This does not fix the demo for me.

erikras:
Nevermind. There was actually a bug in the example.

PR #463, opened 2015-12-24
Correct imported actions

bs1180:
I noticed that the startSubmit and stopSubmit action creators are actually [ exported](https://github.com/erikras/redux-form/blob/master/src/index.js#L20) from the top level redux-form object. This just corrects the documentation.


erikras:
Oops. Thanks!

PR #483, opened 2015-12-29
allow mapDispatchToProps to override dispatch property

quicksnap:
In my project, I create a wrapper around dispatch to catch promise rejections. Passing in that overridden dispatch function as follows would be nice:

```js
const WrappedForm = reduxForm(
  { form: 'login' },
  state => ({}),
  dispatch => ({ dispatch: bindSafeDispatch(dispatch) }),
)(LoginForm);
```

All tests pass =)

### DIFF
--- a/src/wrapMapDispatchToProps.js
+++ b/src/wrapMapDispatchToProps.js
@@ -5,26 +5,26 @@ const wrapMapDispatchToProps = (mapDispatchToProps, actionCreators) => {
     if (typeof mapDispatchToProps === 'function') {
       if (mapDispatchToProps.length > 1) {
         return (dispatch, ownProps) => ({
+          dispatch,
           ...mapDispatchToProps(dispatch, ownProps),
-          ...bindActionCreators(actionCreators, dispatch),
-          dispatch
+          ...bindActionCreators(actionCreators, dispatch)
         });
       }
       return dispatch => ({
+        dispatch,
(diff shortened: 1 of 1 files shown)


erikras:
All you've done is change the order? How does that help? Your wrapped dispatch isn't going to be used by the bound action creators, right?

quicksnap:
Because the order is different, `...mapDispatchToProps` will override the default `dispatch` with the one API users provide, if present.

quicksnap:
Any thoughts on this? It works splendidly =P

erikras:
Just groked it. :+1:

quicksnap:
Wazoo! Thanks!

erikras:
Released as [`v4.1.2`](https://github.com/erikras/redux-form/releases/tag/v4.1.2).

quicksnap:
Oh sweet, thanks for the update.

```diff
--- a/client-next/package.json
+++ b/client-next/package.json
@@ -153,7 +153,7 @@
     "redux-devtools": "3.0.1",
     "redux-devtools-dock-monitor": "1.0.1",
     "redux-devtools-log-monitor": "1.0.1",
-    "redux-form": "https://github.com/quicksnap/redux-form/archive/override_dispatch_build.tar.gz",
+    "redux-form": "4.1.2",
     "redux-router": "1.0.0-beta3",
```

Issue #498, opened 2016-01-04
getValues (getValuesFromState) overrides field value for array of objects with `value` key.


tyfoo:
When using `getValues` ( `getValuesFromState`) on with an array of objects such as:
```jsx
const fields = [
    'filters[].key',
    'filters[].value',
    'filters[].flags',
]; 
```
each element in the `filters` array will only have the value of `.value` instead of the object. for example
```jsx
const state = {
    filters: [
        {key: 'k1', value: 'v1' },
        {key: 'k2', value: 'v2' },
    ],
};
const values = getValuesFromState(state);
// values = { filters: [ { value: 'v1' } , { value: 'v2' } ] }
```

PR #499, opened 2016-01-04
Running all tests + fixed getValuesFromState

### LINKED ISSUE #497: `npm run test` does not execute all tests
When running `npm run tests` not all tests (*.spec.js) are executed (node v0.12.4, npm v2.11.2).

I believe the issue is happening because of the tests path in https://github.com/erikras/redux-form/blob/master/package.json#L19. I think that should be:
```jsx
"test": "npm run lint && mocha --compilers js:babel/register --recursive src/__tests__/* --require src/__tests__/setup.js"
```

tyfoo:
Fixes #497 and #498

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "clean": "rimraf dist lib",
     "lint": "eslint src",
     "prepublish": "npm run test && npm run clean && npm run build",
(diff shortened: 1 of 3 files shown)


erikras:
I've added your tests. Thanks for those!

We've already got a way to determine if a value in the state tree is a field value or not, so I'll make it use that.

tyfoo:
@erikras Sorry, I was unaware of that. Thank for the quick feedback and for resolving this issue so quickly

PR #525, opened 2016-01-09
Pass along mergeProps and options to connect-call

### LINKED ISSUE #326: Support for new options in react-redux connect
With the new version of react-redux, it seems the API has changed a bit, and there are new options in `connect`

https://github.com/rackt/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options

Has redux-form been updated to support these?

SimenB:
Fixes #326 

Didn't add a test as we're not doing anything with them other than pass them along

Docs PR: #526

### DIFF
--- a/src/createHigherOrderComponent.js
+++ b/src/createHigherOrderComponent.js
@@ -22,7 +22,9 @@ const createHigherOrderComponent = (config,
                                     connect,
                                     WrappedComponent,
                                     mapStateToProps,
-                                    mapDispatchToProps) => {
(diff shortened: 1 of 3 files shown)


erikras:
:+1: 

A test still might be useful to detect a breaking change in the 3rd party lib, but...

erikras:
Released as [`v4.1.2`](https://github.com/erikras/redux-form/releases/tag/v4.1.2).

PR #526, opened 2016-01-09
Document changes in #525

SimenB:


erikras:
Thanks for taking the time to document it. Most PR submitters don't. :+1:

Issue #517, opened 2016-01-07
Browser support?


gabrielpugliese-luizalabs:
I'm using reduxForm() on my app and it does not work on IE8 even with es3ify and all babel es3 transforms (I've tested removing it and it works like a charm). Are there any plans to have official browser support and maybe IE8 support?

erikras:
Can you be more specific about "does not work"?

gabrielpugliese-luizalabs:
Yes, sure. It is really hard to debug in IE8, specially on a transpiled code. The error thrown is:
`Exception thrown and not caught`. It occurs when I just add the decorator in the code below:
```
const React = require('react');
const { Component, PropTypes } = React;
const { reduxForm } = require('redux-form');
const { connect } = require('react-redux');

const Form = require('../components/Form');

const productActions = require('../actions/product');

const { func, shape, bool, object } = PropTypes;
const validate = values => {
  let errors = {};

  if (! /^[0-9]{7}$/.test(values.id)) {
    errors = Object.assign({}, errors, { id: 'Must be 7 numbers' });
  }

  return errors;
};

@connect(() => ({}), productActions)
@reduxForm({
  form: 'product',
  fields: ['id'],
  validate
})
class FormContainer extends Component {
  static propTypes = {
    getProduct: func.isRequired,
    fields: shape({ id: object.isRequired }).isRequired,
    handleSubmit: func.isRequired,
    submitting: bool.isRequired
  };

  render() {
    const {
      getProduct,
      fields: { id: idField },
      handleSubmit,
      submitting
    } = this.props;
    const childrenProps = { idField, submitting };

    return (
      <Form
        onSubmit={handleSubmit(({ id }) => getProduct(id))}
        {...childrenProps}
      />
    );
  }
}

module.exports = FormContainer;
```

This is my webpack.config.js
```
const path = require('path');
const bourbon = require('node-bourbon').includePaths;

module.exports = {
  entry: [
    'babel-polyfill',
    'es5-shim',
    'es5-shim/es5-sham',
    'console-polyfill',
    './index'
  ],
  resolve: {
    root: [
      path.join(__dirname, 'node_modules')
    ],
    extensions: ['', '.js', '.jsx', '.json']
  },
  output: {
    path: path.join(__dirname, 'dist'),
    filename: 'app.js'
  },
  module: {
    loaders: [
      { test: /\.json$/, loader: 'json' },
[...]

PR #558, opened 2016-01-16
Update react-lazy-cache to not use getters

SimenB:
IE8: 
image

image

You mentioned tests in #517, not sure what sort of tests you want? The existing tests all pass with this change, and the lib used for caching is an implementation detail not to be tested, right?
EDIT: If you meant test it as in functional test in the browser, the example from @gabrielpugliese-luizalabs runs with this change
My change to his example was just to show the validation

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "deep-equal": "^1.0.1",
     "hoist-non-react-statics": "^1.0.3",
     "is-promise": "^2.1.0",
-    "react-lazy-cache": "^3.0.0"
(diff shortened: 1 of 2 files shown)


gabrielhpugliese:
Wow, that's awesome! :clap:

erikras:
Good enough! :+1:

SimenB:
Great, thanks!
We're in discussion with the powers that be at work to drop IE8-support. Would make our lives much easier, not having to use these stupid workarounds... Great that you keep humoring us!

Issue #559, opened 2016-01-16
defaultValue and initialValue not updated when field value changes to a falsey value


j-clark:
Not sure if this is an intentional design decision or a bug. The source of it is [here](https://github.com/erikras/redux-form/blob/master/src/updateField.js#L46-L50).

This manifests as a problem in the following way:
* I have a form that can be switched between editing and read-only state
* Once the form is saved successfully, it toggles from editing to read-only
* When you switch back to editing mode, the input mounts and if it gets `undefined` as a value, it'll fall back to `defaultValue`
* Therefore if you're editing a form in order to remove a value, after switching back to edit mode it'll read the `defaultValue`, which was really the old value and display that.

I'm assuming this isn't intended behavior since if I'm changing the field's value from `1` to `2`, the default and initial values get updated to `2`, but if I'm changing the field's value from `1` to `undefined` (or really `1` to `0` based on the code), it won't update the default and initial values.

Thoughts?

PR #560, opened 2016-01-16
Allow initial value to change to falsey value

j-clark:
Fixes: https://github.com/erikras/redux-form/issues/559

This allows `initialValue` and `defaultValue` to be changed to falsey values, but prevents updates when there is no `initial` key on the field


erikras:
`in`teresting. I suspect that I don't use the `in` operator as much as I should.

Excellent issue reporting and PR. Thanks!

erikras:
Published as [`v4.1.4`](https://github.com/erikras/redux-form/releases/tag/v4.1.4). Thanks again!

Issue #393, opened 2015-12-08
ResetForm function does not clear all values


ala85:
This can be reproduced here : http://erikras.github.io/redux-form/#/examples/simple?_k=4lzvyx

Fill the Favorite Color with an option and click 'Clear Values',: The option will still be visible.

Calling a forceUpdate() wont work.

j-clark:
I'm able to reproduce this for fields that are being normalized. If they're being normalized, when the values get to the `resetForm` function, they're missing the `_isFieldValue` flag, but if I remove the normalizer completely, the flag is set and `resetForm` works

erikras:
The original issue here was solved by doing this, but #562 is still a good PR.

![redux form 2016-02-11 08-42-02](https://cloud.githubusercontent.com/assets/4396759/12977862/61f9f2e0-d09b-11e5-8876-4112770bd3a4.jpg)

PR #562, opened 2016-01-18
Preserve non-enumerable fields when normalizing

j-clark:
Potentially fixes https://github.com/erikras/redux-form/issues/393

I don't know where the code for the examples is, so I can't be sure if this is the source of the problem that was originally reported, but this fixes the problem that I was having, which is that fields which use a normalizer are unable to be reset with `resetForm`

The source of the issue here is that `_isFieldValue` is non enumerable, so the spread isn't copying it when computing the final object. However, `_isFieldValue` is needed for the reducer to handle the `RESET` action correctly.

There are a couple different approaches I can think of to handle this and this is probably the least clean, but in `fieldValues.makeField`, it's explicitly using `defineProperty`, which I can only assume is to make `_isFieldValue` non enumerable for some reason. Sure enough, making it enumerable breaks a bunch of tests.

I also couldn't figure out a reasonable way to test this, so I'm open to feedback on how I can improve this PR

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -239,14 +239,15 @@ function decorate(target) {
             };
             return {
               ...formResult,
-              ...mapValues(formNormalizers, (fieldNormalizer, field) => ({
-                ...formResult[field],
-                value: makeFieldValue(fieldNormalizer(
+              ...mapValues(formNormalizers, (fieldNormalizer, field) => {
+                const newValue = makeFieldValue(fieldNormalizer(
                   formResult[field] ? formResult[field].value : undefined,         // value
                   previous && previous[field] ? previous[field].value : undefined, // previous value
                   getValuesFromState(formResult),                                  // all field values
-                  previousValues))                                                 // all previous field values
-              }))
+                  previousValues));                                                // all previous field values
+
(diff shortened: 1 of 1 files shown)


erikras:
The code for the examples is hiding in the [`gh-pages`](https://github.com/erikras/redux-form/tree/gh-pages) branch. I will have more time to look at this tomorrow, but if you'd like to do more testing today...

If I understand you correctly, you're saying that
```js
...formResult[field]
```
is striping out the `_isFieldValue` flag, but that
```js
Object.assign(formResult[field] || {}, { value: newValue })
```
preserves it? Is that the main change here?

j-clark:
```javascript
{
  ...formResult[field],
  ...anythingElse
}
```

Strips out the non-enumerable fields because it compiles to this:

```javascript
Object.assign({}, formResult[field], anythingElse)
```

The change makes `formResult[field]` the destination object, which doesn't need copying

erikras:
Yep. That was my understanding. Nice work. :+1:

erikras:
Published as [`v4.1.10`](https://github.com/erikras/redux-form/releases/tag/v4.1.10). Thanks!

PR #589, opened 2016-01-26
Disallow async validation when the form is pristine

cwmanning:
This fix addresses an issue where forms with `asyncValidate` and `asyncBlurFields` specified call `asyncValidate` on every blur, even when the form has not been changed from its initial values.

@erikras please let me know if you have any other ideas on how to handle this one.

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -1589,6 +1589,40 @@ describe('createReduxForm', () => {
     });
   });
 
+  it('should disallow async validation when form is pristine', () => {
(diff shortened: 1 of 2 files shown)


erikras:
Hmm... So you're adding the assumption that the `initialValues` would always pass async validation.

That seems okay for most cases, I guess. If you started with a blank form, and one of the fields was required, you would handle that with sync validation, not async.

erikras:
Published in [`v4.1.6`](https://github.com/erikras/redux-form/releases/tag/v4.1.6). Thanks again.

Issue #148, opened 2015-10-18
async validate doesn't fail on general errors


mschipperheyn:
asyncValidate seems to always resolve in case of general exceptions. On redux-form 2.4.1, the form contains two fields: username and password:

	function asyncValidate(data, dispatch){

	  return new Promise((resolve,reject) => {
		const errors = {};

		dispatch(loginUser(data.username, data.password))
		.then(json => {

		  if(Object.keys(json.errors).length > 0){
			dispatch(showFormException());
			reject(json.errors); <= this stops async validation
		  }else{
			dispatch(saveUser(json.user));
			resolve(errors);
		  }

		})
		.catch(err => {
                  //perhaps the remote server is down. Something went wrong
		  reject(); <= after this, handleSubmit is called
		});
	  });
	}

Or

	function asyncValidate(data, dispatch){

	  return new Promise((resolve,reject) => {
		const errors = {};

		dispatch(loginUser(data.username, data.password))
		.then(json => {

		  if(Object.keys(json.errors).length > 0){
			dispatch(showFormException());
			reject(json.errors);
		  }else{
			dispatch(saveUser(json.user));
			resolve(errors);
		  }

		})
		.catch(err => {
                  //perhaps the remote server is down. Something went wrong
		  resolve({valid:false}); <= after this, handleSubmit is called
		});
	  });
	}

but this works

	function asyncValidate(data, dispatch){

	  return new Promise((resolve,reject) => {
		const errors = {};

		dispatch(loginUser(data.username, data.password))
		.then(json => {

		  if(Object.keys(json.errors).length > 0){
			dispatch(showFormException());
			reject(json.errors);
		  }else{
			dispatch(saveUser(json.user));
			resolve(errors);
		  }

		})
		.catch(err => {
                  //perhaps the remote server is down. Something went wrong
		  reject({password:"bla"}); <= async validation fails successfully
		});
	  });
	}

So, it would seem that for async validation to fail, at least one of the form fields needs to be part of the error object. I would argue that any call to "reject" should lead to async validation to fail.

mschipperheyn:
What are your thoughts on this?

erikras:
> So, it would seem that for async validation to fail, at least one of the form fields needs to be part of the error object. I would argue that any call to "reject" should lead to async validation to fail.

I agree.

mschipperheyn:
Fixed by commit 304c5170fcb15183886e371ffb5d6087ae49316e

PR #602, opened 2016-01-27
fixes #596 - handle unexpected/unhandled async error

### LINKED ISSUE #596: async submit errors are swallowed
hello:

```js
    submit(values, dispatch) {
        const { project } = this.props;
        return new Promise((res, rej) => {
            dispatch(saveProject(project), (err, rslt) => {
                if (err) {
                    return rej({
                        name: err.message
                    });
                }
                res();
            });
        });
    }
```

in the above, `saveProject` had a critical error in it, causing a sync `throw`.  my expectation was that my `uncaughtRejection` handler would catch it.  instead, the error was swallowed because i didn't have  [`returnRejectedPromise` enabled](https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L18).  this seems risky, and in turn made it difficult to debug.

maybe we can come up with a more robust solution together?  thanks for making this lib! :)

cdaringe:
# problem statement
if an unexpected application error occurs in `handleSubmit`, it will swallow it.  this is because handleSubmit _expects_ an error filled with error messages for respective fields.  however, async actions can be complicated.  400s, 500s, bad application code, etc can all be at play, and simply get eaten up because of thoughtful, but risky strategy of using all rejections as mechanisms to put `error`s strictly into form state.

for me, this meant meaningful information about my failing application was hidden from me.

# solution
distinguish between expected errors and unexpected errors.

my proposed strategy is to wrap all submission/validation errors in a `SubmissionError`.  if any other error occurs, the rejection will flow to the users `.catch` or `.on('uncaughtRejection', ...)` handler.

in this regard, we can safely know to update the state, or explode because our app died.  if the reviewer finds such a strategy agreeable, it's well known that this is a major bump and requires documentation updates. i'd imagine some folks may really not like having to type all 15 characters of `SubmissionError`, but I think it's well worth it.  Alternatives may be a `cb(errorFields)` strategy, but no matter what, separating good errors from bad errors is a major.  This is actually a pretty low impact, code-wise!

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -8,6 +8,7 @@ import {combineReducers, createStore} from 'redux';
 import {Provider} from 'react-redux';
 import reducer from '../reducer';
 import createReduxForm from '../createReduxForm';
+import SubmissionError from '../submissionError';
 
 const createRestorableSpy = (fn) => {
   return createSpy(fn, function restore() {
@@ -1696,7 +1697,7 @@ describe('createReduxForm', () => {
       form,
       fields: ['name'],
       initialValues: {name: 'Tom'},
-      onSubmit: () => Promise.reject({name: deepError})
+      onSubmit: () => Promise.reject(new SubmissionError({name: deepError}))
     })(Form);
     const dom = TestUtils.renderIntoDocument(
       <Provider store={store}>
--- a/src/__tests__/handleSubmit.spec.js
+++ b/src/__tests__/handleSubmit.spec.js
@@ -1,6 +1,7 @@
 import expect, {createSpy} from 'expect';
 import isPromise from 'is-promise';
 import handleSubmit from '../handleSubmit';
+import SubmissionError from '../submissionError';
 
 describe('handleSubmit', () => {
 
@@ -97,13 +98,14 @@ describe('handleSubmit', () => {
   it('should not submit if async validation fails', () => {
     const values = {foo: 'bar', baz: 42};
     const fields = ['foo', 'baz'];
+    const validationError = new SubmissionError({_error: 'async validation failed'});
     const submit = createSpy().andReturn(69);
     const dispatch = () => null;
     const touch = createSpy();
     const startSubmit = createSpy();
     const stopSubmit = createSpy();
     const submitFailed = createSpy();
-    const asyncValidate = createSpy().andReturn(Promise.reject());
+    const asyncValidate = createSpy().andReturn(Promise.reject(validationError));
     const validate = createSpy().andReturn({});
     const props = {dispatch, fields, startSubmit, stopSubmit, submitFailed, touch, validate};
 
@@ -160,6 +162,29 @@ describe('handleSubmit', () => {
         expect(startSubmit).toNotHaveBeenCalled();
         expect(stopSubmit).toNotHaveBeenCalled();
         expect(submitFailed).toHaveBeenCalled();
+      });
+  });
+
+  it('should return `reject` on async validation if unhandledException/unhandledRejection occurs', () => {
+    const values = {foo: 'bar'};
+    const fields = ['foo'];
+    const applicationError = new ReferenceError('applicationError');
+    const submit = createSpy().andReturn(69);
+    const dispatch = () => null;
+    const touch = () => null;
+    const startSubmit = () => null;
+    const stopSubmit = () => null;
+    const submitFailed = () => null;
+    const asyncValidate = createSpy().andReturn(Promise.reject(applicationError));
+    const validate = createSpy().andReturn({});
+    const props = {dispatch, fields, startSubmit, stopSubmit, submitFailed, touch, validate,
+      returnRejectedSubmitPromise: false};
+
+    return handleSubmit(submit, values, props, asyncValidate)
+      .then(() => {
+        expect(false).toBe(true); // should not get into resolve branch
+      }, err => {
+        expect(err).toBe(applicationError);
       });
   });
 
@@ -240,8 +265,8 @@ describe('handleSubmit', () => {
   it('should set submit errors if async submit fails', () => {
     const values = {foo: 'bar', baz: 42};
     const fields = ['foo', 'baz'];
-    const submitErrors = {foo: 'error'};
-    const submit = createSpy().andReturn(Promise.reject(submitErrors));
+    const submitError = new SubmissionError({foo: 'error'});
+    const submit = createSpy().andReturn(Promise.reject(submitError));
     const dispatch = () => null;
     const touch = createSpy();
     const startSubmit = createSpy();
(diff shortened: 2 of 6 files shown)


VinSpee:
👍🏼 makes sense to me. I have run into the problem of distinguishing these errors and end up reading into them and handling. I'd be nice to have an over-arching category

quicksnap:
This is a worthy breaking change. It will cause users to more explicitly define their error states and avoid painful error swallowing. `SubmissionError

If users want the old behavior, they may easily create a helper method.

Let's draw out the two approaches:

### `SubmissionError`
```js
import {reduxForm, SubmissionError} from 'redux-form';

const submit = (values, dispatch) => {
    setTimeout(() => {
      if (!['john', 'paul', 'george', 'ringo'].includes(values.username)) {
        throw new SubmissionError({username: 'User does not exist', _error: 'Login failed!'});
      } else if (values.password !== 'redux-form') {
        throw new SubmissionError({password: 'Wrong password', _error: 'Login failed!'});
      } else {
        dispatch(showResults(values));
        resolve();
      }
    }, 1000); // simulate server latency
  });
};
```

### `setErrors(errorFields)`
```js
import {reduxForm} from 'redux-form';

const submit = (values, dispatch, setErrors) => {
    setTimeout(() => {
      if (!['john', 'paul', 'george', 'ringo'].includes(values.username)) {
        setErrors({username: 'User does not exist', _error: 'Login failed!'});
      } else if (values.password !== 'redux-form') {
        setErrors({password: 'Wrong password', _error: 'Login failed!'});
      } else {
        dispatch(showResults(values));
        resolve();
      }
    }, 1000); // simulate server latency
  });
};
```

Do those both look correct? Which one is better? 

I slightly prefer the setErrors method, but am on the fence.

quicksnap:
The former, throwing `SubmissionError`, allows for code outside the callback to hook into redux-form error system. However, using `SubmissionError` somewhere outside of the submit method or a helper, such as an action creator, seems out of place and maybe a bad pattern.

cdaringe:
Thanks for the input.  Ive been wrestling with these tjoughts as well.  As a counter to the last bit, if you're in an action Creator and you're already formatting your errors into the redux form format, you are already concerned with redux form anyway. At least now, its explicit and communicative about it.  What I also like about using submission error is that it follows the natural control flow, then redirects the control internally within r-form.

Either work.  Let's see what Mr man @erikras has to say on the matter

erikras:
First of all, I love this PR and this discussion. Very thoughtful stuff.

For some history on the matter, the old way, [prior to `v3.0`](https://github.com/erikras/redux-form/releases/tag/v3.0.0-beta-1), was to `resolve(submissionValidationErrors)`, but then #148 talked me into making them be returned with rejection.

Looking at the code @quicksnap laid out above, I'm starting to doubt the use of promises entirely. We're trying to fit three options into a binary-shaped hole, scurrying back to OOP to solve our FP problem.

What about something like this?

### `Callbacks`
```js
import {reduxForm} from 'redux-form';

const submit = (values, dispatch, callback) => {
  api.post('/saveWidget') // most ajax stuff is promised based, but it doesn't have to be
    .then(result => {
      // developer decides how to distinguish success from validation error
      if (result.success) {
        callback({
          result: result.anyDataTheServerWantsToPassBack
        });
        // could also just do callback() if didn't care to return data
      } else {
        callback({
          validationError: result
        })
      }
    })
    .catch(error => {
      callback({
        transmissionError: error
      });
    });
};
```
`callback()` could also take three variables for the three possibilities, but I prefer them to be keys in an object, because:

* More explicit
* Easier to expand later if we decide we need a fourth state
* `callback(null, null, thirdParam)` looks dumb and makes it easier to write buggy code

What to call the keys is up for debate, as "naming things" is one of the two hardest problems in Computer Science.

One thing that we should keep in mind, is that it would be nice if we could use the same syntax across all three types of validation: sync, async, and submission. For example, if we went with the key names in my above example, a sync validation function might look like:

```js
const validate = (values) => {
  if (!values.firstName) {
    return {
      validationError: {
        firstName: 'Required'
      }
    };
  }
  return {}; // or just undefined might do as well?
}
```

^^ That might be an argument in favor of the keys being `result`, `error`, and `submissionError`.

And finally.... we should consider the `handleSubmit()` API. How about it still returns a promise, and:

State | Result
--- | ---
Successful submission | Promise is resolved
Validation errors | The promise is never resolved or rejected; wait for your errors to show up as props on the next process tick.
Submission error, something with wrong in the I/O to the server or the response came back as anything but `200 OK` | Promise is rejected

Let's continue this discussion...

cdaringe:
hey all!  thanks for the continued effort :).  BTW, i'm not firmly attached to my PR--i'm much more interested in coming out with the right design, so no worries if that was a concern.

> We're trying to fit three options into a binary-shaped hole

right.  respectfully, IMHO, the your proposed strategy still suffers the same issue.  promise api, cb api--same thing. ok ok, there are some differences, but the existence of `.promisify` or `.nodeify` are testaments to the claim :)

consider the following:

- returning a resolved promise as a means to set form state can be dicey because of namespace conflicts. e.g. `return promiseXhr.get(url)`.  also, if the user chained on to the call, it's misleading that all was well to them.
- using a cb strategy doesn't change the problem.  `return xhr.get(..., cb)`.  the handler will still have an empty `err` first argument.
- returning a rejected promise doesn't work, because of the original intent of the PR--meaningful errors are swallowed
- using a cb strategy in the error field doesn't change the problem.  `myBrokenAsyncCall(..., cb)`.  same as the rejected promise.

i think the **problem lies in that we are trying to detect form error state implicitly** through a control flow which is susceptible to transmitting data of unknown types and content. 

if we pass in a separate method entirely to set form error state (a bound/unbound actionCreator perhaps), then all of reasoning around whether or not the operation was successfully/unsuccessful goes away--the user explicitly declares/defines it.

> Validation errors: The promise is never resolved or rejected; wait for your errors to show up as props on the next process tick.

this reduces usability, and arguably introduces non-determinism.


i'd like to chase something like

```js
const validate = (values, setErrors) => {
  if (!values.firstName) {
    return setErrors({
        firstName: 'Required'
    });
}
```

if it's possible.

what do you think?

erikras:
Interesting Twitter discussion. https://twitter.com/dan_abramov/status/694128907805691905

cdaringe:
it caused quite a stir didn't it

erikras:
I have [gone with `SubmissionError` in `v6`](http://redux-form.com/6.0.0-alpha.6/examples/submitValidation/), as suggested by this PR. Closing this for now. Thanks for the great discussion.

bmatto:
Howdy folks - 

This is causing issues with our testing setup in regards to the promise never being rejected.  Wondering if anyone has a working example of how to test errors using this pattern.

danield-hyprbrands:
@bmatto check out the section where this PR is referenced 
http://redux-form.com/6.3.1/docs/MigrationGuide.md/

danield-hyprbrands:
@erikras IMHO submission errors are too coupled to redux-form.
So far my actions and API calls are oblivious to their part in a chain of promises, forcing me to add specific code to wrap errors is limiting.
So migrating from v5 to v6 is not as clean as I would of hoped in that sense.

PR #608, opened 2016-01-31
Fix for react-native 0.18+ with react-redux 4+

xiaobuu:
redux-form/native depends on react-redux/native which does not exist in react-redux 4+.
So the original redux-form has to be used instead. But it passes isNative as false, so most functions won't work.
A detection for react-native environment in index.js fixes the issue.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,8 @@ import React from 'react';
 import {connect} from 'react-redux';
 import createAll from './createAll';
 
+const isNative = window.navigator && window.navigator.product && window.navigator.product === 'ReactNative';
+
 export const {
   actionTypes,
   addArrayValue,
@@ -27,4 +29,4 @@ export const {
   touchWithKey,
   untouch,
   untouchWithKey
-} = createAll(false, React, connect);
+} = createAll(isNative, React, connect);


src/index.js
```diff
@@ -27,4 +29,4 @@ export const {
   touchWithKey,
   untouch,
   untouchWithKey
-} = createAll(false, React, connect);
```
mattiamanzati: Leave a line note

erikras:
Does this mean that the `redux-form/native` endpoint can be retired?

xiaobuu:
@erikras It still needs to exist until everyone is switched to react-redux 4+

erikras:
Fix published as [`v4.1.7`](https://github.com/erikras/redux-form/releases/tag/v4.1.7).

mschipperheyn:
I'm upgrading an older project (now on redux-form 4.2.2, react-redux 4.4.0). I get Unable to find the module react-redux/native in redux-form. Am I missing something?

mattiamanzati:
At the moment you can manually modify the .js file in node_modules/redux-form/lib/native.js and where requiring the react-redux/native module, remove the /native module at the end. We need a fix about that.

erikras:
Can you PR this, @mattiamanzati? I'm not 100% sure I understand.

Issue #580, opened 2016-01-22
Can not get promise with calling handleSubmit


haradakunihiko:
I think it's related to #538. 
I'd like check the submit sync validation result so that user is alerted, but returnRejectedSubmitPromise does not work for me.

Maybe it's because I does not pass onSubmit props to redux-form-connected Component. 

Before, my code was like bellow.
```js

MyComponent extends Component {

  handleSave(values) {
    // do something.
  }

  render() {
    return <button onClick={this.handleSubmit(::this.handleSave)}></button>
  }
}
```

and I changed this so that I can check the validation result using Promise.
```js

MyComponent extends Component {

  handleSave(values) {
    // do something.
    return Promise.resolve();
  }

  handleOnClick(e){
    this.props.handleSubmit(::this.handleSave)(e) // => this does not return a promise.
    .then(() => {
    }, (errors) => {
      // => I'd like to get sync errors!
      showAlertSothatYouCanKnowWhatsGoingOn(errors );
    });
  }

  render() {
    return <button onClick={::this.handleOnClick}></button>
  }
}

```

I guess returning the result in the callback function would work. 
https://github.com/erikras/redux-form/blob/master/src/createHigherOrderComponent.js#L92

Also, is that possible to return syncErrors with Promise when it is rejected?
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L37

How do you think about this?

kristian-puccio:
I just ran into this as well. I think what is going on is that a promise is being return but only when the validation fails and null when validation passes.

So you get a for me `Uncaught TypeError: Cannot read property 'catch' of undefined` error on success.

kristian-puccio:
work around: 
```
		const x = handleSubmit(e);
		if (x) {
			x.catch(() => this.props.onSubmitFailed(this.props.errors));
		}
```

erikras:
@haradakunihiko's fix published as [`v4.1.10`](https://github.com/erikras/redux-form/releases/tag/v4.1.10).

erikras:
Please confirm that `v4.1.10` fixes this problem, @kristian-puccio.

kristian-puccio:
No unfortunately :(

I looked quickly at the pull request and it looks like it should but I'm
still getting undefined error.

```
<form
autoComplete={false}
onSubmit={ this._handleSubmit }>
```

And here is  where I handle the submit, x print out as undefined when there
are no async errors.
[...]

PR #618, opened 2016-02-04
Return promise when callback function is passed to handleSubmit

haradakunihiko:
fix for #580 
I hope there is no problem how it should be!

I'm not very sure if other Promise.reject in handleSubmit.js should return rejected promise with the validation errors. 
To be more specific,  here. https://github.com/haradakunihiko/redux-form/blob/master/src/handleSubmit.js#L30


erikras:
Good PR. It makes total sense for the sync errors to be passed in the rejected promise.

erikras:
Published as [`v4.1.10`](https://github.com/erikras/redux-form/releases/tag/v4.1.10). Thanks!

PR #625, opened 2016-02-07
Make spinner pretty

msafi:
The current `Loading...` text on the website doesn't look good.

This makes it pretty, like [this one](https://msafi.github.io/string-pattern/), which is copied from the bottom right spinner [here](http://projects.lukehaas.me/css-loaders/).

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,7 +22,56 @@
   <meta property="twitter:creator" content="@erikras"/>
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


msafi:
Either this or delete the `Loading...` text and just show a white page. That would be less jarring :)

erikras:
Ha! Great. :+1:

Issue #516, opened 2016-01-07
Normalize nested fields


joshgagnon:
Trying to combine nested fields with normalization, but can't see the intended way to do this.

I would have thought it would similar to:
```
const form = formReducer.normalize({
    'form': {
        'parent.child': someNormalizeFunction
    }
})...
```

Any suggestions?  Do I need to implement this?

erikras:
I've never done that. I'll investigate and get back to you. :mag:

erikras:
Yeah. The answer is that normalization of deep fields is currently not supported. :disappointed:

Misaka42:
What about normalizing forms with formKey?
It seems that normalizer it attempting to set a global field on form with no regard to formKey.

erikras:
@Misaka42 Normalizing with `formKey` is supposed to work. Someone PR'd a test for it, at least. :smile: As long as your subform is flat, I'd expect it to work.

joshgagnon:
Normalizing with formKeys works fine for me (warning: I submitted the PR).  Hopefully I find some time to implement nested normalization.  

If I do have a stab at it, would you prefer style in the example given in the parent comment?

Misaka42:
Accidentally submitted the issue anyway, and I can't close it for some reason.
Well, my form is not flat, but should that really matter?
If I, say, instead of trying normalizing a nested FIELD, normalize the whole nested object as I suggested previously, would that work?
Currently though, redux form just creates and empty field in form's root, as in the same level as form keys. Of course it's empty and of course it's not being normalizing.

erikras:
@joshgagnon Indeed you are! My initial instinct is that for normalizing, for example, a phone number in:
```js
{
  billing: {
    phone: '555-555-5555'
  },
  shipping: {
    phone: '555-555-5555'
  }
}
```
It would be nice to nest the function:
```js
reducer.normalize({
  contactForm: {
    billing: {
       phone: normalizePhone
    },
    shipping: {
       phone: normalizePhone
    }
  }
})
```
But that won't let you do array values, e.g. `address.phones[]`, so maybe the string key is better. i.e.
```js
reducer.normalize({
  contactForm: {
    'billing.phone': normalizePhone
    'shipping.phone': normalizePhone
  }
})
```
Needs investigation and trial and error iteration...

[...]

PR #637, opened 2016-02-11
Normalize nested fields

jfairbank:
This is my first hacky stab at fixing #516 after I realized I needed this functionality yesterday :)

The code has some repetitiveness and isn't the cleanest at the moment. All current tests pass, and I added some minimal nested fields to existing tests to ensure they work. Regardless, we might want some more tests. I also made sure to replicate the work done in #562 recently to preserve non-enumerable keys, but we might want tests for that as well.

I took the approach of nested normalizers instead of using the nested key sytax (e.g. `foo[].bar`). I found it easier to walk down the state and normalizer objects at the same time with this approach. Using the nested key syntax was a little more difficult.

For an idea of the syntax here is a test I changed:

```js
it('should normalize simple form values', () => {
  const defaultFields = {
    _active: undefined,
    _asyncValidating: false,
    [globalErrorKey]: undefined,
    _initialized: false,
    _submitting: false,
    _submitFailed: false
  };
  const normalize = reducer.normalize({
    foo: {
      name: () => 'normalized',
      person: {
        name: (name) => name && name.toUpperCase()
      },
      'pets[]': {
        name: (name) => name && name.toLowerCase()
      }
    }
  });
  const state = normalize({
    foo: {
      name: {
        value: 'dog'
      },
      person: {
        name: {
          value: 'John Doe',
        }
      },
      pets: [
        { name: { value: 'Fido' } },
        { name: { value: 'Tucker' } }
      ]
    }
  });
  expect(state)
    .toExist()
    .toBeA('object');
  expect(state.foo)
    .toExist()
    .toBeA('object')
    .toEqual({
      ...defaultFields,
      name: {
        value: 'normalized'
      },
      person: {
        name: {
          value: 'JOHN DOE'
        }
      },
      pets: [
        { name: { value: 'fido' } },
        { name: { value: 'tucker' } }
      ]
    });
});
```

One thing I'm still figuring out is how to handle this array situation:

```js
const normalize = reducer.normalize({
  // Array contains string field values
  'hobbies[]': (value) => value && value.toUpperCase()

  // Or array IS the field value
  'hobbies[]': (array) => array && array.map(hobby => hobby.toUpperCase())

  // Or array contains nested fields
  'people[]': (person) => {
    if (person) {
      return { name: person.name.toUpperCase() };
    }
    return { name: '' };
  }
});
```

Basically, if you don't specify nested properties, do you assume the array is a field value or that it still contains nested field values? What happens if the nested values have nested properties of their own? What would you expect to be the use case for this syntax? Or don't support it? If those questions don't make sense, let me know. Right now, I just have it assume that the array itself is the field value if you don't specify nested properties.

Please let me know if this is what you would like to see for normalizing nested fields and what I need to modify with this implementation.

### DIFF
--- a/src/__tests__/reducer.spec.js
+++ b/src/__tests__/reducer.spec.js
@@ -2945,7 +2945,9 @@ describe('reducer', () => {
     it('should initialize form state when there is a normalizer', () => {
       const state = reducer.normalize({
         foo: {
-          myField: () => 'normalized'
+          'myField': () => 'normalized',
+          'person.name': () => 'John Doe',
+          'pets[].name': () => 'Fido'
         }
       })();
       expect(state)
@@ -2964,7 +2966,13 @@ describe('reducer', () => {
           _submitFailed: false,
           myField: {
             value: 'normalized'
-          }
+          },
+          person: {
+            name: {
+              value: 'John Doe'
+            }
+          },
+          pets: []
         });
     });
 
@@ -2979,7 +2987,9 @@ describe('reducer', () => {
       };
       const normalize = reducer.normalize({
         foo: {
-          myField: () => 'normalized'
+          'myField': () => 'normalized',
+          'person.name': () => 'John Doe',
+          'pets[].name': () => 'Fido'
         }
       });
       const state = normalize({
@@ -3006,7 +3016,13 @@ describe('reducer', () => {
             ...defaultFields,
             myField: {
               value: 'normalized'
-            }
+            },
+            person: {
+              name: {
+                value: 'John Doe'
+              }
+            },
+            pets: []
           }
         });
       expect(nextState.foo)
@@ -3015,13 +3031,25 @@ describe('reducer', () => {
             ...defaultFields,
             myField: {
               value: 'normalized'
-            }
+            },
+            person: {
+              name: {
+                value: 'John Doe'
+              }
+            },
+            pets: []
           },
           secondSubForm: {
             ...defaultFields,
             myField: {
               value: 'normalized'
-            }
+            },
+            person: {
+              name: {
+                value: 'John Doe'
+              }
+            },
+            pets: []
           }
         });
     });
@@ -3037,14 +3065,25 @@ describe('reducer', () => {
       };
       const normalize = reducer.normalize({
         foo: {
-          name: () => 'normalized'
+          'name': () => 'normalized',
+          'person.name': (name) => name && name.toUpperCase(),
+          'pets[].name': (name) => name && name.toLowerCase()
         }
       });
       const state = normalize({
         foo: {
           name: {
             value: 'dog'
-          }
+          },
+          person: {
+            name: {
+              value: 'John Doe',
+            }
+          },
+          pets: [
+            { name: { value: 'Fido' } },
+            { name: { value: 'Tucker' } }
+          ]
         }
       });
       expect(state)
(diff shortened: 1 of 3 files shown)


jfairbank:
Woops. I forgot I temporarily disabled linting when I was testing. Let me fix that :disappointed:

erikras:
I think I vote for
```js
'hobbies[]': (array) => array && array.map(hobby => hobby.toUpperCase())
```
It's not hard to wrap your normalizing function in an `array.map()`. And this also allows you to do stuff like limiting array length or sorting or something.

How about this format? So that the keys match up exactly with the `fields` array syntax.
```js
'people[].name': name => name.toUpperCase()
```

jfairbank:
> I think I vote for
> 
> `'hobbies[]': (array) => array && array.map(hobby => hobby.toUpperCase())`

That works for me. Seems least surprising.

> How about this format? So that the keys match up exactly with the fields array syntax.
> 
> `'people[].name': name => name.toUpperCase()`

Last night when I started fiddling with this, that syntax was not as easy to work with, but it may have been I was more tired at the end of my day ;)

The nested syntax which I did implement made it easier for walking down nested state even for arrays:

```js
normalize({
  foo: {
    'people[]': {
      name: (name) => name && name.toUpperCase()
    }
  }
});
```

This doesn't mean the fields syntax `'people[].name'` isn't doable, but I was also worried if I would have to walk the normalizer object more times. For example, if you wanted to specify `'people[].firstName'` and `'people[].lastName'`, then you would re-walk the state from the top. With the nested `'people[]': { firstName: ..., lastName: ... }`, you keep the context of where you are at as you walk down the state. If I'm not explaining that well, let me know.

That being said, I'd be happy to use what I've got to build out support for fields syntax (i.e. `'people[].name'`) because I like that mapping to the original fields syntax better too.

erikras:
There a [many examples](https://github.com/erikras/redux-form/search?utf8=✓&q=openIndex) of tree crawling via the field string. Make sure you test all the possibilities, because, while it's not complicated, it's easy to make mistakes.

jfairbank:
You're right, walking the tree wasn't so bad :sweat_smile:. My mind must've been exhausted or going down the wrong path the other night. I've switched the function over to support the field syntax (e.g. `'people[].name'`) instead. Here is the one test rewritten with the new syntax:

```js
it('should normalize simple form values', () => {
  const defaultFields = {
    _active: undefined,
    _asyncValidating: false,
    [globalErrorKey]: undefined,
    _initialized: false,
    _submitting: false,
    _submitFailed: false
  };
  const normalize = reducer.normalize({
    foo: {
      'name': () => 'normalized',
      'person.name': (name) => name && name.toUpperCase(),
      'pets[].name': (name) => name && name.toLowerCase()
    }
  });
  const state = normalize({
    foo: {
      name: {
        value: 'dog'
      },
      person: {
        name: {
          value: 'John Doe',
        }
      },
      pets: [
        { name: { value: 'Fido' } },
        { name: { value: 'Tucker' } }
      ]
    }
  });
  expect(state)
    .toExist()
    .toBeA('object');
  expect(state.foo)
    .toExist()
    .toBeA('object')
    .toEqual({
      ...defaultFields,
      name: {
        value: 'normalized'
      },
      person: {
        name: {
          value: 'JOHN DOE'
        }
      },
      pets: [
        { name: { value: 'fido' } },
        { name: { value: 'tucker' } }
      ]
    });
});
```

I also added a new test for arbitrarily deeply nested fields, one with and one without arrays. Are there other types of tests you would like to see?

(FYI, I'm piling commits on until we get to a satisfactory merge state and then I'll squash.)

erikras:
Looks pretty good to me! :+1: :shipit:

jfairbank:
Cool. I made one tiny tweak to remove an unnecessary variable, but it's all squashed and ready unless you want any more tests or tweaks.

PR #645, opened 2016-02-13
Handle array normalization properly

jfairbank:
Here's code to handle the array syntax without nested properties (i.e. `'cats[]': (array) => ...`). This should pass the test introduced in dc581172784686af13754c0d07cb9f502eb9fde2. I also added a couple of other tests to ensure nested arrays work similarly.


erikras:
Excellent work. :+1:

PR #646, opened 2016-02-15
added deep modify test

kristian-puccio:
Ok I've added a test.

What is strange is that I'm not seeing the behaviour about mutating of child fields but I am seeing that the parent container field isn't mutating.

I'll look more at a fix, just getting my head around the code.


erikras:
This passes with [`v4.1.15`](https://github.com/erikras/redux-form/releases/tag/v4.1.15).

PR #647, opened 2016-02-15
Normalizing array fields

erikras:
This PR fixes several bugs related to deep field mutations and normalization.

@jfairbank I apologize for [misunderstanding you before](https://github.com/erikras/redux-form/commit/dc581172784686af13754c0d07cb9f502eb9fde2#commitcomment-16075774) (I was on a layover between flights and not thinking straight). You were right. The array values need to be wrapped with `{value: }`.


jfairbank:
No worries :smile:

PR #689, opened 2016-02-26
Failing tests from deep field mutation

stevoland:
Hi @erikras, would you be able to confirm whether the behaviour in these tests is expected?

The 1st new test passes - `fields.addresses[0].postalCode` is a new object after `fields.addresses[0].postalCode.onChange()` is called. All good.

The 2nd fails - `fields.addresses[0]` is the same object. Perhaps this is expected behaviour?

The 3rd also fails - `fields.addresses[0].postalCode` _isn't_ a new object when checked in component lifecycle methods before the subsequent render.

Apologies if this has been raised before. If the 3rd behaviour is a bug, any pointers to how I can help fix it would be appreciated. Thanks!

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -2351,9 +2351,10 @@ describe('createReduxForm', () => {
           return;
         }
 
-        expect(nextProps.fields.addresses[0].postalCode)
-          .toNotBe(this.props.fields.addresses[0].postalCode);
-        done();
+        if (nextProps.fields.addresses[0].postalCode !==
+          this.props.fields.addresses[0].postalCode) {
+          done();
+        }
       }
 
       render() {
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -2245,6 +2245,149 @@ describe('createReduxForm', () => {
     expect(barRenders).toBe(2);
   });
 
+  it('should change the deep field in an array', (done) => {
+    const store = makeStore();
+    const form = 'testForm';
+    class TestForm extends Component {
+      componentWillReceiveProps(nextProps) {
+        if (!this.props.fields.addresses.length) {
+          return;
+        }
+
+        expect(this.props.fields.addresses[0].postalCode)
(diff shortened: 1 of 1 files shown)


stevoland:
Updated the 3rd test to avoid failing if `componentWillReceiveProps` is called for some reason other than the tested field changing.

Don't know why it's passed on Travis - only a few of the tests have been run - not including the ones in this PR.

erikras:
@stevoland I've been looking into this. I wrote a test in `readField` a little closer to the source of the bug. It's a tricky one.

stevoland:
Thanks @erikras, I'll have a look at that and see if I can help.

erikras:
Here's my `readField` test.

```js
 it('should mutate a changed array subvalue', () => {
    const fields = {};
    readField({
      foo: [
        {
          bar: {
            value: 'baz'
          }
        }
      ]
    }, 'foo[].bar', undefined, fields, {}, undefined, false, defaultProps);
    expectField({
      field: fields.foo[0].bar,
      name: 'foo[0].bar',
      value: 'baz',
      dirty: true,
      touched: false,
      visited: false,
      error: undefined,
      initialValue: undefined,
      readonly: false
    });
    const beforeFoo = fields.foo;
    const beforeBar = fields.foo[0].bar;
    readField({
      foo: [
        {
          bar: {
            value: 'baq'
          }
        }
      ]
    }, 'foo[].bar', undefined, fields, {}, undefined, false, defaultProps);
    expectField({
      field: fields.foo[0].bar,
      name: 'foo[0].bar',
      value: 'baq',
      dirty: true,
      touched: false,
      visited: false,
      error: undefined,
      initialValue: undefined,
      readonly: false
    });
    const afterFoo = fields.foo;
    const afterBar = fields.foo[0].bar;
    expect(beforeBar).toNotBe(afterBar);
    expect(beforeFoo).toNotBe(afterFoo);
  });
```

erikras:
I'm gonna close this because it's part of the horrible `fields` mess that made me abandon that strategy when writing `v6`.

ooflorent:
@erikras Should me close all prior v6 related issues since it won't get fixed?

PR #716, opened 2016-03-07
Readability: replace reduce with filter

abroz:
Filter is definitely easier to read here, and probably more idiomatic.


erikras:
Duh. Thanks! :+1:

PR #743, opened 2016-03-16
DO_NOT_MERGE

kentcdodds:
1. Update PayPal donate badge to be more crisp
2. Use `paypal.me` URL which has a much better experience

You'll need to register your own slug, and then you can update this with your own.

Note, you can also provide a default amount like so: https://www.paypal.me/kentcdodds/10


kentcdodds:
Here's a before:

screen shot 2016-03-16 at 3 19 43 pm

screen shot 2016-03-16 at 3 20 17 pm

And after:

screen shot 2016-03-16 at 3 19 33 pm

screen shot 2016-03-16 at 3 20 55 pm

(Though, of course, you'd need to provide your own paypal.me slug :wink:)

erikras:
Thanks, @kentcdodds!

And thanks for apparently increasing my code coverage %. :laughing:

erikras:
I swear that I looked into this before doing it the other way, but I cannot recall why I went the way I did.

PR #744, opened 2016-03-16
Added start task since it's the expected standard.

coryhouse:
I had to reopen the docs to determine which task I should run. Suggest adding a start task that either delegates to dev task or replaces it. Should help avoid confusion for others.


erikras:
Yep. This improvement was already planned for future docs, but didn't make it into current ones. Thanks! :+1:

PR #761, opened 2016-03-23
Updated readme to include a working example

Carnewal:
The "Simple Form Example" link is broken so I changed it to a working one.


erikras:
Thanks for drawing my attention to this.

PR #763, opened 2016-03-25
Cache handleSubmitPassback function

yesmeck:
The original writing will create a new function on every time ConnectedForm renders which leads shouldComponentUpdate not working.


erikras:
Thanks!

PR #799, opened 2016-04-13
[WIP] support FSA

brikou:
This PR is still WIP but can you confirm you still agree with adding FSA support, thx

### DIFF
--- a/src/__tests__/actions.spec.js
+++ b/src/__tests__/actions.spec.js
@@ -8,71 +8,95 @@ describe('actions', () => {
   it('should create add array value action', () => {
     expect(addArrayValue('foo', undefined, 1)).toEqual({
       type: ADD_ARRAY_VALUE,
-      path: 'foo',
-      index: 1,
-      value: undefined,
-      fields: undefined
+      payload: {
+        path: 'foo',
+        index: 1,
+        value: undefined,
+        fields: undefined
+      }
     });
     expect(addArrayValue('bar.baz')).toEqual({
       type: ADD_ARRAY_VALUE,
-      path: 'bar.baz',
-      index: undefined,
-      value: undefined,
-      fields: undefined
+      payload: {
+        path: 'bar.baz',
+        index: undefined,
+        value: undefined,
+        fields: undefined
+      }
     });
     expect(addArrayValue('bar.baz', 'foo', 2)).toEqual({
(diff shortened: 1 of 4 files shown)


erikras:
Two questions:

1. It's just a matter of wrapping everything but `type` in a `payload` object?
2. Why is this superior? Are there middlewares that it enables or something?

brikou:
FSA is sometime expected by generic middlewares, a FSA actions not only wrap everything under payload but also allows an error and a key entry. It is used by famous lib like react router. You can read more here https://github.com/acdlite/flux-standard-action

Keep me in touch to tell me if you would agree to merge this kind of feature. thx

erikras:
I looked into implementing this, but decided to pull back. `redux-form` makes extensive use of [a pattern I devised](https://github.com/erikras/redux-form/blob/master/src/bindActionData.js) to enhance action creators by binding additional information to their return values. That process is _much_ more complicated if FSA is to be adhered to. If you can figure a way around that problem that does not introduce a _lot_ more complexity, I would be happy for this library to follow FSA.

ooflorent:
@erikras Actually such pattern is possible in FSA using `metadata`.

* `payload` contains the intent of the action
* `metadata` contains info about the target (e.g. `form` or `key`)

grrowl:
@ooflorent that should be `meta` ;)

`bindActionData.js` can be rewritten to fold the extra data into the meta key pretty simply:

```js
import mapValues from './mapValues';

/**
 * Adds additional properties to the results of the function or map of functions passed
 */
export default function bindActionData(action, data) {
  if (typeof action === 'function') {
    return (...args) => ({
      ...action(...args),
      meta: {
        ...action.meta,
        ...data
    });
  }
  if (typeof action === 'object') {
    return {
      ...action,
      meta: {
        ...action.meta,
        ...data
      };
    }
  }
  return action;
}
```

erikras:
Good news, for FSA fans. I've seen the light and I've implemented FSA compliance into `v6`. Not released yet, but it will be in next released version.

erikras:
Also, `bindActionData` is being retired. It was cool functional trickery, but it added needless complexity.

erikras:
FSA compliance is now published in [`v6.0.0-alpha-7`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.7).

Issue #776, opened 2016-03-31
onSubmitFailed


MiMaMuh:
Hi,

it would be great if there were a onSubmitFailed prop like the onSubmit one to handle things like shake the submit button when the validation fails ...

I could do it with 'submitting' and see if it changes from true to false, but it would be more convenient with onSumbitFailed ... Thx!

pcgilday:
You can use the `returnRejectedSubmitPromise` flag, and the approach listed in [this](https://github.com/erikras/redux-form/issues/616#issuecomment-209105697) comment to perform any actions when validation fails.

pcgilday:
@erikras Does it make sense to add an example for how to do this, since it seems to be a common question? I am happy to make the example and submit a PR if you believe it warrants one.

MiMaMuh:
Okay, I got it. But I think it is a little bit cumbersom. It would be much easier and semantic of there would be just a onSubmitFailed prop like the onSubmit one. What so you think?

puradox:
@MiMaMuh I agree that an `onSubmitFailed` prop would be a better approach than the current one.

pcgilday:
@MiMaMuh @puradox I would check out the proposal in #726 and see if any of those changes account for the behavior desired.

puradox:
@pcgilday The only proposal that stood out to me, relating this issue, is his section on Form Submission (10 on #726). He doesn't get too in depth on his planned implementation, but it *could* account for our desired behavior. 

While we are waiting for `v6.0`, I might just start a PR in the meantime for `onSubmitFail`, as it seems simple enough to implement.

erikras:
You guys might be interested in [`v5.1.0`](https://github.com/erikras/redux-form/releases/tag/v5.1.0).

Issue #616, opened 2016-02-03
Callback when submit failed?


flyfy1:
What I want to achieve: upon clicking the `submit` button, and when we have error, move the screen to the occurrence of the first error.

Method I can come up with: when the submit fails, use jquery to search for error nodes, and `scrollIntoView` the first node.

Issue I am facing: I cannot find the callback when submit fails, from the ReduxForm api / source code.

flyfy1:
Issue resolved (via a hack).

Here's how I solved it, based on the assumption that:

- I'm rendering all the errors using one component.
- I only have one form in my whole application (this restriction can be eliminated)


Basic Idea: manually trace each of the errors into an `error_registry`; when the form component updated after submit, check if there're any error (based on `error_registry`), and scroll to the top elements.


So here's my code:

### `error_registry`
    var error_registry = {}
    var should_scroll = false;

    export default {
      set_error: function(form_name, element){
        error_registry[form_name] = element
      },

      del_error: function(form_name){
        delete error_registry[form_name]
      },

      scroll_to_top_elment: function(){
        var top_ele = null, top_height = null;
        for(var k in error_registry){
          const v = error_registry[k];
          const box = v.getBoundingClientRect();

          if(top_ele == null){
            top_height = box.top;
            top_ele = v;
          } else if(top_height > v.top){
            top_height = v.top;
            top_ele = v;
          }
        }
        if(should_scroll && top_ele){
          top_ele.scrollIntoView();
          should_scroll = false;
        }
      },

      should_scroll: function(){
        should_scroll = true;
      }
    }


### ErrorField:

    export class ErrorField extends Component{
      render(){
        const {field} = this.props

        if(field){
          return (<div className="error-message">{error_from_field(field)}</div>)
        } else{
          return (<div/>)
        }
      }

      componentDidUpdate(){
        const {field} = this.props;
        const field_name = error_field_name(field);
        const error_msg = error_from_field(field);

        if(!field_name) return;

        if(error_msg) DarkMagic.set_error(field_name, ReactDOM.findDOMNode(this).parentNode);
        else DarkMagic.del_error(field_name);
      }
    }



### Form:

    function submit_wrapper(f){
      return function(){
        DarkMagic.should_scroll();
        f.apply(null, arguments);
      }
    }
    

    render: function(){
    ... some code ...
      <form
        onSubmit={ submit_wrapper(handleSubmit((values, dispatch) => dispatch({type: 'review-form-submit', payload: values, history: this.history})))}
[...]

Issue #488, opened 2015-12-31
Focus field after validation error


yleclanche:
Hello,

After a validation, I try to focus the first field with an error. 
Do you know a proper way to do this ?

Thanks.

erikras:
You have to call `focus()` on the DOM node itself, which means you need a `ref={myField.name}` on all your inputs so that you can later do `this.refs.myField.focus()`. There's no other way, really.

yleclanche:
The validate function is called from handleSubmit so where can I do the
focus ?


2015-12-31 17:02 GMT+01:00 Erik Rasmussen <notifications@github.com>:

erikras:
You could monitor when `submitFailed` flips from `false` to `true` in `componentWillReceiveProps` and do the `focus()` there? But that will only work the first time.

kristian-puccio:
or use returnRejectedSubmitPromise and catch it in the returned promise

erikras:
Yes, @kristian-puccio's idea is better.

yleclanche:
@kristian-puccio solution only works for async validation, right ? Or I am missing something ?
To be honest I don't really understand the `returnRejectedSubmitPromise` property.
handleSubmit already returns a promise, right ?

erikras:
I'm about to implement something similar in my own projects. I'll share how it goes.

erikras:
Followup: I used `returnRejectedSubmitPromise: true` and got back the errors in my promise `catch()` block just fine.

yleclanche:
Can you provide an exemple on how to use it please ?

kristian-puccio:
I can give you some sudo code:
```
<form
autoComplete={false}
onSubmit={ e => {
handleSubmit(e)
.catch(errors => {  // if the errors isn't passed here you can always grab
it from props.errors.
    // errors is map of field name to error string. so you'll need a way to
find the first field in the screen that errored.
  //  have an array of the fields so I do this:

const widget = this.props.widgets.find( w => errors[w.Name] );
// Scroll to it. useing: https://github.com/fisshy/react-scroll
scrollTo(`widget_${this.props.viewerReference}__${widget.Name}`);

})
}}>
```

yleclanche:
Thanks (both of you :))

tomduncalf:
FYI my solution for this looked like:

```
  componentWillReceiveProps(nextProps: Props): void {
    if (this.state.submitted && Object.keys(nextProps.errors).length) {
[...]

Issue #291, opened 2015-11-18
"Unhandled promise rejection" only in 3.0.9 


approots:
My "submitHandler" returns a rejected Promise when the server responds with an error due to an invalid field. When this happens, I see this error in the console "Unhandled promise rejection". I'm still able to display the error message in my form though. The console error doesn't occur in 3.0.8.

Maybe I'm not handling the Promise correctly? Here's my submit handler:

    handleSubmit(values) {
        const postData = {method: 'post',body: {...values}};

        return new Promise((resolve, reject) => {
            return fetch.json('/api/signup', postData)
                .then((json) => {
                    return resolve();
                })
                .catch((error) => {
                    switch (error.json.message) {
                        case 'signup_code_not_exists' :
                            return reject({signupCode: 'The code does not exist'});
                        default :
                            return reject({_error: 'Error'});
                    }
                });
        });
    },

This is the exact error I see:

    Unhandled promise rejection Object {signupCode: "The code does not exist"}

The error occurs in the Babel es6.promise.js file.

erikras:
Aha! _That's why_! https://github.com/erikras/redux-form/issues/256#issuecomment-157192563

erikras:
Fixed in `v3.0.10`.

approots:
Thanks!

ntdb:
@approots @erikras Just found this after a half hour of confusion. Thank you! :+1:

oyeanuj:
@ntdb @approots Did upgrading it resolve it for you? I have run into a similar problem but can't seem to resolve it.

rkbrewer:
Just FYI @oyeanuj, I'm using Babel Polyfill 6.3.14 and saw the Error until I chained a catch with an actual function parameter in it (solving exactly what the error was complaining I wasn't doing).

Just my 2 cents, but I'd rather this come across as a Warning in the console than an Error.

erikras:
You guys might be interested in [`v5.1.0`](https://github.com/erikras/redux-form/releases/tag/v5.1.0).

PR #820, opened 2016-04-15
A new prop: onSubmitFail

puradox:
This pull request is partly stemming from the conversation over at #776.

The way that we currently are submitting forms is great! Except for one thing... we don't know whether or not the submission failed. There is a current (quite *hacky*) way of checking this through the use of the `returnRejectedSubmitPromise` flag. Using this flag is quite tricky and cumbersome. It causes what some might call a programmer's headache (#488, #256, #291, #616). It even called a noble programmer, @pcgilday, to resolve some confusion with [his comment](https://github.com/erikras/redux-form/issues/616#issuecomment-209105697) which acted as a guide for many.

I believe that handling a failed submission should be more programmer-friendly, in a way that is also React oriented. So all of this is a major cause for a new prop: `onSubmitFail`.

The new prop `onSubmitFail` will be passed to your form upon initialization, and will act like one would think. It gets called when the form gets submitted and failed validation. Simple enough. No ES6 promises. No more headaches. Just a callback. And maybe even a smile :smile:

Callback syntax: `onSubmitFail(errors: object | undefined)`

The `errors` parameter will contain the errors returned from *synchronous* validation. This parameter will be an `object` if a synchronous validation fails in the form `{ field: error, ... }`. However, it could also be `undefined` if an asynchronous validation fails.

Edit: Added new `errors` parameter


puradox:
As a worthy mention, I only intend for this to be a temporary solution to the problem. We getting closer to the big rewrite, `v6.0`. :tada:

erikras:
Nice work, @puradox. :+1:

erikras:
Someone was also asking for an `onSubmitSuccess` callback. While you're in there and clearly understand the tests and `handleSubmit()`, can you add that, too?

puradox:
@erikras Thanks! I assumed that the existing `onSubmit` callback was equivalent to an `onSubmitSuccess` callback, am I wrong? I'll double check in the code.

erikras:
Both `onSubmitFailed` and `onSubmitSuccess` _could_ be done with the existing promise API, but it's about convenience.

puradox:
@erikras If I were to implement an `onSubmitSuccess` callback, there would be some a bit of extra redundancy. Are you sure you want me to implement this?

In my own code, I use `onSubmit` **without** the `returnRejectedSubmitPromise` flag to check if the submission was successful (it only fires if the submit was successful). Then I use `onSubmitFail` to check if the submission was erroneous.

This whole proposal was to eliminate the need for the `returnRejectedSubmitPromise` flag and promise rejection altogether, while maintaining backwards compatibility.

However, I do see your point in creating a more explicit callback that does exactly what it says. I agree that this would completely eliminate a lot of the existing confusion. I'll get working on it.

erikras:
No, I'm not 100% sure, but, at the risk of increasing API surface area, I do plan on implementing both in v6. You are more than welcome to say no. This PR still stands as an improvement.

puradox:
@erikras What's the status on this PR? Can we get it merged into `master`?

erikras:
I'll document and publish soon (tomorrow). 👍

puradox:
@erikras Thanks man! I'm very excited to see this go out. I appreciate all that you've done for the React / Redux community :smile: If you need any help with the new rewrite, you can always mention me. I'll be glad to help out.

erikras:
Published as [`v5.1.0`](https://github.com/erikras/redux-form/releases/tag/v5.1.0). Thanks again.

You've shown that you know your way around the library, so if I need someone for the rewrite, you're on the shortlist. 😉

CodyReichert:
Hi, I'm looking for the documentation for `onSubmitSuccess` and `onSubmitFailure`, but can't find it here: http://redux-form.com/5.2.3/#/api/props?_k=j59cbr

I see it's pretty know, so understandable that it's not up there yet. Is this something you're intentionally holding off on, or can someone (me or anyone else) provide a PR adding these to the v5 docs?

erikras:
@CodyReichert Ha, no, those are props that you pass to the decorated form, not that `redux-form` provides to your wrapped component. So their docs are [here](http://redux-form.com/5.2.3/#/api/reduxForm).

I suppose they _could_ be, but that was not the intention when `onSubmitSuccess` and `onSubmitFailure` were proposed and implemented.

CodyReichert:
@erikras - heh, I realized that _right_ after I submitted the comment. Thank you!

brennancheung:
How can I get access to React `refs` from inside of `onSubmitFail`?  I'm trying to do something like:

```
class LoginPage extends React.Component {
  render () {
    ...
    <input ref="password" />
    ...
}
export default reduxForm(
  {
    form: 'login',
    fields: [ 'username', 'password' ],
    validate,
    onSubmitFail: (errors) => {
      React.findDOMNode(this.refs.password).focus()
    }
  }
)(LoginPage)
```

Strajk:
Hi @brennancheung , have you found a way to do that?

brennancheung:
@Strajk I went with `document.getElementById('password').focus()` for now.  I know it's not the best way to do it and it won't work for server side rendering but I couldn't find any other way.  I still think it is an issue not being about to have access to the component within the `onSubmitFail` handler though.

puradox:
@brennancheung @Strajk You are going to want to utilize the prop version of `onSubmitFail`. 
> IMPORTANT: All of these configuration options may be passed into reduxForm() at "design time" **or passed in as props** to your component at runtime. ([source](http://redux-form.com/5.2.3/#/api/reduxForm?_k=v6g2l8))

Your refactored React Component would then look **something** like this (I haven't tested this):
```javascript
@reduxForm({
  form: 'login',
  fields: [ 'username', 'password' ],
  validate
})
class LoginPage extends Component {
  focusPasswordField() {
    this.refs.password.focus();
  }
  render() {
    ...
    <input ref="password" />
    ...
  }
}

class App extends Component {
  handleSubmitFail(errors) {
    this.refs.login.focusPasswordField();
  }
  render() {
    ...
    <LoginPage ref="login" onSubmitFail={this.handleSubmitFail} />
    ...
  }
}
```

brennancheung:
@puradox, `this.refs.login.focusPasswordField` does not exist in your example.  `this.refs.login.refs` exists but it doesn't contain the actual refs (it's an empty object).  None of the methods I define in `LoginPage` seem to be visible from `this.refs.login`.

puradox:
@brennancheung Ah yes, I forgot that the class is being wrapped by Redux-Form. I doubt that we'd be able to access the underlying component (`LoginPage` in our case) since `createHigherOrderComponent` doesn't keep a `ref` to it. That means that we have to think of another way...

puradox:
@brennancheung Instead, another approach would be one that is completely unrelated to `onSubmitFail`, is to provide your own `handleSubmit` method. I tested this time :smile: 

```javascript
function validate(values) {
  const errors = {};
  if (!values.password || values.password !== 'secret') {
    errors.password = 'Invalid';
  }
  return errors;
}

@reduxForm({
  form: 'login',
  fields: ['username', 'password'],
  validate
})
class LoginPage extends Component {

  handleSubmit = (event) => {
    const { fields, handleSubmit } = this.props;

    handleSubmit(event);

    if (fields.password.error) {
      this.refs.password.focus();
    }
  }

  render() {
    const { fields: { username, password } } = this.props;
    return (
      <form onSubmit={this.handleSubmit}>
        <input placeholder="Password" type="password" ref="password" {...password} />
        <input placeholder="Username" {...username} />
        <button type="submit">Submit</button>
        {password.touched && password.error && <div>{password.error}</div>}
      </form>
    );
  }
}

class App extends Component {
  handleSubmit = values => console.log(values)
  handleSubmitFail = errors => console.log(errors)
  render() {
    return (
      <LoginPage onSubmit={this.handleSubmit} onSubmitFail={this.handleSubmitFail} />
    );
  }
}
```

PR #828, opened 2016-04-19
[BUG] Fix support for async validation with onSubmitFail

puradox:
The initial implementation of onSubmitFail was not reporting failures correctly for asynchronous validation functions, instead it didn't do anything. This was not the intended behavior.

This was an overlook on my part, as I didn't include this in the initial implementation. I hope I fixed up the issue before anyone else experienced this in their code.


erikras:
Does this fix #831?

aight8:
I tested it in my use case with the this PR and it works :)

PR #833, opened 2016-04-20
fix error thrown when validation result is undefined

pjvds:
The `readFields` function throws an error when the `validate` function returns `undefined`.

This pull request fixes that and includes a test to describe the scenario that went wrong.


erikras:
👍 Thanks.

PR #843, opened 2016-04-22
Fix #621 - Account for initialValues in getValues

### LINKED ISSUE #621: Form values not initialized from initialValues until 2nd render cycle
We have been struggling w/ this for a while now and following some of the related issues. I'm not sure if our problem is directly related to these others or not (using release `4.1.6`):

- https://github.com/erikras/redux-form/issues/176
- https://github.com/erikras/redux-form/issues/578

To summarize, we are finding that values are not being initialized until the second render cycle for our reduxForm-ed components. Example being that none of the following properties would be initialized until the first call of `componentWillReceiveProps` (i.e. the 2nd render):

- `props.fields.someField.value`
- `props.values.someField`

Although, the following properties **are** initialized (even for the 1st render cycle) according to `initialValues`

- `props.fields.someField.defaultValue`
- `props.fields.someField.initialValue`
- etc...

My question would be, is what I've described above the _expected_ behavior from `redux-form`? This causes issues for us where children component who depend on valid props being passed down to them receive `undefined` on the 1st render cycle and we can not rely on the "values" being reliably initialized.

We have resorted to something similar to

```js
render() {
  if (typeof values.someField === 'undefined') {
    return null;
  }

  // ...
}
```

but then we can not rely on componentDidMount, etc...

I will make an effort to circle back and try and clarify this issue. I just want to get this out there for possible discussion. Any feedback/advice is much appreciated!

---

Here is some mostly-valid code below to illustrate the issue we have been experiencing

```js
const SomeForm = React.createClass({

  componentWillMount() {
    const {values} = this.props;
    if (typeof values.someString === 'undefined') {
      // All values = undefined
      // props.fields.someString.defaultValue is set as expected
      // props.fields.someString.initialValue is set as expected
      debugger;
    } else {
      debugger;
    }
  },

  componentDidMount() {
    const {values} = this.props;
    if (typeof values.someString === 'undefined') {
      // All values = undefined
      // props.fields.someString.defaultValue is set as expected
      // props.fields.someString.initialValue is set as expected
      debugger;
    } else {
      debugger;
    }
  },

  componentWillReceiveProps(newProps) {
    const {values} = newProps;
    if (typeof values.someString === 'undefined') {
      debugger;
    } else {
      // Values initialized on 1st call (2nd render)
      debugger;
    }
  },

  render() {
    const {
      fields: {someString, someArray, someObject},
    } = this.props;

    return (
      <div>
        <ComponentUsesSomeString
          value={someString.value}
          onChange={someString.onChange}
        />
        <ComponentUsesSomeArray
          value={someArray.value}
          onChange={someArray.onChange}
        />
        <ComponentUsesSomeObject
          value={someObject.value}
          onChange={someObject.onChange}
        />
      </div>
    )
  }

});

const ReduxFormedSomeForm = reduxForm(
  {name: 'SomeForm', fields: ['someString', 'someArray', 'someObject']},
  (state) => {
    return {
      initialValues: {
        someString: 'Some String',
        someArray: ['some', 'array'],
        someObject: {a: 'some', b: 'object'},
      },
    };
  }
)(SomeForm);

const App = React.createClass({
  render() {
    return (
      <div>
        <ReduxFormedSomeForm />
      </div>
    )
  }
})
```

erikthedeveloper:
- [x] getValues - should fallback to initialValues when values are undefined

Previously, form values were not being initialized from initialValues until the 2nd render cycle. This caused difficulties with server side rendering, double render cycles, and other related pains.

See https://github.com/erikras/redux-form/issues/621 for more discussion and links to possibly related issues.

This is addressed by falling back to `field.initialValue` when `field.value` is `undefined`

```js
/**
 * Given a state[field], get the value.
 *  Fallback to .initialValue when .value is undefined to prevent double render/initialize cycle.
 *  See {@link https://github.com/erikras/redux-form/issues/621}.
 */
const itemToValue =
  ({value, initialValue}) => typeof value !== 'undefined' ? value : initialValue;
```

:memo: this also affected array fields. This was accounted for in both implementation and existing test.

### DIFF
--- a/src/getValues.js
+++ b/src/getValues.js
@@ -1,3 +1,11 @@
+/**
+ * Given a state[field], get the value.
+ *  Fallback to .initialValue when .value is undefined to prevent double render/initialize cycle.
+ *  See {@link https://github.com/erikras/redux-form/issues/621}.
+ */
+const itemToValue =
+  ({value, initialValue}) => typeof value !== 'undefined' ? value : initialValue;
+
 const getValue = (field, state, dest) => {
   const dotIndex = field.indexOf('.');
   const openIndex = field.indexOf('[');
@@ -24,7 +32,7 @@ const getValue = (field, state, dest) => {
         getValue(rest, item, dest[key][index]);
       });
     } else {
-      dest[key] = array.map(item => item && item.value);
+      dest[key] = array.map(itemToValue);
     }
   } else if (dotIndex > 0) {
     // subobject field
@@ -35,7 +43,7 @@ const getValue = (field, state, dest) => {
     }
     getValue(rest, state && state[key] || {}, dest[key]);
   } else {
-    dest[field] = state[field] && state[field].value;
+    dest[field] = state[field] && itemToValue(state[field]);
   }
 };
 
--- a/src/__tests__/getValues.spec.js
+++ b/src/__tests__/getValues.spec.js
@@ -18,6 +18,21 @@ describe('getValues', () => {
       });
   });
 
+  it('should fallback to initialValues when values are undefined', () => {
+    const form = {
+      someText: {value: 'Hello Again!', initialValue: 'Hello'},
+      someArray: {value: undefined, initialValue: [1, 2, 3]},
+      someBool: {value: undefined, initialValue: false},
+    };
+    const fields = ['someText', 'someArray', 'someBool'];
+    expect(getValues(fields, form))
+      .toEqual({
+        someText: 'Hello Again!',
+        someArray: [1, 2, 3],
+        someBool: false,
+      });
+  });
+
   it('should allow undefined values', () => {
     const form = {
       foo: {value: 'bar'}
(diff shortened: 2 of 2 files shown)


src/getValues.js
```diff
@@ -24,7 +32,7 @@ const getValue = (field, state, dest) => {
         getValue(rest, item, dest[key][index]);
       });
     } else {
-      dest[key] = array.map(item => item && item.value);
+      dest[key] = array.map(itemToValue);
```
erikthedeveloper: I think this may need to be updated to

```diff
-.map(itemToValue)
+.map(item => item && itemToValue(item))
```

All tests are currently passing, but I'm not familiar enough with array fields to know off hand if we are missing an edge case. See [should get values from array form](https://github.com/erikthedeveloper/redux-form/blob/bug/621-values-not-initializing/src/__tests__/getValues.spec.js#L73-L89) for current coverage.

PR #847, opened 2016-04-22
Adding example on how to use stopSubmit

nicangeli:
I'm keen to help out writing some docs, just submitting this one to see what you think and if you have any suggestions on where I should focus?

### DIFF
--- a/examples/submitValidation/src/StopSubmitReadme.md
+++ b/examples/submitValidation/src/StopSubmitReadme.md
@@ -1,0 +1,5 @@
+# Using Redux Middlware
+
+If you can't return a promise to the `onSubmit` method (because you're using some redux middleware to perform your side effects), instead trigger the stopSubmit action when your server side validation has failed.
+
+Pass `stopSubmit` as a parameter the key of your form and an error object, use `_error` to indicate a generic error, rather than an error with a specific field. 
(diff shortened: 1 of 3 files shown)


erikras:
I'm not sure what you're looking for. `STOP_SUBMIT` is the action that is fired when the submission promise finishes. It's not really something that should be called manually in most cases.

erikras:
Whoa. I'm gonna need to think more about this. I'll get back to you.

erikras:
First of all, that's a clever usage of the `stopSubmit` action. I often don't have a good response to people who are using middleware that prevents promises being returned.

I think I'd like to keep all the actual js files in the examples as eslint-able actual code that runs in the examples. So perhaps you could just put the js directly into the `StopSubmitReadme.md`?

patrickgordon:
This would be super useful for me as I am currently trying to unpack how to achieve this with my API middleware! Good initiative

patrickgordon:
I took the opportunity to try this out and there are a few things I'm still unclear about. I think the example in your docs is a good start but it doesn't really go in to the depth I was hoping for. 

For example, my form values are being cleared after issuing the `stopSubmit` and I'm not sure if there's a way to ensure this doesn't happen. 

Do you have any intention of expanding these docs? If not, would anyone be able to help me unpick this further?

patrickgordon:
Further investigation - looks like if I set `destroyOnUnmount: false` when creating the redux form then it persists through the API call.

Most likely will need to dispatch the action `destroy(formName)` after a successful submit occurs.

PR #850, opened 2016-04-23
Adding Multi page wizard form example for v6

mihirsoni:
This is my attempt to moving wizard multipage example for v6 , please guide me if something is wrong with what I have done.

I see below issue , 
* Still partial form validation is not functioning as it should. Despite of giving validation on each step separately it validates whole form. Is this is a issue or something I have done wrong.

Haven't added links yet to `gh-pages`. I will add if this PR is wroth.

If it is not required please ignore.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -44,4 +44,6 @@ open [`localhost:3030`](http://localhost:3030) in your browser.
 > How to use Immutable JS for `redux-form` state storage.
 
 ---
-  
+### [Multi-page wizard](wizard)
+
+> How to separate a single form out into sepearate pages of inputs.
--- a/examples/wizard/.babelrc
+++ b/examples/wizard/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["react", "es2015"]
+}
--- a/examples/wizard/README.md
+++ b/examples/wizard/README.md
@@ -1,0 +1,10 @@
+# Multi Page Wizard Form Example
+
+## To run locally
+
+```
+npm install
+npm start
+```
+
+Then open [`http://localhost:3030/`](http://localhost:3030/).
--- a/examples/wizard/devServer.js
+++ b/examples/wizard/devServer.js
@@ -1,0 +1,27 @@
+var path = require('path');
+var express = require('express');
+var webpack = require('webpack');
+var config = require('./webpack.config.dev');
+
+var app = express();
+var compiler = webpack(config);
+
+app.use(require('webpack-dev-middleware')(compiler, {
+  noInfo: true,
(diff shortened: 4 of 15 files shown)


erikras:
Definitely going to check this out.

I've implemented `v6` validation in such a way that it will only validate fields that appear on your form, so _in theory_, the wizard example could use the same full-form validation function on all pages. But I had not really tested that yet.

erikras:
> Still partial form validation is not functioning as it should. Despite of giving validation on each step separately it validates whole form. Is this is a issue or something I have done wrong.

Ahhh, I know what is happening. In `v5`, because we had a list of fields, when you hit `submit`, it was marking each field as `touched`. In `v6`, because we _don't_ have a list of fields, it's just setting an `allTouched` flag which overrides the `touched` flag for each field. So when you submit the first page, all the fields on the second page think they have been touched. I might be able to work around this, though.

Thank you for your work on this, @mihirsoni! It's much appreciated.

If you feel like tackling another one, I'd really love some examples that show off popular third-party inputs, like (in order of preference) [Material UI](http://www.material-ui.com), [react-select](http://jedwatson.github.io/react-select/), and [Belle](http://nikgraf.github.io/belle). I'm not sure if they each deserve their own example project or could be lumped in together. TBD.

mihirsoni:
@erikras  Sure, you can run this example to reproduce issue, this example has validation on only two step. 

Steps to reproduce 

1. After running the example , just click on `Next` It will throw required errors.
2. Now populate data, and press next, you will see required error on second step as well. Ideally it shouldn't be as we haven't touched / hit submit on that form.

mihirsoni:
> If you feel like tackling another one, I'd really love some examples that show off popular third-party inputs, like (in order of preference) Material UI, react-select, and Belle. I'm not sure if they each deserve their own example project or could be lumped in together. TBD.

@erikras  Let me know what are your thoughts on that.  I will try if I can.

erikras:
My "thoughts on that" are, 1) first do a Mat UI only version, 2) then add others, 3) see if it feels too bulky, 4) if it does, break them up.

mihirsoni:
@erikras would you like to put into the same example repo?

erikras:
Yes

erikras:
Different PR though

erikras:
Nice work, @mihirsoni! Published: http://redux-form.com/6.0.0-alpha.6/examples/wizard/

mihirsoni:
@erikras  Thanks. It seems , you did CLI merge. You can close this if that is what it has happened.

erikras:
Yes, I thought the CLI merge would close this. Oh well. Thanks again.

Issue #742, opened 2016-03-16
Async validation does not fire with initial values


VladShcherbin:
Hey, seems like I have the same [issue](https://github.com/erikras/redux-form/issues/614) when async function is not called with initial values.

What am I doing:
I've got a simple form with 3 fields:
- input text
- checkbox
- dropzone (with input inside)

The checkbox has an initial value. The onBlur won't trigger the async validation until I change the value in one of the fields. As in the issue above, [this line](https://github.com/erikras/redux-form/blob/master/src/createHigherOrderComponent.js#L78) is probably causing this.

Is there any fix?

jeromebridge:
I am trying to run validations (at least the async validation) on the initial values of the form when it first comes up.  Is this possible or maybe there is a recommend way to run the async service when the form first comes up?

erikras:
That line is assuming that you would not initialize your form with values that fail async validation. Can you tell me more about your use case? If you just need to do a "required" validation on one of the other fields if the checkbox is checked, that could be done with sync validation, no?

jeromebridge:
The user is entering a new transaction on the page.  I would like the input of the user to be "quoted" while they are entering data on the form as soon as the minimum input is entered by the user.

In some cases the "Initial Values" may be enough minimum input to quote the transaction and sometimes it may not.  I was thinking I would like the redux-form validations to determine if the initial values are adequate to do an initial quote when the user first opens the transaction input page and then do the quote if it is enough input to do the quote.

Maybe there is a better way to approach the problem.  BTW...the [React Redux Example](https://github.com/erikras/react-redux-universal-hot-example) you did is fantastic!!

VladShcherbin:
@erikras if I remember correct, I wanted to validate the file's (image) width & height, so I needed to use async FileReader. I wanted to wrap it in validation rule.

When I was using dropzone and onDrop event, I used **field.onChange**, but it didn't trigger the async validation.

I solved it by moving this image validation into the Dropzone component (to check and only then to call field.onChange), but I can probably reproduce my issue.

jeromebridge:
I am trying to run validations (at least the async validation) on the initial values of the form when it first comes up.  Is this possible or maybe there is a recommend way to run the async service when the form first comes up?

tomduncalf:
[...]

Issue #614, opened 2016-02-02
asyncValidate not called when field empty (even when field has been touched)


tomduncalf:
I'm not sure if this behaviour is by design or not, but I have found that if I have a form field which is async validated, the validation function is not called when the field is empty - even if it has previously been touched/edited. This means that I can't perform "required string" validation async, as far as I can tell.

Steps to reproduce:
1. Create a redux-form with one field, with an `asyncValidate` function and the field in `asyncFields`
2. Enter some text in the field and blur it - the validation function will be called
3. Clear the text in the field and blur it - the validation function will not be called

I would expect the function to be called in 3 as I have previously touched the field, and I might for example want to check if a required value is present.

A tiny bit of debugging shows that at 3, `allPristine` in https://github.com/erikras/redux-form/blob/a53f22b816d72dca726d7a220b0e1428dbe522f8/src/createHigherOrderComponent.js#L72 is `true` - perhaps I've misunderstood what pristine refers to in this context, but I think perhaps `allPristine` should be false, as this field has been touched.

Any thoughts? At the minute this prevents me from using async validation for the form I am working on (trying to use https://github.com/jquense/yup to validate against a schema, but it only works with callbacks/promises, not synchronously)

Thanks!
Tom

matt-tock:
I'm also running into this issue. I want to run async validations on a pristine form. The most recent change to check for `allPristine` prior to running async validations breaks my ability to do so.

One use case for this is something like a login form, where sync validations are overkill. I expect my users to understand they need to fill out the entire form to login. But I do want to show an error if they try to submit a blank form or if they accidentally click the button.

tomduncalf:
Ah yes, rolling back to `4.1.5` fixes the issue for now :+1: Cheers!

erikras:
This was introduced by #589, by @cwmanning. My reasoning there is that "require" validation should be "sync".

> One use case for this is something like a login form, where sync validations are overkill. I expect my users to understand they need to fill out the entire form to login. But I do want to show an error if they try to submit a blank form or if they accidentally click the button.

That makes some sense. What should be the correct default behavior, I wonder?

One option we have available is that `redux-form` knows whether or not a form has been initialized with values, so maybe we do the async validation if `syncValidationPasses && (!allPristine || !initialized)`?

tomduncalf:
[...]

PR #858, opened 2016-04-25
Add alwaysAsyncValidate option, to always perform async validation on blur

tomduncalf:
Submitting this as a "straw man" solution to https://github.com/erikras/redux-form/issues/614. For my use case, I always want to perform asynchronous validation, even if the form is pristine (e.g. for required fields which have been touched) and this seemed like the most direct way to achieve it. Happy to discuss further if this is not ideal though!

--

This allows forms which use purely async validation to perform validation even if a field's value has been changed (e.g. show an error message if a user touches a required field, does not change it, then blurs it).

See https://github.com/erikras/redux-form/issues/614 and https://github.com/erikras/redux-form/issues/742 for example use cases.

### DIFF
--- a/docs/api/ReduxForm.md
+++ b/docs/api/ReduxForm.md
@@ -1,7 +1,7 @@
 # `reduxForm(config:Object, mapStateToProps?, mapDispatchToProps?, mergeProps?, options?)`
 
-Creates a decorator with which you use `redux-form` to connect your form component to Redux. It takes a `config` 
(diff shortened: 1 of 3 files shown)


tomduncalf:
Any thoughts on this @erikras?

erikras:
Seems reasonable.

tomduncalf:
Thanks :)

erikras:
Published as [`v5.2.3`](https://github.com/erikras/redux-form/releases/tag/v5.2.3).

Issue #848, opened 2016-04-22
Autofill/custom field property


trevorr:
Consider a postal address form where typing in a postal code can (in some cases) auto-fill the city and state/province fields. If the postal code is changed, it seems the auto-fill should only occur (again) if the city/state fields were not modified by the user (in other words, if they're blank or the previously auto-filled value). However, I'm not sure how to determine this.

- `dirty` and `pristine` seem to be relative to the initial value, so an auto-filled field would be considered `dirty`.
- `touched` and `visited` don't seem relevant, as tabbing through the fields shouldn't make them permanently un-auto-fillable.
- Component state can't be reliably used since fields can migrate across component instances (e.g. when an element of an array in a deep form is removed).

Therefore, it seems like a new field property is necessary. Is it possible/supported to create custom field properties? It seems like a reducer plugin could (and would be necessary to) accomplish this, but I'd want to make sure that redux-form wouldn't inadvertently drop it in some cases. It might also be nice to have explicit support for this, similar to how change actions can currently update the `touched` property.

erikras:
So you're pining for a boolean flag like, `wasEverChangedByUser`?

Or, to be more succinct, let's call it `modified`. And then, you want, presumably via an async call, to be able to fire off an ajax request when `postalCode` changes `&& !city.modified && !state.modified`, to look up `city` and `state` values and set them?

That's both a totally reasonable use case and totally beyond the scope of most form libraries.

trevorr:
`modified` is totally beyond the scope of most form libraries? That statement seems somewhat hyperbolic at least. :-)

What I'm really asking is whether the library would/could permit and propagate a custom field property in its state. I actually just finished trying it out, and the only problem was that the custom property was not propagated from the form state to the component props. Here's the reducer plugin:

```
const autofillReducer = (next: ?Reducer) =>
  (state: Object = {}, action: { type: string }): Object => {
    if (action.type === FormAction.AUTOFILL) {
      const { field, value } = action
      state = write(field + '.autofill', value, state)
      state = write(field + '.value', value, state)
    }
    return next ? next(state, action) : state
  }
```

The `autofill` property shows up in the state correctly, but doesn't make it to the form props.

davidkpiano:
> That's both a totally reasonable use case and totally beyond the scope of most form libraries.

[...]

PR #861, opened 2016-04-26
Add autofill action and autofilled field property. #848

trevorr:
As we discussed. I tried it out in my app, and it works well.


erikras:
WOW! You really knocked that one out of the park! Excellent tests and even documentation!

trevorr:
Well, I didn't want to make you have to ask. 😉  Thanks for the quick new version release!

Issue #859, opened 2016-04-25
Lodash in v6


anaibol:
Hi, putting the new version has highly increased the build size (more than 100kb) due mainly to Lodash:
Can we make the dependencies optional? Object.assign polyfill could be also optional. 
![selection_022](https://cloud.githubusercontent.com/assets/3210177/14791542/9b11dd94-0b16-11e6-8594-2c789ceec1bb.png)

erikras:
Two of them were superfluous (see above commit). [The others](https://github.com/erikras/redux-form/blob/v6/package.json#L47-L51) are used, aside from `noop`, which is the most hilarious "dependency" ever.

I'm not an `npm` expert, so I don't really know what you mean by making them "optional". How would I go about that?

anaibol:
Well, maybe "optional" is not the right term. Due the overweight of adding these dependencies I would say to rewrite where they are been used.
Lodash is becoming really fat even if you load parts of it.
Sorted by asc size:

 [239] ./~/lodash.rest/index.js 8.22 kB {0} [built]
 [235] ./~/lodash.every/index.js 9.56 kB {0} [built]
 [236] ./~/lodash.mapvalues/~/lodash.keys/index.js 11.9 kB {0} [built]
 [232] ./~/lodash._baseeach/index.js 14.5 kB {0} [built]
 [105] ./~/lodash._stringtopath/index.js 19.1 kB {0} [built]
 [234] ./~/lodash._createwrapper/index.js 25.7 kB {0} [built]
 [104] ./~/lodash._baseiteratee/index.js 54.2 kB {0} [built]

Lodash gives some abstraction and helpfull utils but I am not sure it worth the cost.

erikras:
Okay. Good info. In `v5`, I had [written `mapValues` myself](https://github.com/erikras/redux-form/blob/master/src/mapValues.js), which is about as trivial an exercise in iteration as you can get, but I leaned on lodash to cut some corners on `v6`, and I thank you for speaking up about it.

anaibol:
So you think it will keep lodash? This library is becoming the defacto form management mechanism with React, like if React had jQuery as dependency but in this case it's like 100kb.
I don't know much of the Redux Form code, but I could try to remove that dependency.

erikras:
Those kilobyte sizes you posted above are ridiculously bloated for the functionality that lodash is giving this library. They should be removed (implemented locally). Thank you for pointing this out.

erikras:
[...]

PR #863, opened 2016-04-26
[V6] Use lodash.values instead of Object.values

ingro:
Since Object.values is not currently widely supported, it may be a safer option to use lodash version instead!

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "lodash.mapvalues": "^4.2.0",
     "lodash.noop": "^3.0.1",
     "lodash.partial": "^4.1.3",
-    "lodash.topath": "^4.2.0"
+    "lodash.topath": "^4.2.0",
+    "lodash.values": "^4.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


anaibol:
See #859. Object.values is polyfilled by Babel.
https://www.reindex.io/blog/you-might-not-need-underscore/

erikras:
What @anaibol said.

ingro:
Ah ok, nevermind, just saw the error `Uncaught TypeError: Object.values is not a function` on alpha6 and thought it was a typo.

steffenmllr:
@ingro How did you resolve this

Also ran into this while testing `6.0.0-alpha.6` - `Uncaught TypeError: Object.values is not a function` - any ideas on how to fix this ? Do I need to adjust my babel settings ?

ingro:
@steffenmllr I ended up requiring `babel-polyfill` at the top of my index script.

steffenmllr:
Oh that was easy! thanks :)

PR #886, opened 2016-04-29
fixes File transformed to Object after setErrors()

### LINKED ISSUE #740: dealing with File objects in fields
I have a problem with handling files:

1. I add a file to fields, e.g. to an "attachments" array
2. Stuff is validated at some point -> everything is fine
3. I submit the form -> everything is fine
4. However, if submission fails, and I try to re-submit, the File objects in the attachments array have been reduced to Objects after the first submission. At this points the objects have also lost their properties `lastModified`, `size`, `type` etc.

My field config is following:
```
"attachments[].id",
"attachments[].url",
"attachments[].file", <- the original File
```
I imagine that the problem has probably something to do with deep-cloning the fields at some-point (due to immutability) where the File object is converted to a plain object.

Any opinions whether I should continue building a hack where I store only ids of the files in the fields and the File references mapped to ids somewhere else, or is someone perhaps working with the problem already?

vitalis:
fixes #740
I was trying to reproduce it with tests without a success, but was easy to reproduce it in  
browser, I guess jsdom acts differently then browsers.
So i'm attaching screenshots of the issue before the fix and after:
Before failed validation:
before_validation

After failed validation before the fix:
after validation_broken

After failed validation with fix:
after_validation_fixed

I also had to upgrade the jsdom because previous version has no File support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "eslint-plugin-react": "^3.16.1",
     "expect": "^1.14.0",
     "isparta": "^4.0.0",
-    "jsdom": "~5.4.3",
+    "jsdom": "^8.4.1",
     "mocha": "^2.4.5",
     "mocha-lcov-reporter": "^1.1.0",
     "react": "^15.0.0",
--- a/src/__tests__/setup.js
+++ b/src/__tests__/setup.js
@@ -3,3 +3,4 @@ import { jsdom } from 'jsdom';
 global.document = jsdom('<!doctype html><html><body></body></html>');
 global.window = document.defaultView;
 global.navigator = global.window.navigator;
+global.File = global.window.File;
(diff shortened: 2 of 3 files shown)


erikras:
Published as [`v5.2.3`](https://github.com/erikras/redux-form/releases/tag/v5.2.3).

antoinerousseau:
On React Native, it throws me:
> Can't find variable: File

vitalis:
@antoinerousseau I'll try to fix it for react-native

antoinerousseau:
@vitalis thanks!
I guess you could do something like
```js
if (typeof File === 'function' && state instanceof File) {
  ...
```

vitalis:
https://github.com/erikras/redux-form/pull/902

Issue #879, opened 2016-04-28
v6 removes keys from `form.<formname>.values` object


anyong:
I'm not sure if this is the intended behavior, so please correct me if I'm wrong.

If you set the `initialValue` on an input field to a string value, and then delete (i.e., press backspace) the string in the input field, the `form.<formname>.values.field` property for that input field is missing in the state tree. It seems to me it would make more sense to leave the empty string there, rather than deleting the property.

For me, this came up when using the `<formname>.initial` and `<formname>.values` objects to create a JSON patch. I expected the patch to look like this:

`{ op: "replace", path: "/field", value: "" }`

but instead it looks like this:

`{ op: "remove", path: "/field" }`

which is incorrect. I am not *removing* the field from the form, I am simply setting it equal to a null string. That's when I noticed that the key was actually missing from the state tree.

However, I can see that this would be useful in the case where you actually *wanted* a null-string to eventually resolve to `undefined` somewhere down the line. Perhaps this should be a prop we can set on the form - `emptyStringToNull` or something?

erikras:
Yeah, ever since React `v15`, #703, the lines between `undefined`, `null` and `''` as form values have become very blurry. The best path forward is unclear to me.

anyong:
https://github.com/facebook/react/issues/6222

It seems like the correct thing to do is use `""`, as `undefined` would be more appropriate for a form input that didn't exist rather than one for a form with an empty input.

erikras:
`redux-form` _does_ pass `''` instead of `undefined` to the input, so that's not the issue. The issue, as I understand it, is what should the data look like _upon submission_ if a field has been touched, but left blank. That, I'm more unclear about.

anyong:
Right, that's what I meant though - if redux form passes `""` to the input, I think it ought to be consistent and pass `""` as the value on submit.

Also, if a user wants them to be removed, it's a lot easier to remove keys with a value of `""` than it is to figure out which ones are missing and add them back.

```js
_.pickBy(formData, (val, key) => val !== "")
```

vs.

```js
// need to pass around extra array of fields that are supposed to be on the object
expectedFields.forEach(field => {
    if (!formData.hasOwnProperty(field)) formData[field] = '';
});
```

But as I mentioned above, a config option passed to `reduxForm()` also makes sense, then you can just let the user decide which format they want.

erikras:
[...]

PR #924, opened 2016-05-07
Support setting a field that has been initialized to ""

mhahn:
@erikras I think this addresses the issue in https://github.com/erikras/redux-form/issues/879 while still maintaining the pristine behavior.


erikras:
This looks pretty reasonable. Thanks!

erikras:
Published as [`v6.0.0-alpha.8`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.8).

Issue #919, opened 2016-05-06
"Uncaught RangeError: Maximum call stack size exceeded": deepEqual chokes on props.children


gabrielecirulli:
In `6.0.0-alpha.7`,  when using the new `<Field>` element, I've been encountering a strange error when trying to compose elements that accept `children` props and passing them their children via the `<Field>` component.

Here is the most reduced working example I could get:

```jsx
import React, { Component } from "react";
import { reduxForm, Field } from "redux-form/immutable";

export default class InnerComponent extends Component {
  render() {
    return (
      <div>
        Something
      </div>
    );
  }
}

export default class CustomField extends Component {
  render() {
    return (
      <div onClick={() => this.props.onChange(true)}>
        {this.props.children}
      </div>
    );
  }
}

@reduxForm({
  form: "event",
})
export default class EventForm extends Component {
  render() {
    return (
      <Field
        name="customField"
        component={CustomField}>
          <InnerComponent/>
      </Field>
    );
  }
}
```

In a nutshell, `CustomField` will render anything that's passed as children to it. It has a simple `onClick` handler to trigger the `onChange` callback passed by `redux-form`, which sets the whole issue in motion.

`InnerComponent` is passed as a child. In my experiments, this issue does not happen when you pass just text inside `<Field>`.

From what I was able to figure out, the following file is at the root of the stack overflow: https://github.com/erikras/redux-form/blob/v6/src/ConnectedField.js#L19

What it does is it performs a deep comparison between the previous props and the next props of `<ConnectedField/>`, which is in-between `<Field>`, `<CustomField>` and `<InnerComponent>`.

![set up a new event impraise 2016-05-06 6 pm-48-10](https://cloud.githubusercontent.com/assets/886011/15079840/2500f18c-13bb-11e6-97b5-7faeaafce945.png)

`<ConnectedField/>`'s `props.children` is an `InnerComponent` element. The deep comparison moves on to `InnerComponent`'s `_owner` (private React property), which in this case is the top-level `EventForm` component. 

From that point, it's extremely easy to identify a circular reference, as you would only have to follow the `props.children` chain back down to find `InnerComponent` in order to close the loop.

When this happens, the component fails to update.

Potential solutions that I can think of:

 - Explicitly skip comparing the `children` property in the props object. Potential drawback: the component won't be able to tell when props changed. Potential workaround: only perform a shallow comparison of `children` objects.
[...]

PR #929, opened 2016-05-09
Fix deepEqual on cyclic data structure

gabrielecirulli:
This PR addresses the issue shown in #919 where cyclic data structure would cause updates to fail, as `deepEqual` would keep comparing until reaching the maximum call stack size.

The solution is implemented by adding a dependency on `lodash.isEqualWith`, which offers support for cyclic structures.

This PR also preserves the requirements that the original deepEqual had, such as considering `''` and `undefined` equal, and comparing both Immutable and basic data structures in any nesting order.

This is specifically implemented through a customiser passed to `isEqualWith`, which, on encountering an `Immutable` data structure, performs the comparison manually through `Iterable`'s `every` method, and delegates back to `isEqualWith` again (this allows supporting plain objects inside Immutable structures, which, although rare, are contemplated in the test suite).

As it stands, the test suite seems to pass. I also added a few new test cases concerning Lists, as the initial solution I implemented created a regression in deepEqual's `List` comparison.

Open points for discussion:
 - Is the dependency on `lodash.isEqualWith` worth it?
 - This implementation of deepEqual could supersede the `plain` version as well, and be used in both flavours at once. Should it be refactored to do so?

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "deep-equal": "^1.0.1",
     "es6-error": "^2.1.0",
     "hoist-non-react-statics": "^1.0.5",
-    "is-promise": "^2.1.0"
+    "is-promise": "^2.1.0",
+    "lodash.isequalwith": "^4.1.4"
   },
   "devDependencies": {
     "babel-cli": "^6.3.17",
--- a/src/structure/immutable/deepEqual.js
+++ b/src/structure/immutable/deepEqual.js
@@ -1,20 +1,22 @@
 import { Iterable } from 'immutable'
-import every from '../../util/every'
+
+import isEqualWith from 'lodash.isEqualWith'
+
+const customizer = (obj, other) => {
+  if (obj === undefined && other === '') return true
+  if (obj === '' && other === undefined) return true
+
+  if (Iterable.isIterable(obj) && Iterable.isIterable(other)) {
+    return obj.count() === other.count() && obj.every((value, key) => {
+      return isEqualWith(value, other.get(key), customizer)
+    })
+  }
+
+  return void 0
+}
 
 const deepEqualValues = (a, b) => {
-  if (a === undefined && b === '') return true
-  if (a === '' && b === undefined) return true
-  if (Iterable.isIterable(a)) {
-    if (Iterable.isIterable(b)) {
-      return a.count() === b.count() &&
-        a.every((value, key) => deepEqualValues(value, b.get(key)))
-    }
-    return false
-  }
-  if (!a || !b || typeof a != 'object' && typeof b != 'object') {
-    return a === b
-  }
-  return every(a, (value, key) => deepEqualValues(value, b[key]))
+  return isEqualWith(a, b, customizer)
 }
 
 export default deepEqualValues
--- a/src/structure/immutable/__tests__/deepEqual.spec.js
+++ b/src/structure/immutable/__tests__/deepEqual.spec.js
@@ -1,5 +1,5 @@
 import deepEqual from '../deepEqual'
-import { fromJS } from 'immutable'
+import { fromJS, List } from 'immutable'
 import expectations from '../expectations'
 import addExpectations from '../../../__tests__/addExpectations'
 
@@ -189,6 +189,33 @@ describe('structure.immutable.deepEqual', () => {
       },
       f: 4
     }, false)
+  })
+
+  it('should work with Immutable.Lists', () => {
+    let firstObj = { a: 1 }
+    let secondObj = { a: 1 }
+    let thirdObj = { c: 1 }
+
+    testBothWays(
+      List([ 'a', 'b' ]),
+      List([ 'a', 'b', 'c' ]),
+      false
+    )
+    testBothWays(
+      List([ 'a', 'b', 'c' ]),
+      List([ 'a', 'b', 'c' ]),
+      true
+    )
+    testBothWays(
+      List([ 'a', 'b', firstObj ]),
+      List([ 'a', 'b', secondObj ]),
+      true
+    )
+    testBothWays(
+      List([ 'a', 'b', firstObj ]),
+      List([ 'a', 'b', thirdObj ]),
+      false
+    )
   })
 
   it('should work with plain objects with cycles', () => {
--- a/src/structure/immutable/__tests__/deepEqual.spec.js
+++ b/src/structure/immutable/__tests__/deepEqual.spec.js
@@ -52,7 +52,7 @@ describe('structure.immutable.deepEqual', () => {
     }), false)
   })
 
-  it('work with plain objects', () => {
+  it('should work with plain objects', () => {
     testBothWays({
       a: {
         b: {
@@ -189,6 +189,31 @@ describe('structure.immutable.deepEqual', () => {
       },
       f: 4
     }, false)
+  })
+
+  it('should work with plain objects with cycles', () => {
+    // Set up cyclical structures:
+    //
+    // base1, base2 {
+    //   a: 1,
+    //   deep: {
+    //     b: 2,
+    //     base: {
+    //       a: 1,
+    //       deep: { ... }
+    //     }
+    //   }
+    // }
+
+    let base1 = { a: 1 }
+    let deep1 = { b: 2, base: base1 }
+    base1.deep = deep1
+
+    let base2 = { a: 1 }
+    let deep2 = { b : 2, base: base2 }
+    base2.deep = deep2
+
+    testBothWays(base1, base2, true)
   })
 
   it('should treat undefined and \'\' as equal', () => {


package.json
```diff
@@ -41,7 +41,8 @@
     "deep-equal": "^1.0.1",
     "es6-error": "^2.1.0",
     "hoist-non-react-statics": "^1.0.5",
-    "is-promise": "^2.1.0"
+    "is-promise": "^2.1.0",
+    "lodash.isequalwith": "^4.1.4"
```
ooflorent: Require `lodash` and then import `lodash/isEqualWith` would be better.

erikras: Why?

ooflorent:
In my opinion, using lodash is OK since most project that rely on Redux Form would probably use lodash as well. What's important when requiring lodash is to keep its version wide. Moreover, `lodash.*` should be avoided since they don't merge with `lodash`.

package.json
ooflorent: @erikras Because `lodash.*` packages inline all internal functions and variables instead of sharing them. If you have multiple `lodash.*` packages it would result in duplication.

erikras:
Plus, `redux` and `react-redux` both depend on `lodash`. So it doesn't seem to be going anywhere.

ooflorent:
I think it's better to depend on battle-tested utilities (like lodash) and report abusive file size their than reimplementing the wheel. @jdalton did an amazing job pushing lodash to redux and investigating the output size.

erikras:
I wonder what semver value we should give for the dep?

I just did a `npm install --save lodash` and got `"^4.12.0"`

`redux` and `react-redux` are using `"^4.2.1"`. Perhaps we should echo that?

package.json
erikras: I did this in a separate commit. https://github.com/erikras/redux-form/commit/7107c888ada3f43e0f21276ecf2564edd831223c

ooflorent:
Since `isEqualWith` is available since `4.0.0`, `^4.2.0` should be fine.

gabrielecirulli:
Thanks for reviewing! Any thoughts on reusing this implementation of deepEqual for the plain version as well? As far as I can tell it should be fully compatible.

erikras:
I understand that shared code savings will come when other lodash stuff is used, either in this lib or in another, but one way is more bloated. Here are the results of building `redux-form` with each dep option.

 | Bundle | Minified |
---- | ---- | ---- |
| `lodash/isEqualWith` | `219 kB` | `74.2 kB` |
| `lodash.isequalwith` | `216 kB` | `71.8 kB` |

erikras:
@gabrielecirulli Already redid the plain one just now. :-)

gabrielecirulli:
@erikras on your point about bloat, I think it's a good incentive in terms of making valuable use of what comes with lodash all throughout the project, as I've seen many areas that would benefit from it (simply coming from the perspective that using lodash instead of hand-rolled solutions will make this library generally more reliable, as it happened with deepEquals).

gabrielecirulli:
Also as @ooflorent also said, in 90% of the cases the bloat of lodash is absorbed back into the project using this library as they most likely already include it, and webpack/browserify already do a good job of preventing repetition.

erikras:
> I think it's a good incentive in terms of making valuable use of what comes with lodash all throughout the project

Yes, perhaps the `util` folder where I put all my implementations of `lodash` stuff should die.

The fact that using `lodash/specificThing` shares code that `lodash.specificthing` does not might have been causing some of the bloat seen in #859.

jdalton:
For smaller bundles you should use `lodash/thing` over `lodash.thing`. To satisfy devs the `lodash.thing` inlines many dependencies to avoid large dep graphs (read code duplication).

Tools like [lodash-webpack-plugin](https://github.com/lodash/lodash-webpack-plugin) and [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) only work on the `lodash/thing` form too.

For example using `babel-plugin-lodash`
```js
export { isEqualWith } from 'lodash';
```
Is transpiled to
```js
var _isEqualWith = require('lodash/isEqualWith');
module.exports.isEqualWith =  _isEqualWith;
```
Which creates a build that's 3.96 kB gzipped.
If you apply [lodash-webpack-plugin](https://github.com/lodash/lodash-webpack-plugin) it's further reduced to 3.15 kB.

gabrielecirulli:
Thanks @jdalton. Out of curiousity (since you're here😄) I'm wondering if the way we implemented the compatibility for Immutable using `isEqualWith` makes sense. I was a bit baffled that I couldn't find any references or prior art when it comes to comparing heterogeneous data structures. I understand `lodash` has to stay generic, but is there any discussion on supporting some "lingua franca" of sort that allows to perform operations the same way when it comes to libraries like Immutable?

src/structure/immutable/deepEqual.js
```diff
@@ -1,20 +1,22 @@
...
+  if (obj === '' && other === undefined) return true
+
+  if (Iterable.isIterable(obj) && Iterable.isIterable(other)) {
+    return obj.count() === other.count() && obj.every((value, key) => {
+      return isEqualWith(value, other.get(key), customizer)
+    })
   }
-  return every(a, (value, key) => deepEqualValues(value, b[key]))
+
+  return void 0
```
jdalton: I know this may be here for linting but if not you don't have to explicitly return `undefined`, it will be implicit.

src/structure/immutable/deepEqual.js
```diff
@@ -1,20 +1,22 @@
...
-        a.every((value, key) => deepEqualValues(value, b.get(key)))
-    }
-    return false
-  }
-  if (!a || !b || typeof a != 'object' && typeof b != 'object') {
-    return a === b
+import isEqualWith from 'lodash.isEqualWith'
+
+const customizer = (obj, other) => {
+  if (obj === undefined && other === '') return true
```
jdalton: Is `null ` also something you want to lump together? If so an `obj == null` check will capture `null` and `undefined`.

src/structure/immutable/deepEqual.js
```diff
@@ -1,20 +1,22 @@
...
+    })
   }
-  return every(a, (value, key) => deepEqualValues(value, b[key]))
+
+  return void 0
+}
+
+const deepEqualValues = (a, b) => {
+  return isEqualWith(a, b, customizer)
 }
```
jdalton: Arrow function shave implicit return so if it's within your style preference you could do:
```js
const deepEqualValues = (a, b) => isEqualWith(a, b, customizer)
```

src/structure/immutable/deepEqual.js
gabrielecirulli: @erikras thoughts on that?

jdalton:
@gabrielecirulli 
> Out of curiousity (since you're here 😄 ) I'm wondering if the way we implemented the compatibility for Immutable using isEqualWith makes sense. 

Yep, looks great!

>  I understand `lodash` has to stay generic, but is there any discussion on supporting some "lingua franca" of sort that allows to perform operations the same way when it comes to libraries like Immutable?

There hasn't been but I'd be for it. Our wiki is open (editable) so if you wanted to start an Immutable cook book or something that would be rad.

gabrielecirulli:
Great to hear, it could be really interesting! And thanks for checking out the code! I'll see if I can submit a follow-up PR to address your notes.

src/structure/immutable/deepEqual.js
erikras: I removed it. 😄 That was the first time I've seen that construct.

src/structure/immutable/deepEqual.js
erikras: > Is `null` also something you want to lump together?

I suspect so. Thanks, I think that's a bug.

> If so an `obj == null` check will capture `null` and `undefined`.

Yes, I'm aware of that, but I prefer to use `===` whenever I can to be explicit. I think a linter beat that into me a few years ago, and now I [punish all the other monkeys](http://i.snag.gy/kdu77.jpg).

src/structure/immutable/deepEqual.js
erikras: Yes, I love the shorter version.

erikras:
@jdalton Cheers for the code review. 👍

erikras:
Published as [`v6.0.0-alpha.8`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.8).

jdalton:
> Published as v6.0.0-alpha.8.

Do you have the size deltas (minified + gzipped) handy (before and after lodash).

erikras:
@jdalton I'll get them to you tomorrow. I will say that `babel-plugin-lodash` made a big difference. Especially after I put it into `plugins` instead of `presets` [facepalm].

jdalton:
>  I put it into `plugins` instead of `presets`

No worries, I think everyone has done that.
I made sure to specify both in the webpack config examples to help folks out.

erikras:
@jdalton 23% increase gzipped, but, as stated above, most likely absorbed by usage in a project of any size.

 | Bundle | Minified | gzipped |
---- | ----: | ----: | ----: |
| Before `lodash` | `166 kB` | `63.5 kB` | `14.1 kB` |
| After `lodash` | `218 kB` | `71.6 kB` | `17.4 kB` |
| Delta | `31.3%` | `12.8%` | `23.4%` |

jdalton:
@erikras Cool! I've bumped v0.3.0 of `lodash-webpack-plugin` which should further reduce things.

erikras:
| Bundle | Minified | gzipped |
---- | ----: | ----: | ----: |
| `lodash-webpack-plugin v0.3.0` | `214 kB` | `70.6 kB` | `17.0 kB` |

Slight, but appreciated. 👍

jdalton:
Sanity checking I created a test build with:
```js
export { every, mapValues, noop,  partial, partialRight, toPath } from 'lodash';
```
and the build size was `1.7 kB` gzipped. Did I miss a method?

ooflorent:
@jdalton `isEqualWith` I guess. That was the purpose of adding lodash back ;)

jdalton:
Ah yes! `isEqual` is heavier for reasons why it's needed here :D

erikras:
[_Some_ other coding](https://github.com/erikras/redux-form/compare/4a457d7f87aeec1e54eb6b6b26e6fbdf86a47369...6df75958133fe2f007a1007420211b87e04c5605) happened with commits around the lodash upgrade, but pretty minimal, and almost exclusively unit tests.

erikras:
> Ah yes! isEqual is heavier for reasons why it's needed here :D

Some of my implementations of the other ones you listed were literally one line. Not to mention `noop`. 😆

PR #937, opened 2016-05-10
Add es build

xdissent:
Lets package consumers tree-shake with webpack2, just like redux. In fact, I lifted the lodash-es babel plugin from the redux repo. 😉

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,14 @@
 {
-  "presets": ["es2015", "react", "stage-2"],
-  "plugins": ["lodash"]
+  "presets": ["es2015-webpack", "react", "stage-2"],
+  "env": {
+    "development": {
+      "plugins": ["lodash", "transform-es2015-modules-commonjs"]
+    },
+    "production": {
+      "plugins": ["lodash", "transform-es2015-modules-commonjs"]
+    },
+    "es": {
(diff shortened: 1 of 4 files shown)


.babelrc
```diff
@@ -1,4 +1,14 @@
...
+  "presets": ["es2015-webpack", "react", "stage-2"],
+  "env": {
+    "development": {
+      "plugins": ["lodash", "transform-es2015-modules-commonjs"]
+    },
+    "production": {
+      "plugins": ["lodash", "transform-es2015-modules-commonjs"]
+    },
+    "es": {
+      "plugins": ["lodash", "./babel-lodash-es"]
```
xdissent: Sucks to have to duplicate config, but babel always *adds* plugins apparently, and the order is important for the lodash-es plugin. I did both `development` and `production` so you don't have to remember to set `BABEL_ENV` all the time for non-es builds. All `es2015-webpack` does is load `es2015` and remove the `transform-es2015-modules-commonjs` plugin.

erikras:
Okay. If redux is doing it... 😄

erikras:
Published as [`v6.0.0-alpha.9`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.9).

Issue #931, opened 2016-05-09
How to show validation errors for FieldArray on change of child attribute (v6.0.0-alpha7)


tacomanator:
I'm trying out v6.0.0-alpha7 and having an issue in which a validation error set on the FieldArray is not showing when checking/unchecking the child field. Validation runs and sets the error, but the FieldArray component is not re-rendered.

Please excuse the syntax (LiveScript):

```livescript
class Form extends Component

  render: ->
    element FieldArray,
      name: "children"
      component: (children) ->
        div null,

           # log statement only called on first render and when adding new children
           # it's not called when checking Field inside, even though that may clear or result in error
           console.log children.error

           choices.map (child) ->
            element Field,
              name: "#{child}.selected",
              component: "input",
              type: "checkbox"

          element Button,
            on-click: ~> children.push {selected: false}
            "Add child"

form-options =

  form: \group

  validate: ({children}) ->

    selected-count = sum map (.selected), children

    children: 
      if selected-count < 1
      then {_error: "Please select a child"}


module.exports = redux-form form-options <| Form
```

The log statement only fires at the beginning or when a new child is pushed into the FieldArray. It's not logging anything when checking/unchecking the checkbox.

Is this the intended behavior? Do I have to manually ```touch``` the FieldArray field when one of the children is changed?

erikras:
This is a bug. There's a `shouldComponentUpdate` on `FieldArray` that is only updating if the length changes, and it needs to be looking at the sync errors, too. Thanks!

Man, that's some crazy-looking code. 🙈 👀 😆

tacomanator:
Ok, thank you :v:

halayli:
http://redux-form.com/6.0.0-alpha.7/examples/fieldArrays/

If you just set the club name and hit submit the error 'At least one member must be entered' doesn't show up. I am guessing it's for the same reason?

erikras:
Published fix as [`v6.0.0-alpha.9`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.9).

PR #940, opened 2016-05-11
Use shallowEqual in shouldComponentUpdate

### LINKED ISSUE #936: [v6] external props in Field will not be updated when props changed
```javascript
<Field
  component={celebrity => (
    <span>
      Celebrity: {celebrity.value.name}
      External: {celebrity.externalProp}
    </span>
  )}
  name={name}
  externalProp={this.state.externalProp}
/>
```

`externalProp` changes will not be updated in `Field`.

leesiongchan:
Do you think it makes sense to use `shallowEqual` or even remove the entire `shouldComponentUpdate` as `Field` should be treated as a simple component to connects form's state and so `shouldComponentUpdate` shouldn't be part of the job.

Fixes #936

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "invariant": "^2.2.1",
     "is-promise": "^2.1.0",
     "lodash": "^4.12.0",
-    "lodash-es": "^4.12.0"
+    "lodash-es": "^4.12.0",
+    "react-addons-shallow-compare": "^15.0.0"
   },
   "devDependencies": {
     "babel-cli": "^6.3.17",
--- a/src/ConnectedFieldArray.js
+++ b/src/ConnectedFieldArray.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux'
 import createFieldArrayProps from './createFieldArrayProps'
 import { partial, mapValues } from 'lodash'
 import plain from './structure/plain'
+import shallowCompare from 'react-addons-shallow-compare'
 
 const createConnectedFieldArray = ({
   arrayInsert,
@@ -23,7 +24,7 @@ const createConnectedFieldArray = ({
 
   class ConnectedFieldArray extends Component {
     shouldComponentUpdate(nextProps) {
-      return size(this.props.value) !== size(nextProps.value)
+      return shallowCompare(this, nextProps)
     }
 
     get syncError() {
--- a/src/Field.js
+++ b/src/Field.js
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react'
 import invariant from 'invariant'
 import createConnectedField from './ConnectedField'
+import shallowCompare from 'react-addons-shallow-compare'
 
 let keys = 0
 const generateKey = () => `redux-form-field-${keys++}`
@@ -18,7 +19,7 @@ const createField = ({ deepEqual, getIn }) => {
     }
 
     shouldComponentUpdate(nextProps) {
-      return this.props.name !== nextProps.name
+      return shallowCompare(this, nextProps)
     }
 
     componentWillMount() {
--- a/src/FieldArray.js
+++ b/src/FieldArray.js
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react'
 import invariant from 'invariant'
 import createConnectedFieldArray from './ConnectedFieldArray'
+import shallowCompare from 'react-addons-shallow-compare'
 
 let keys = 0
 const generateKey = () => `redux-form-field-array-${keys++}`
@@ -15,6 +16,10 @@ const createFieldArray = ({ deepEqual, getIn, size }) => {
       }
       this.key = generateKey()
       this.ConnectedFieldArray = createConnectedFieldArray(context._reduxForm, { deepEqual, getIn, size }, props.name)
+    }
+
+    shouldComponentUpdate(nextProps) {
+      return shallowCompare(this, nextProps)
     }
 
     componentWillMount() {
@@ -40,14 +45,14 @@ const createFieldArray = ({ deepEqual, getIn, size }) => {
     get name() {
       return this.props.name
     }
-    
+
     getRenderedComponent() {
       invariant(this.props.withRef,
         'If you want to access getRenderedComponent(), ' +
         'you must specify a withRef prop to FieldArray')
       return this.refs.connected.getWrappedInstance().getRenderedComponent()
     }
-    
+
     render() {
       const { ConnectedFieldArray } = this
       return <ConnectedFieldArray {...this.props} ref="connected"/>
--- a/src/__tests__/Field.spec.js
+++ b/src/__tests__/Field.spec.js
@@ -202,7 +202,7 @@ const describeField = (name, structure, combineReducers, expect) => {
       const stub = TestUtils.findRenderedComponentWithType(dom, Field)
       expect(stub.name).toBe('foo')
     })
-    
+
     it('should provide access to rendered component', () => {
       const store = makeStore({
         testForm: {
@@ -224,7 +224,7 @@ const describeField = (name, structure, combineReducers, expect) => {
       )
       const field = TestUtils.findRenderedComponentWithType(dom, Field)
       const input = TestUtils.findRenderedComponentWithType(dom, TestInput)
-      
+
       expect(field.getRenderedComponent()).toBe(input)
     })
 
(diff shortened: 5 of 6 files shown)


src/Field.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component, PropTypes } from 'react'
 import invariant from 'invariant'
 import createConnectedField from './ConnectedField'
+import shallowEqual from 'fbjs/lib/shallowEqual'
```
ooflorent: `fbjs` isn't a `redux-form` dependency.

ooflorent:
One drawback of this change is that it will rerender a lot more!
Let's consider the following `Field`:

```js
render() {
  return (
    <Field name="color" component="select">
      <option value="r">Red</option>
      <option value="g">Green</option>
      <option value="b">Blue</option>
    </Field>
  )
}
```

When `render` will be called, the `Field`'s `children` will change but still represent the same VDOM and a rerender will occur.

erikras:
I'm fine with a dep on [`shallowequal`](https://www.npmjs.com/package/shallowequal).

Please do `FieldArray` too.

erikras:
> When `render` will be called, the `Field`'s children will change but still represent the same VDOM and a rerender will occur.

Perhaps `children` could be excluded from the shallow equal check?

The tests should be sure that it only rerenders once when the `otherProp` changes.

ooflorent:
@erikras I think [`react-addons-shallow-compare `](https://facebook.github.io/react/docs/shallow-compare.html) would be better. It is widely used and so would virtually not enlarge the lib size.

ooflorent:
> Perhaps children could be excluded from the shallow equal check?

Same behavior when using a `render`-time created closure for `component`. If this change is shipped, it would be a good idea to add a note to the documentation.

leesiongchan:
@ooflorent [react-addons-shallow-compare](https://github.com/facebook/react/blob/db175052c00a65e6a852011f889c12fea50bb34b/src/addons/shallowCompare.js) takes 3 parameters which include `nextState` as well, does it really fit into our need?

ooflorent:
@leesiongchan Since `Field` and `FieldArray` don't have `state`, it doesn't matter. Shallow equal on `undefined` should be fast enough. What to you think @erikras about the addon?

erikras:
> What to you think @erikras about the add-on?

👍 Sounds good.

leesiongchan:
But this change to `FieldArray` will fail the test as [ConnectedFieldArray](https://github.com/erikras/redux-form/blob/v6/src/ConnectedFieldArray.js#L26) is comparing using the `size` of its `value`. Are we going to use `deepEqual` like [ConnectField](https://github.com/erikras/redux-form/blob/v6/src/ConnectedField.js#L18) does?

package.json
```diff
@@ -70,6 +70,7 @@
     "lodash-webpack-plugin": "^0.3.0",
     "mocha": "^2.4.5",
     "react": "^15.0.0",
+    "react-addons-shallow-compare": "^15.0.0",
```
ooflorent: Should be added to `peerDependencies` as well

src/FieldArray.js
```diff
@@ -17,6 +18,10 @@ const createFieldArray = ({ deepEqual, getIn, size }) => {
       this.ConnectedFieldArray = createConnectedFieldArray(context._reduxForm, { deepEqual, getIn, size }, props.name)
     }
 
+    shouldComponentUpdate(nextProps, nextState) {
+      return this.props.name !== nextProps.name || shallowCompare(this, nextProps, nextState)
```
ooflorent: You can skip `nextState` I guess

erikras:
> But this change to FieldArray will fail the test as `ConnectedFieldArray` is comparing using the size of its value. Are we going to use deepEqual like `ConnectedField` does?

No, `ConnectedFieldArray` should just use shallow equals. We don't want it rerendering the whole array as you type into one of the inputs.

leesiongchan:
Updated.

package.json
leesiongchan: ping @erikras

ooflorent: This is critical. If you don't add it to `dependencies` or `peerDependencies`, the package will be broken.

package.json
```diff
@@ -43,7 +43,8 @@
     "hoist-non-react-statics": "^1.0.5",
     "invariant": "^2.2.1",
     "is-promise": "^2.1.0",
-    "lodash": "^4.12.0"
+    "lodash": "^4.12.0",
+    "react-addons-shallow-compare": "^15.0.0"
```
leesiongchan: @ooflorent I think is more appropriate to add to dependency compare to peer. @erikras  Let me know if I am wrong.

erikras: I'm no npm expert, but I would think it belongs in `dependencies`.

src/Field.js
```diff
@@ -18,7 +19,7 @@ const createField = ({ deepEqual, getIn }) => {
     }
 
     shouldComponentUpdate(nextProps) {
-      return this.props.name !== nextProps.name
+      return this.props.name !== nextProps.name || shallowCompare(this, nextProps)
```
erikras: This is redundant, is it not? Should just be `shallowCompare()`.

src/FieldArray.js
```diff
@@ -17,6 +18,10 @@ const createFieldArray = ({ deepEqual, getIn, size }) => {
       this.ConnectedFieldArray = createConnectedFieldArray(context._reduxForm, { deepEqual, getIn, size }, props.name)
     }
 
+    shouldComponentUpdate(nextProps) {
+      return this.props.name !== nextProps.name || shallowCompare(this, nextProps)
```
erikras: Same as `Field`. Should just be `shallowCompare()`

ooflorent: @erikras Yes you're right. `this.props.name !== nextProps.name` would not bring perf wins

leesiongchan:
Updated.

src/ConnectedFieldArray.js
```diff
@@ -23,7 +24,7 @@ const createConnectedFieldArray = ({
 
   class ConnectedFieldArray extends Component {
     shouldComponentUpdate(nextProps) {
-      return size(this.props.value) !== size(nextProps.value)
+      return shallowCompare(this, nextProps)
```
ooflorent: Are you sure for this one? cc @erikras

leesiongchan: Discussed on above. [Check here](https://github.com/erikras/redux-form/pull/940#issuecomment-218401294)

erikras: I'm pretty sure. I think this will fix #931.

@leesiongchan, can you write a test where there is a `_error` validation error for the array based on some quality of the values inside the array (e.g. if none of them have values), and verify that the array component rerenders when the error appears or disappears?

leesiongchan: @erikras Are you saying something like this? [Link](https://github.com/erikras/redux-form/pull/940/files#diff-c51b679ed2957be9891b8e3164a067b1R312)

erikras: Yes.

erikras:
Published as [`v6.0.0-alpha.9`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.9). Thanks, @leesiongchan!

erikras:
@leesiongchan and @ooflorent, as you may get an earlier start on your Friday than I...

The examples have been giving me `Uncaught Error: Cannot find module "react-addons-shallow-compare"`. I looked up the code in `node_modules/react-addons-shallow-compare`, and it was literally just `require()`ing `react/lib/shallowCompare`, so, out of desperation, I tried just importing `react/lib/shallowCompare` instead of `react-addons-shallow-compare`, but I am _still_ getting the same error message. e.g. [`6.0.0-alpha.10` Simple Example](http://redux-form.com/6.0.0-alpha.10/examples/simple/). Do you have any insights into what the problem might be? I cannot find any references to `react-addons-shallow-compare` anywhere in the project.

Thanks!

leesiongchan:
Can't see the problem too :'(

ooflorent:
@erikras I can't reproduce the issue. I did a fresh install and ran the builds and test suites and everything worked fine.

erikras:
Okay. I swear I was `rm -rf`ing all of the `node_modules` directories last night, but I'll try with a fresh `git clone` this morning to be absolutely sure.

Thanks for confirming my insanity. 😄 👍

ooflorent:
Maybe you were hitting Babel's cache of webpack's one

PR #942, opened 2016-05-11
add `field` parameter to docs for startAsyncValidation Action Creator

rouanw:
A small update to the ActionCreator API docs. Adds the `field` parameter to the `startAsyncValidation` signature and describes the relevant behaviour.


ooflorent:
Thanks for your contribution! It looks good to me.

ericbiewener:
Looks like this is still missing from the docs page: http://redux-form.com/6.7.0/docs/api/ActionCreators.md/

Issue #383, opened 2015-12-04
An empty object is returned if a field is undefined when calling getValues


xcatliu:
```js
reduxForm({
  form: 'simple',
  fields: [ 'name' ]
})(SimpleForm);

// action creator
import { getValues } from 'redux-form';

console.log(getValues(getState().form.simple));
// { name: {} }
```

erikras:
Thanks. Will investigate. :mag:

ooflorent:
I'm unable to reproduce the bug.

```js
it('should ignore undefined fields', () => {
  const state = {
    name: makeFieldValue({}),
  };
  expect(getValuesFromState(state))
    .toBeA('object')
    .toEqual({});
});
```

As expected, the result is `{}`.
@erikras Can we close this one?

erikras:
Sure. I never got around to investigating. 😦 

Go ahead and PR that test if you want. You can never have too many tests, especially if they demonstrate functionality reported as buggy.

PR #944, opened 2016-05-11
Add a test to invalid #383

ooflorent:
The following test shows that the behavior described in #383 is no longer reproducible using `v5.1.3`.


erikras:
👍

PR #952, opened 2016-05-12
Fix several FAQ typos

ztanner:
Just a few small things I spotted while reading the FAQ.


erikras:
Wow, those are embarrassing. Thanks!

PR #954, opened 2016-05-13
v6 — Add React to peer dependencies

ooflorent:
Since Redux Form requires React, it is a good idea to add it to `peerDependencies`. Moreover it would indicate what React versions are compatible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "webpack": "^1.12.13"
   },
   "peerDependencies": {
+    "react": "^15.0.0",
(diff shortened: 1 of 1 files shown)


JakeGinnivan:
Is react v15 actually required or is v0.14 enough?

ooflorent:
`react@15` is required for `redux-form@6`

ooflorent:
@erikras Are you ok with that?

erikras:
👍 Sure.

Issue #958, opened 2016-05-13
docs(props): fix propTypes link & change error type from string to any


fredrikolovsson:
When upgrading to v6 I read the docs and noticed in props that `error` is a `string`. This was a problem in our app, because we used to set an object on the `_error` key (to be able to display more and nicely formatted information about what happened).

At first I thought the API might have changed, but also the v4 and v5 docs said `error : **string**`, while in fact the [propType is `any`](https://github.com/erikras/redux-form/blob/master/src/createPropTypes.js). I would like to keep it that way, but thought it could be good to edit the docs. Also, the link to the `propTypes` file is broken in the docs.

I will submit a PR with these edits. Thanks for everything you do!

erikras:
Fixed by https://github.com/erikras/redux-form/commit/f2fe04310ce115530abbce61cb469f4be09c2a5c.

PR #959, opened 2016-05-13
docs(props): fix propTypes link and error type

fredrikolovsson:
Fixes https://github.com/erikras/redux-form/issues/958

Maybe the `error` prop for `Field` and `FieldArray` should also be `any` instead of `string`? However, I did not find those files in the repo, so this PR is just updating Props.md. Thanks.

### DIFF
--- a/docs/api/Props.md
+++ b/docs/api/Props.md
@@ -5,7 +5,7 @@ to your decorated form component. The `props` that _you pass into your wrapped c
 [here](#/api/reduxForm).
 
 > If you are a strict `PropTypes` completionist, `redux-form` exports all of these
-[`propTypes`](https://github.com/erikras/redux-form/blob/master/src/propTypes.js), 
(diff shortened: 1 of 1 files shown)


ooflorent:
What version of Redux Form is referenced here? I'm asking this because `Field` and `FieldArray` are available since v6 but you linked v5 codebase.

erikras:
The intention is good here, it just links to `master` instead of `v6`.

ooflorent:
Since we cannot change to target branch of a PR, a new one must be submitted... Could you do this?
By the way, thanks @fredrikolovsson for the contribution!

erikras:
I did it manually. Thanks, @fredrikolovsson.

fredrikolovsson:
My apologies for the noob PR. I selected the v6.0.0-alpha.8 tag and then forked, but realize now I should've forked the v6 branch 😄

Issue #962, opened 2016-05-13
[v6-alpha.10] Immutable import is broken


erikras:
The [Immutable Example](http://redux-form.com/6.0.0-alpha.10/examples/immutable/) won't load, because of a strange error:
```
SyntaxError: Unexpected token import
```
It's like it sees the first `import` line in `immutable.js` and it gets confused because it's not webpack/babel built fully or something. Happens in the development environment, too.

Possibly related to #937? @xdissent?

ooflorent:
I'm on it.

erikras:
Fixed by #963.

PR #963, opened 2016-05-13
v6 — Add missing Immutable.js dependency

ooflorent:
`redux-immutablejs` requires `immutable` as `peerDependencies` but it wasn't added to the example's `dependencies` causing a `TypeError`.

See #962 for more details.


erikras:
This fixes #962?

ooflorent:
Yes. I no longer experience the bug once the issue was added.

Since previous versions of `redux-form` required `immutable`, if you were using `npm@3`, everything was fine.

erikras:
Aha! Good work!

PR #964, opened 2016-05-13
v6 — Improve redux connector performances

ooflorent:
Use Redux's memoization API to improve `ConnectedForm` performances by avoiding closure re-creations.

**NOTE:** It introduces 2 breaking changes:

- `react-redux` MUST match `>=4.3` (previously `>=4.0`)
- `form`, `touchOnBlur` and `touchOnChange` can't change over time. In order to prevent WTF behaviors, we could add a warning when `NODE_ENV` is not `production` if these props change.

**Disclaimer:** I haven't tested the examples since they rely on the latest published version on Redux Form instead of the local one.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -1,4 +1,4 @@
-import React, { Component, PropTypes } from 'react'
+import { Component, PropTypes, createElement } from 'react'
 import invariant from 'invariant'
 import createConnectedField from './ConnectedField'
 import shallowCompare from 'react-addons-shallow-compare'
@@ -54,8 +54,10 @@ const createField = ({ deepEqual, getIn }) => {
     }
 
     render() {
-      const { ConnectedField } = this
-      return <ConnectedField {...this.props} ref="connected"/>
+      return createElement(this.ConnectedField, {
+        ...this.props,
+        ref: 'connected'
+      })
     }
   }
 
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
   },
   "peerDependencies": {
     "react": "^15.0.0",
-    "react-redux": "^4.0.0",
+    "react-redux": "^4.3.0",
     "redux": "^3.0.0"
   },
   "files": [
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -1,4 +1,4 @@
-import React, { Component, PropTypes } from 'react'
+import { Component, PropTypes, createElement } from 'react'
 import { connect } from 'react-redux'
 import createFieldProps from './createFieldProps'
 import { partial, mapValues } from 'lodash'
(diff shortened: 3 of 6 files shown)


ooflorent:
Fixed linter errors

src/Field.js
```diff
@@ -54,8 +54,10 @@ const createField = ({ deepEqual, getIn }) => {
     }
 
     render() {
-      const { ConnectedField } = this
-      return <ConnectedField {...this.props} ref="connected"/>
+      return createElement(this.ConnectedField, {
+        ...this.props,
+        ref: 'connected'
+      })
```
erikras: Anything in particular gained by avoiding JSX here? or just using the `createElement()` since we already have it in scope?

erikras:
> we could add a warning when NODE_ENV is not production if these props change.

`invariant` is already a dep. Use it for this.

src/Field.js
ooflorent: I replaced JSX in favor of `createElement` to take control over what transpilation is doing. The end result is cleaner in my opinion because by reading the code we immediately understand what react is doing. I've done these changes in a separate commit so it could be easily reverted.

Note: It is also how `react-redux` instantiate React elements.

erikras: 👍 Good enough. Was just curious.

erikras:
The benefit of this optimization just clicked in my mind, and now I understand it. That's a _lot_ of binding we will be avoiding on every render!  🎉 🎉  Definitely worth the cost of prohibiting the mutation of those three props!

ooflorent:
I've updated the PR to avoid creating the action map on each render.

erikras:
> Disclaimer: I haven't tested the examples since they rely on the latest published version on Redux Form instead of the local one.

You should _probably_ do that before we merge this, as it's a pretty big change. It's just a matter of copying the `lib` (and maybe `es`) folders over the ones in `node_modules/redux-form` in every example.

However, that the tests are all passing is promising. That's the nice thing about having so many tests.

---

Actually, it never occurred to me that now that the examples are in the same branch as the library code, that they could just `import { reduxForm } from '../../reduxForm'`. I think that's how most libs do their examples. I wonder if that would be better than the current set up. I sort of like how I can confidently say that "this example project is using _this_ published version". Your thoughts?

ooflorent:
I see several options:

- using `npm link` (which causes issues when deps got duplicated)
- using relative paths
- using webpack aliases with relative paths

> I sort of like how I can confidently say that "this example project is using this published version"

Since each release has a git tag it is not an issue.

erikras:
Published as [`v6.0.0-alpha.13`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.13).

erikras:
> * using `npm link` (which causes issues when deps got duplicated)

Yes, `npm link` sucks. 👎 

> * using relative paths
> * using webpack aliases with relative paths

The aliases don't excite me that much, but the problem with relative paths is that the examples don't get to inherit any of `redux-form`'s deps. i.e. they would have to include them all. Still might be worth it, though.

Issue #971, opened 2016-05-15
v6: build error on test


bsr203:
just cloned the v6 branch.

```
git clone -b v6 --single-branch https://github.com/erikras/redux-form    
```

and `npm install` gives error  (`npm run build` is fine)

```
npm i                                                                                                                                                      v6
npm WARN deprecated jade@0.26.3: Jade has been renamed to pug, please install the latest version of pug instead of jade
npm WARN deprecated graceful-fs@2.0.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.

> fsevents@1.0.12 install /Users/bsr/Work/Programming/javascript/repo/redux-form/node_modules/fsevents
> node-pre-gyp install --fallback-to-build

[fsevents] Success: "/Users/bsr/Work/Programming/javascript/repo/redux-form/node_modules/fsevents/lib/binding/Release/node-v47-darwin-x64/fse.node" is installed via remote

> redux-form@6.0.0-alpha.10 prepublish /Users/bsr/Work/Programming/javascript/repo/redux-form
> npm run test && npm run clean && npm run build


> redux-form@6.0.0-alpha.10 test /Users/bsr/Work/Programming/javascript/repo/redux-form
> mocha --compilers js:babel-register --recursive --recursive 'src/**/__tests__/*' --require src/__tests__/setup.js

/Users/bsr/Work/Programming/javascript/repo/redux-form/src/__tests__/setup.js:11
global.navigator.product = 'NotReactNative';
                         ^

TypeError: Cannot set property product of [object Object] which has only a getter
    at Object.<anonymous> (setup.js:7:1)
    at Module._compile (module.js:413:34)
    at loader (/Users/bsr/Work/Programming/javascript/repo/redux-form/node_modules/babel-register/lib/node.js:158:5)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/bsr/Work/Programming/javascript/repo/redux-form/node_modules/babel-register/lib/node.js:168:7)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at /Users/bsr/Work/Programming/javascript/repo/redux-form/node_modules/mocha/bin/_mocha:310:3
    at Array.forEach (native)
[...]

Issue #967, opened 2016-05-14
Migration to v6 from v5 basic questions


brotzky:
Hi, I'm working on migrating from v5 to v6 and have some questions. 

1. Within the new Field what is the best way to handle onChange? In v5 I had ```<input onChange={this.func()}  {...fieldName} />```, but that seems to break for me with Field wrapper.

2. What's the best way to use field values within a component class? In v5, with the fields array, I would often go ```const { fields: { inputField } } = this.props;```at the top of a method and could then use inputField.value, for example. With the new Fields way are we supposed to mapStateToProps and then pull the value directly off props?

what I often did in v5 
```
 parseNumber() {
    const { fields: { inputField } } = this.props;
    ...
  }
```

In the documentation, I think there may be a small typo in the examples for migrating from v5 to v6. It may have been on purpose, but confused me initially. 
http://redux-form.com/6.0.0-alpha.8/docs/MigrationGuide.md/
```
export default reduxForm({
  name: 'myForm' // <-- Is this supposed to be **form**: 'myForm' ?
})(MyForm)
```
http://redux-form.com/6.0.0-alpha.8/examples/simple/
```
export default reduxForm({
  form: 'simple'  // a unique identifier for this form
})(SimpleForm)
```

Thanks guys. Migrating looked straight forward, but.. 🙈

erikras:
Right on, on the typo. Thanks for that. It's so easy to make mistakes in documentation code, because you don't actually run it. This is why I took great pains to be sure that the code shown on the examples pages _is actually the same code that is running_.

To your earlier questions...

1. That specific form would not have worked in `v5` either, because the subsequent destructuring of `{...fieldName}` into the `<input>` will overwrite the previous `onChange` prop. I keep meaning to write up a doc page about prop destructuring, because the React documentation is pretty light about it, and it causes headaches.
2. Yes, this is a fairly big difference, and probably the most annoying thing about moving from `v5` to `v6`. You are exactly right that the answer is to `connect()` to the form values manually. However, I'm planning on making a selector to make that more convenient. Something like:

```js
import { reduxForm, formValue } from 'redux-form'

@connect(
  state => ({
    aValue: formValue(state, 'myForm', 'aValue'),
    bValue: formValue(state, 'myForm', 'even.deeper.bValue'),
  })
)
@reduxForm({
  form: 'myForm'
})
class MyForm extends Component {
  calculateSomething() {
    const { aValue, bValue } = this.props
    return aValue + bValue
  }
 
[...]

Issue #945, opened 2016-05-11
v6: Form state is empty using @connect 


anaibol:
On the last realease, alpha 8, the form state seems to be empty using connect:

![image](https://cloud.githubusercontent.com/assets/3210177/15187389/3fff1790-17a1-11e6-9980-2909eeb63f80.png)

```javascript
@connect(state => {
  console.log(state);
  return ({ type: state.form.adWizard.values.type })
})
@reduxForm({ form: 'adWizard' })
export default class extends Component {
```

erikras:
The form structure in the reducer doesn't get initialized until the first `redux-form` action. This has been the behavior for a long time.

However, with `v6`, the pattern you are using will be very common. One idea I have to help with this is to have a `getFormValue()` selector that would fail nicely to `undefined` when the value isn't there yet.

```js
import { reduxForm, getValue } from 'redux-form';

@connect(
  state => ({
    type: getValue('adWizard', 'type')
  })
)
@reduxForm({ form: 'adWizard' })
export default class AdWizardForm extends Component {
}
```

But it doesn't exist it. 😦 Would be a good first feature for someone to PR.

browniefed:
This would essentially be a replacement for the `values` props that was passed down previously correct?
In your example not sure how `getValue` is actually accessing store state there?

erikras:
Published my "form value selector" API as [`v6.0.0-alpha.11`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.11). Feedback welcome!

PR #972, opened 2016-05-15
v6: getFormValue() to connect conveniently

fredrikolovsson:
@erikras I noticed you wrote that the `getFormValue()` selector mentioned in https://github.com/erikras/redux-form/issues/945#issuecomment-218522902 and https://github.com/erikras/redux-form/issues/967#issuecomment-219205319 could be a good first feature for someone to PR and I made an attempt.

Since the syntax is not set in stone, I am fully aware that the code in this PR could be completely discarded or require major refactoring. Also, please excuse me (and enlighten me!) if there are any beginner mistakes in the code or stupid questions below.

Some considerations when writing this:

**1. Handle different form reducer mount points:** I could not figure out a way to access `getFormState()`, hence opted to to ask the user to provide the form state slice (normally `state.form`).

**2. Handle Immutable and plain state:** I have not worked with ImmutableJS and am not sure if an immutable state looks different than in the tests, but if they are OK I figure `getIn()` from `/structure/immutable` can handle both an immutable and a plain state.

**3. Have a clear function signature:** Sometimes 2-3 unnamed arguments can be harder to read/understand than having an object with named arguments, though it requires more typing. I went with `getFormValue({ formState, formName, field })`, but of course `getFormValue(state, form, field)` could be an option. A slightly different approach, to allow removing one argument, could be `getFormValue(formState, field)` where `formState` would be one step deeper, like so: `state.reducerMountPoint.formName`.

What are your thoughts & suggestions?

Update: Here's the error from the CI, it is weird because `npm run test` worked fine locally while I was coding. After seeing this error, I tried making a new clone and got the error locally too. Commenting out the line with the error makes the tests run.
screenshot 2016-05-16 00 49 58
Update 2: CI error discussed in https://github.com/erikras/redux-form/issues/971 and https://github.com/erikras/redux-form/pull/977

### DIFF
--- a/src/__tests__/getFormValue.spec.js
+++ b/src/__tests__/getFormValue.spec.js
@@ -1,0 +1,115 @@
+import expect from 'expect'
+import { Map, List } from 'immutable'
+import getFormValue from '../getFormValue'
+
+describe('getFormValue', () => {
+  it('should throw if any argument is missing', () => {
+    const formState = {
+      myForm: {
+        values: {
+          fieldName: 'value'
+        }
+      }
+    }
+    const formName = 'myForm'
+    const field = 'fieldName'
+
+    expect(() => {
+      getFormValue()
+    }).toThrow()
+    expect(() => {
+      getFormValue({})
+    }).toThrow(/Missing argument/)
+    expect(() => {
+      getFormValue({ formState, formName })
+    }).toThrow(/Missing argument/)
+    expect(() => {
+      getFormValue({ formState, field })
+    }).toThrow(/Missing argument/)
+    expect(() => {
+      getFormValue({ formName, field })
+    }).toThrow(/Missing argument/)
+  })
+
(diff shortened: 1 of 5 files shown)


erikras:
@fredrikolovsson Thank you very much for picking up on my hint and taking this initiative, and you did a commendable job, especially with explaining your motivations. However, before I saw this PR, I had an idea of how to best implement this selector. Your way is fine, but I think my way feels little more elegant to use. I am, as always, open to comments, criticisms, and suggestions.

Here's what I came up with: [`formValueSelector()`](http://redux-form.com/6.0.0-alpha.11/docs/api/FormValueSelector.md/)

Again, thanks for taking this initiative, but I'm not going to end up using your code.

fredrikolovsson:
@erikras thanks for the feedback, your solution looks way more elegant, haha. From a learning perspective it was good to have made an attempt before seeing your code 😄

Issue #908, opened 2016-05-04
Submit with a redux action (v5 & v6)


clayne11:
The docs talk about submitting from a parent, but that requires having a ref to the form. I'm finding myself more and more with multiple containers wrapping around some of my components as HOCs. This means having refs to a form instance is often pretty ugly and hacky and relies on implementation details of the UI which I don't believe it should have to.

There's currently no method for calling `handleSubmit` from a redux action. If there was, you could submit the form from anywhere in the app by simply firing a redux action and you don't have to worry about the implementation of the UI.

An example of where this is very valuable:

I have a form which I use in multiple different locations. In one of the locations, there are "forward / back" buttons at the bottom of the page. The forward button should submit the form (if it is valid) and move on the to the "next" page. If the form is not valid, however, I don't want to move on to the next page.

Currently I don't see a way to do this through redux. Since the "forward / back" buttons aren't part of the form itself, but actually part of the form's container, and since the form might have data fetching containers (or other HOCs) around it, I can't directly access it with a ref to call `submit()`. There's also no way that I can see to tell if the form is valid (or pass down async validation errors) without calling `handleSubmit`.

Using a redux action seems like a much cleaner solution to me than using `ref`.

erikras:
The problem is that the `onSubmit` function lives with the component. So how would dispatching an action through the reducer result in the component submitting?  The `SUBMIT` action would have to set some "submit as soon as you see this" flag in the Redux store, and the component (only one of the potentially many connected form components) would have to respond to it. It gets very messy.

But alternative suggested solutions are welcome.

clayne11:
I think it can be done better than that.

If you fully decouple your UI from your business logic, and all of your validation and submit logic is in the redux reducer, you will have access to all of the validation and submission logic from anywhere in your app where you have access to `dispatch`.

In order to submit, the [`handleSubmit`](https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L6) function takes in data and actions from the redux store, as well as a few user provided functions: `validate`, `submit`, and `asyncValidate`.

If we turn `handleSubmit` into a redux action, as long as we provide the form name and the functions `validate`, `submit`, and `asyncValidate` when invoking the action it should function exactly the same way as it does invoking `handleSubmit` from within the form.

[...]

PR #973, opened 2016-05-16
Move registered field state off of the form and into the redux store

clayne11:
This pull request moves the registered field state out of the Form component and into the redux store.

It also moves the validation logic out of the fields and into the Form component.

This is the first step towards fixing #908.


src/__tests__/reduxForm.spec.js
```diff
@@ -185,25 +185,25 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
       const spy = createSpy()
       const { dispatch } = propChecker({}, spy, {
         validate: values => {
-          const animal = getIn(values, 'animal')
-          return animal && animal.length > 5 ? { animal: 'Too long' } : {}
+          const foo = getIn(values, 'foo')
+          return foo && foo.length > 5 ? { foo: 'Too long' } : {}
```
erikras: You got something against animals? 🐘 🐶 🐱 🐴 

I'm just kidding. Just amused by arbitrary change. 😃 Don't change it back.

animals

clayne11: It's not arbitrary - the validation now uses `registeredFields` to determine which fields are required to be validated, and since `foo` is what's mounted in the fixture, not `animals`, it was not validating the field.

I had to change the name to match.

erikras:
This looks very, very good. 👍 

Two things:

1. I'm gonna totally mess you up by merging #964 first, so you'll have to do redo some of your changes around the `connect()` call in `reduxForm.js`.
2. Before I can merge this, I'd like you to "squash" your commits. Don't ask me how that's done, but it seems to be common good PR practice.

src/__tests__/reduxForm.spec.js
erikras: Aha! Good call, then. 👍

clayne11:
OK no problem. I'll rebase my changes off of master and squash it.

clayne11:
Rebased and squashed. Ready to go.

erikras:
Rock on, @clayne11! Looks great. 🎉

Issue #970, opened 2016-05-15
Trouble setting up development environment


clayne11:
I'm working on a PR for #908 and I'm having trouble getting my environment set up. Here's what I've tried:

1. Forked and cloned `redux-form` and checked out `ebaeca3` which is `v6.0.0-alpha.10`.
1. In the `redux-form` directory, run `npm link`.
1. In my project ruin `npm link redux-form`.

When I run my app and navigate to a form I get the following error:

```
Uncaught Invariant Violation: removeComponentAsRefFrom(...): Only a ReactOwner can have refs. You might be removing a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner).
```

I also tried running using one of the examples with the same steps (using my local development version of `redux-form`) and I'm getting the same error. What am I doing wrong?

erikras:
That is the proper theory, as I understand it, but `npm link` has never worked well for me. I normally do:

```
cd whatever/redux-form
npm run build
cp -r dist/* ../my-project/node_modules/redux-form/dist
```
and then restart `my-project`, because the hot reloading does not always work.

erikras:
This is why, when I'm trying to implement a new feature or fix a bug in `redux-form`, I generally write a test to make sure it works. Not only is that a better way to code, but it lets me avoid all the project context switching crap of copying the built `lib` dir.

clayne11:
Fair enough. Thanks for the help on this.

PR #989, opened 2016-05-17
v6 — Use local redux-form

### LINKED ISSUE #983: [v6] Switch examples to use relative path for redux-form dep
Per discussion here, https://github.com/erikras/redux-form/pull/964#issuecomment-219069981, I think it might be a superior design to have [the `v6` example projects](https://github.com/erikras/redux-form/tree/v6/examples) import their version of `redux-form` via a webpack alias, [like the `react-router` examples do](https://github.com/reactjs/react-router/blob/master/examples/webpack.config.js#L31-L35). 

It would make it easier to immediately test the library as it is being developed and alleviate confusion like #970.

If you would like to take this on, leave a note that you are working on it in the comments. It should not be _too_ difficult. Your efforts would be most appreciated.

ooflorent:
Fixes #983
Relates to #970

### DIFF
--- a/examples/asyncValidation/package.json
+++ b/examples/asyncValidation/package.json
@@ -1,11 +1,5 @@
 {
-  "name": "redux-form-async-validation-example",
-  "version": "6.0.0-alpha.13",
-  "description": "An async validation example for redux-form",
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/erikras/redux-form"
-  },
+  "private": true,
   "scripts": {
     "clean": "rimraf dist",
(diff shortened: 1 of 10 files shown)


erikras:
Ha! No webpack trickery needed. 👍

ooflorent:
It is even better since npm will proceed like a classic install.
However, there is something really annoying, it will re-run `prepublish` of redux-form.

erikras:
Maybe I'm asking for too much, but it would be nice if I could tweak something in `reduxForm.js` and have it hot reloaded into my running example. I wonder if that's possible somehow? A symlink?

PR #992, opened 2016-05-17
v6 — Remove extraneous file

ooflorent:
I don't what this file is but I think it should be removed 😆


erikras:
Was the start of something and accidentally got pushed. 😆

PR #994, opened 2016-05-17
v6 fix doc example of FormValueSelector

eliseumds:
Object destructuring assignment had wrong variables names


erikras:
Thanks! 👍

Issue #1001, opened 2016-05-18
Race condition between blur and focus causes active state to be lost


kpdecker:
In my app I have a meta field that consists of multiple other fields (in this case, checkboxes). Rather than sending many focus and blur operations, I try to filter this to a single external `onFocus` and `onBlur` call via a setTimeout, i.e:

```
  onFocus() {
    if (!this.props.onFocus) {
      return;
    }

    clearTimeout(this.blurTimeout);

    if (!this.focused) {
      this.focused = true;
      this.props.onFocus(this.props.value);
    }
  }
  onBlur() {
    if (!this.props.onBlur) {
      return;
    }

    this.blurTimeout = setTimeout(() => {
      this.focused = false;
      this.props.onBlur(this.props.value);
    }, 50);
  }
```

This pattern apparently exposes a race condition between `onBlur` and `onFocus` calls between two fields. The timeout above causes `onBlur` to occur after `onFocus` for the field that receives focus, causing the `_active` state to be removed even though the other field still maintains focus at the user level. So my question is if the blur handler should check to ensure that it's not interacting with the form-level state when it's no longer the target field?

Something like:

```
  [BLUR](state, {field, value, touch}) {
    // remove _active from state
    const {_active, ...stateCopy} = state;  // eslint-disable-line prefer-const
    if (field !== _active) {
      stateCopy._active = _active;
    }
```

erikras:
In my experience, the `BLUR` is _always_ dispatched before the `FOCUS`. Obviously, your `setTimeout` call messes with that.

>  I try to filter this to a single external `onFocus` and `onBlur` call via a `setTimeout`

Why? This sounds like [probably-unwarranted fear of dispatching many actions](https://github.com/erikras/redux-form/issues/991#issuecomment-219774646). Can you explain why you think this is a good idea?

kpdecker:
This is to avoid marking the fields as touched until they are "fully" blurred. Basically I don't want some of the validation logic that is place to trigger until the user is focused on something completely different.

ooflorent:
I had experienced this race condition in an application. The issue is not that `blur` is dispatched after `focus` but the way the reducer handles the active field.

```js
// This is not a working snippet.
// It illustrates what's going on here.

// Action log
focus("form", "foo")
focus("form", "bar")
blur("form", "foo")

// Result state
{
  foo: {_active: false},
  bar: {_active: false}, // expected `true`
}
```

ooflorent:
[...]

PR #1002, opened 2016-05-18
blur should do anything if it does not target the active field

ooflorent:
This PR adds a test to demonstrate the issue exposed in https://github.com/erikras/redux-form/issues/1001#issuecomment-220128834.
It also fixes the issue #1001.

**DO NOT MERGE if this is the expected behavior!**
**DO NOT MERGE if we consider this bug as a _feature_ because it lives for too long!**

### DIFF
--- a/src/__tests__/reducer.spec.js
+++ b/src/__tests__/reducer.spec.js
@@ -1,4 +1,5 @@
 import expect from 'expect';
+import {createStore} from 'redux';
 import reducer, {globalErrorKey} from '../reducer';
 import bindActionData from '../bindActionData';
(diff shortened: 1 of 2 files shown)


erikras:
This is definitely the right way to handle this. 👍

erikras:
Can you implement this in `v6`, too, if it's not already?

erikras:
`< .01%` code coverage increase! 🎉 😄

ooflorent:
> Can you implement this in v6, too, if it's not already?

I'll do it tomorrow!

erikras:
Released as [`v5.2.4`](https://github.com/erikras/redux-form/releases/tag/v5.2.4).

PR #1004, opened 2016-05-19
blur should not change active field if the blurred field is not active

ooflorent:
This the v6 implementation of #1002


erikras:
Released in [`v6.0.0-alpha.14`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.14).

PR #1025, opened 2016-05-23
Ensure initial values are present on first render

### LINKED ISSUE #621: Form values not initialized from initialValues until 2nd render cycle

yesmeck:
Attempt to fix #621

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -35,7 +35,7 @@ describe('createReduxForm', () => {
   const expectField = ({ field, name, value, initial, valid, dirty, error, touched, visited, readonly, autofilled }) => {
     expect(field).toBeA('object');
     expect(field.name).toBe(name);
-    expect(field.value).toEqual(value === undefined ? '' : value);
+    expect(field.value).toEqual(value);
     if (readonly) {
       expect(field.autofill).toNotExist();
       expect(field.onBlur).toNotExist();
@@ -99,8 +99,8 @@ describe('createReduxForm', () => {
     expectField({
       field: stub.props.fields.foo,
       name: 'foo',
-      value: undefined,
-      initial: undefined,
+      value: '',
+      initial: '',
       valid: true,
       dirty: false,
       error: undefined,
@@ -111,8 +111,8 @@ describe('createReduxForm', () => {
     expectField({
       field: stub.props.fields.bar,
       name: 'bar',
-      value: undefined,
-      initial: undefined,
+      value: '',
+      initial: '',
       valid: true,
       dirty: false,
       error: undefined,
@@ -182,7 +182,7 @@ describe('createReduxForm', () => {
       field: stub.props.fields.foo,
       name: 'foo',
       value: 'fooValue',
-      initial: undefined,
+      initial: '',
       valid: true,
       dirty: true,
       error: undefined,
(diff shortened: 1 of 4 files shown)


erikras:
👍 @yesmeck

erikras:
Published as [`v5.2.5`](https://github.com/erikras/redux-form/releases/tag/v5.2.5).

cpsubrian:
👍  FWIW fixed my issue with checksum fail on server-side-rendered forms.

mpolichette:
@yesmeck You're my hero! thanks!

PR #1026, opened 2016-05-23
Mention SubmissionError in onSubmit docs

epeli:
Does not work without it.


erikras:
Oversight on my part. 🙈   👍 Thanks.

Issue #961, opened 2016-05-13
v6 (alpha 9): Some input fields lose focus after typing


jervi:
Fixing issue #936 solved a lot of problems I had with validation across multiple fields (for instance when one field is required based on the value of another field).
However, some of my custom input fields now lose focus after typing. This did not happen in alpha < 9.

```JSX
        <Field name='lumpSum' component={lumpSum =>
            <InputBox id='lumpSum' label='Lump sum' type='number' white={true} size='small' {...lumpSum}>
                {lumpSum.touched && lumpSum.error && <ValidationTooltip>{lumpSum.error}</ValidationTooltip>}
            </InputBox>
        }/>
```
`InputBox` looks like this:

```JSX
const InputBox = ({ id, label, help, value, size, white, children, ...rest }) =>
    <div className={ classNames('form-group', `input-${size}`)}>
        <label className={classNames({'hide' : !label})} htmlFor={id}>{label}</label>
        <input
            {...rest}
            className={classNames({'form-control' : !white, 'form-control-alternate' : white})}
            id={id}
            value={value}
            />
        {children}
        {help && <div className="help-text-under short-text">{help}</div>}
    </div>;
```

erikras:
I have seen this. It happens on the sync validation example on [`v6.0.0-alpha.10`](http://redux-form.com/6.0.0-alpha.10/examples/syncValidation/), but not on [`v6.0.0-alpha.8`](http://redux-form.com/6.0.0-alpha.8/examples/syncValidation/). It must be related to #940 somehow. @leesiongchan?

jervi:
Yeah, that is the same behavior. It only happens after I write the first character, and not on all fields.
Seems like it only happens when the state is populated for the first time or something. In the sync validation example, it happens on all fields, but _only_ if the two other fields are blank.

erikras:
It's related to the `valid` and `dirty` flags getting flipped, I think. Very, very strange.

leesiongchan:
Ok, I will look into it ASAP.
#tgif

fredrikolovsson:
For what it's worth, I encountered the same problem when having a higher order component that consumed the `dirty` prop from reduxForm() and caused a render when it flipped. The setup was like this:
```javascript
export default reduxForm({
  form: 'formName'
})(higherOrderComponent(FormComponent));
```
[...]

PR #1029, opened 2016-05-23
Update examples to avoid fat arrow component creation in render()

erikras:
Addresses decision made after discussion in #961.


erikras:
Released as [`v6.0.0-alpha.14`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.14).

PR #1033, opened 2016-05-24
treats null the same as undefined when setting initial values

yesmeck:
Same as this commit https://github.com/erikras/redux-form/commit/476a4e0517166489a079e263327e65d4220da90b on v6.

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -2760,4 +2760,34 @@ describe('createReduxForm', () => {
     // FAILS
     //expect(lastPrevBarValue).toNotEqual(lastNextBarValue);
   });
+
+  it('treats null the same as undefined when setting initial values', () => {
+    const store = makeStore();
+    const form = 'testForm';
+    class NullInitialValuesTestForm extends Component {
+      render() {
+        const {fields: { name }} = this.props;
+        expect(name.initialValue).toBe('');
+        expect(name.value).toBe('');
+
+        return (
+          <div>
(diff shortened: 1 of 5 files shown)


ooflorent:
The following will not work properly on `select` if `multiple` is `true` because `value` has to be defaulted to `[]` otherwise React will raise a warning.

Let's consider the following component:

```js
const SomeSelect = field => (
  <select multiple {...field}>
    <option value="a">A</option>
    <option value="b">B</option>
    <option value="c">C</option>
  </select>
)
```

If `value` is `null` or `undefined` then the component is not managed:

```md
Warning: `value` prop on `select` should not be null. Consider using the empty 
string to clear the component or `undefined` for uncontrolled components.
```

If `value` is `""` (empty string) then `PropTypes` are not met:

```md
Warning: The `value` prop supplied to <select> must be an array if `multiple` is true.
```

ooflorent:
Why did you create `isNil` instead of using the same trick as Erik did on v6?

yesmeck:
> Why did you create isNil instead of using the same trick as Erik did on v6?
 
Just don't want repeat `eslint-disable-line eqeqeq`

yesmeck:
I have not looked into v6, do you known how v6 handle default value for multiple select?

yesmeck:
> The following will not work properly on select if multiple is true because value has to be defaulted to [] otherwise React will raise a warning.

IMHO, if people want to use multiple select, they should given a initial array value to the field. This behaviour is same like v6 dose.

PR #1069, opened 2016-05-29
Adding react-widgets example

mihirsoni:
Hi @erikras 

I have added an example which will demonstrate how to use `redux-form` with `react-widgets` components. 

Mostly it works out of the box but we need to trick `onChange` and `onBlur` event as and when required.

### Notes

* `react-widgets` is heavily dependent on their css so there isn't exact UI in our example. I have tweaked root css from `index.html`.
* Their [MultiSelect](http://jquense.github.io/react-widgets/docs/#/multiselect?_k=os7mqb) component has a warning error for the `value` props as it expect `array` and we are passing string , though it works as expected.
* I haven't covered `DateTime` and `Number` Components (Please let me know if you would like to add) 
* I have also implemented how to initialize the form so that users have better understanding how to do that with `react-widgets`
* I have added `style-loader` and some of css related loaders in webpack as we are importing css. 
* I haven't used ExtractTextPlugin let me know if you would like to put into prod.config

### DIFF
--- a/examples/react-widgets/README.md
+++ b/examples/react-widgets/README.md
@@ -1,0 +1,10 @@
+# Simple Material UI Form Example
+
+## To run locally
+
+```
+npm install
+npm start
+```
+
+Then open [`http://localhost:3030/`](http://localhost:3030/).
--- a/examples/react-widgets/README.md
+++ b/examples/react-widgets/README.md
@@ -1,4 +1,4 @@
-# Simple Material UI Form Example
+# Simple React Widgets UI Form Example
 
 ## To run locally
 
--- a/examples/react-widgets/src/ReactWidgetsForm.js
+++ b/examples/react-widgets/src/ReactWidgetsForm.js
@@ -28,6 +28,36 @@ const validate = values => {
   return errors
 }
 
+const renderDropdownList = props => (
+  <div>
+    <DropdownList
+      {...props}
+      onChange={(value) => props.onChange(value)}
+    />
+    {props.touched && props.error && <span>{props.error}</span>}
+  </div>
+)
+const renderMultiselect = props => (
+  <div>
+    <Multiselect
+      {...props}
+      onChange={value => props.onChange(value)}
+      onBlur = {event => props.onBlur()}
+    />
+    {props.touched && props.error && <span>{props.error}</span>}
+  </div>
+)
+const renderSelectList = props => (
+  <div>
+    <SelectList
+      {...props}
+      onChange={(value) => props.onChange(value)} /* Simulating onChange manually*/
+      onBlur = {event => props.onBlur()} /* Simulating onBlur manually*/
+    />
+    {props.touched && props.error && <span>{props.error}</span>}
+  </div>
+)
+
 let ReactWidgetsForm = props => {
   const { handleSubmit, pristine, reset, submitting, load } = props
   return (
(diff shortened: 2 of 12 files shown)


examples/react-widgets/README.md
```diff
@@ -0,0 +1,10 @@
+# Simple Material UI Form Example
```
ooflorent: Could you change the title please?

examples/react-widgets/src/ReactWidgetsForm.js
```diff
@@ -0,0 +1,109 @@
...
+  const { handleSubmit, pristine, reset, submitting, load } = props
+  return (
+    <form onSubmit={handleSubmit}>
+    <div>
+        <button type="button" onClick={() => load(initialData)}>Initilalize</button>
+    </div>
+    <div>
+      <label>Favorite Color</label>
+      <Field name="favoriteColor"
+        component={ favoriteColor =>
```
ooflorent: Please don't use arrow functions in `component`.
Redux Form no longer supports this.

erikras:
@mihirsoni Please look again at the Material UI example. It changed a bit since you originally wrote it.

mihirsoni:
@erikras  Sure , Do I need to write it as Adapter ?

examples/react-widgets/src/ReactWidgetsForm.js
mihirsoni: @ooflorent  I wasn't aware of this API change. I will look around and modify it accordingly. Thanks for reviewing.

ooflorent: No problem! The announcement was made on the release notes:
https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.14

erikras:
No. Write it like the current Mat UI example is, stopping just short of making an adapter.

If you'd like to make a separate repo and lib for the adapter, you may.

ooflorent:
> If you'd like to make a separate repo and lib for the adapter, you may.

`adapter` is gone in a43233e365ad2bc3c0c85250c8b5bef7ca9a9a89.

erikras:
While `adapter` is gone – RIP 💀, we hardly knew ye – I _do_ think there is room for libraries like [`redux-form-material-ui`](https://github.com/erikras/redux-form-material-ui).

mihirsoni:
@erikras  Sure I will try to create ```react-widgets``` wrapper. 

@ooflorent  I have modified it sorry for little delay.

erikras:
I was able to greatly simplify this. [Check it out](http://redux-form.com/6.0.0-alpha.15/examples/react-widgets/). You don't even need render functions. It's almost so easy that it doesn't warrant an example. 😄

mihirsoni:
@erikras this is great and simplified amazingly 💃  I should have tried this. But yeah thanks for inputs.

mihirsoni:
@erikras One question though , how does that field takes our arguments as ```props``` ??

erikras:
Any additional props given to `Field` are passed along to the `component`. And they can even override (e.g. `onBlur`) the normal props that `Field` generates for `component`.

mihirsoni:
@erikras  make sense and this will be great approach as well so that we don't need a wrapper library or class to do that job.

Issue #1063, opened 2016-05-27
Poll: FieldArray API – passing through additional props


erikras:
Let's say your field has an array of hobbies, like [the current FieldArray example](http://redux-form.com/6.0.0-alpha.14/examples/fieldArrays/). At the moment, the API looks like:

```js
const renderHobbies = (fields) =>
  fields.map(field => <Field name={field} component="input" type="text"/>)
...
<FieldArray name="hobbies" component={renderHobbies}/>
```

The important thing to note, for the sake of this discussion, is that `renderHobbies` takes _only one parameter_, which is the pseudo-array object that provides methods like `map` and `forEach` to iterate through field names.

Okay, fine, but what if you want to pass a config parameter into `renderHobbies`. Let's say you want to limit the number shown for some reason, so you want to pass a `maxShown` prop.

Currently, this would be accomplished like this:

```js
const renderHobbies = (fields) =>
  fields.map((field, index) =>
    index < fields.maxShown ?
    <Field name={field} component="input" type="text"/> : 
    null)
...
<FieldArray name="hobbies" component={renderHobbies} maxShown={5}/>
```

Note that the `maxShown` prop gets attached _to the pseudo-array object_. On #1038, @tnrich has questioned how much sense that makes.

So we're having a poll. Using the new emoji response feature of GitHub, please vote for the comment below that best represents your preference. You may suggest alternatives and vote for the suggested alternatives.

erikras:
## Option 1

Leave it like it is, with the props being placed directly onto the pseudo-array.

```js
const renderHobbies = ({ maxShown, ...fields }) =>
```

erikras:
## Option 2

Put the props into a separate parameter

```js
const renderHobbies = (fields, props) =>
```

erikras:
## Option 3

Combine them both into an object, with a key of.... `fields`?

```js
const renderHobbies = ({ fields, maxShown }) =>
```

ooflorent:
In my opinion, option 3 is the best because `component` is a real React component unlike option 2. Option 1 would be a pain to implement because `fields` would not be an array...

erikras:
I think @ooflorent just nailed the argument against Option 2. That's why I went with Option 1 initially.

epeli:
I agree. Just changed my vote from 2 to 3.

MrEfrem:
[...]

PR #1074, opened 2016-05-30
Separate field array props

erikras:
As voted in #1063.

### DIFF
--- a/docs/api/FieldArray.md
+++ b/docs/api/FieldArray.md
@@ -52,71 +52,73 @@ The following properties and methods are available on an instance of a `FieldArr
 
 ## Props
 
-These are props that `FieldArray` will pass to your wrapped component.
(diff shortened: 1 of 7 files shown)


ooflorent:
It would be a good idea to refactor `createFieldArrayProps` because it creates multiple closures and is called on every render of `FieldArray`. Adding some kind of memoization would improve perf and prevent additional GC.

erikras:
@ooflorent I nominate you for doing that memoization. 😃

jayphelps:
@erikras is there any plan to make the same change to `Field` components? The same issue for passing along props exists and it's been awkward that they're inconsistent.

i.e. 

```js
const MCHammer = ({ field, hammerTime }) => (
  <input {...field} placeholder={hammerTime ? 'Stop, hammer time' : ''} />
);

// etc

<Field component={MCHammer} hammerTime />
```

Issue #887, opened 2016-04-29
Getting form values breaks when using a form field with the name of "value"


Aaronius:
I currently have a nested field named `value`:

`foo.bar.value`

When calling `getValues` for the form, `foo.bar.value` is not the value of the value field. The problem comes from https://github.com/erikras/redux-form/blob/e61517b23e4e3cde0c422658d624434fec18d1ea/src/getValuesFromState.js#L17-L20 when `getValues` is recursing through the fields. When it gets to the `bar` object, it sees that it has a value prop and assumes that `bar` itself is a field and that `value` is the value rather than detecting that `value` is actually a field.

Is there a better way to detect when a field is reached than just looking to see if has a `value` property?

erikras:
Yes, that's what [`isFieldValue()`](https://github.com/erikras/redux-form/blob/master/src/fieldValue.js#L11-L13) is for. This must have been overlooked.

Aaronius:
Thanks @erikras!

Aaronius:
I've confirmed that this has fixed the issue. Thanks again for fixing and publishing so quickly.

erikras:
You got lucky. 😄

johanneslumpe:
@erikras This seems to break the usage of `getValues` with forms which are initialized on the server and the hydrated on the client. I have a form which has a `type` field and that gets initialized on the server. On  the client i see the correct `initialValue` and `value` props in the reducer, but `_isFieldValue` is missing. This breaks part of the application I'm working on, as the field is no longer returned via `getValues`.

Was this intentional or is this a bug/edge case? I hope you will be able to provide some insights here, because this really caught us off-guard :/

MitchWeiss:
I can confirm @johanneslumpe's findings. After upgrading to 5.2.5 `getValues` returns an empty object. Downgrading to 5.2.2 before this change was made fixes the issue.

erikras:
Fix published in [`v5.3.0`](https://github.com/erikras/redux-form/releases/tag/v5.3.0).

Issue #1071, opened 2016-05-29
[V6] Field Arrays Example with immutablejs doesn't work.


tdziezyk:
I changed reducer. Instead of:

```js
import { Field, FieldArray, reduxForm } from 'redux-form'
```

I put:

```js
import { Field, FieldArray, reduxForm } from 'redux-form/immutable'
```

and example is displaying properly, but when I add club name and try type text into 'first name' field I get an error:

```
Uncaught Error: invalid keyPath
```

Is there any solution/work around for it?

clayne11:
Can you post a reproduction repo?

kkwiatkowski:
I got the very same issue, but in version 5.2.5.

As far as I can see, everything started with this commit 7b5cb0f64911dffd2ac80e254719ec75333cb8f4
and introduction of `fieldValue.js`.

`makeFieldValue` is assigning not enumerable property to field object, so when trying to make Immutable from such object, not-enumerable property is lost.
It does matter, because on store, there is no `_isFieldValue` property assigned to field object, and because of that `isFieldValue` is failing when used in [`src/getValuesFromState`](https://github.com/erikras/redux-form/blob/v5.2.5/src/getValuesFromState.js#L17).

This is not only related to Immutable.js, but also to any form of state serialization, for example JSON.stringify.
I wanted to make a pull request with easy fix to make this prop enumerable, but it comes with loads of tests changes.
I think that `makeFieldValue` and `isFieldValue` should be rewritten in some way to get rid of using not enumerable props, or adding any props at all.

I hope it helps.

erikras:
@kkwiatkowski Anything in `v5` is unrelated to the original issue posted here.

@tdziezyk Are you remembering to use a special [immutable-enabled version of `combineReducers` as well as the immutable version of the `redux-form` reducer](https://github.com/erikras/redux-form/blob/v6/examples/immutable/src/index.js#L5-L6)?

jccf091:
I got the same issue in v6.0.0-rc.1.
@erikras I did use the combineReducers from 'redux-immutable' and the immutable version of redux-form reducer.

jccf091:
tl;dr
look at the onclick logic for adding new field under your renderArrayFields function.
it should be like this:
```
<button type="button" onClick={() => fields.push(Immutable.Map())}>Add new member</button>
````
instead of:
```
<button type="button" onClick={() => fields.push({})}>Add new member</button>
````

[...]

PR #1077, opened 2016-05-30
Make _isFieldValue property enumerable in store

kkwiatkowski:
Fix of issue https://github.com/erikras/redux-form/issues/887 introduced another issue that didn't let you serialize state or use it with immutable.js. Whole background is presented below.

As far as I can see, everything started with this commit 87579e0, which was released with version 5.2.3.

In commit 7b5cb0f64911dffd2ac80e254719ec75333cb8f4 ([release 4.0.7](https://github.com/erikras/redux-form/releases/tag/v4.0.7)) two functions `makeFieldValue` and `isFieldValue` were introduced.
`makeFieldValue` is assigning not enumerable property to field object, so when trying to make Immutable from such object, not-enumerable property is lost.
It does matter, because on store, there is no `_isFieldValue` property assigned to field object, and because of that `isFieldValue` is failing when used in [`src/getValuesFromState`](https://github.com/erikras/redux-form/blob/v5.2.5/src/getValuesFromState.js#L17).

This is not only related to Immutable.js, but also to any form of state serialization, for example `JSON.stringify`.

That's why I've created this pull request, with simply patch that fixes enumeration of mentioned field. Unfortunately, it comes with loads of test changes.
In general this is only patch. To make it work well, it's probably required to rewrite whole concept of nested field.

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -1495,12 +1495,14 @@ describe('createReduxForm', () => {
     expect(store.getState().form.testForm.users[ 0 ].name)
       .toEqual({
         initial: 'Bob',
-        value: 'Bob'
+        value: 'Bob',
+        _isFieldValue: true
       });
     expect(store.getState().form.testForm.users[ 0 ].age)
       .toEqual({
         initial: 27,
-        value: 27
+        value: 27,
+        _isFieldValue: true
       });
   });
 
@@ -1794,7 +1796,8 @@ describe('createReduxForm', () => {
     expect(store.getState().form.testForm.children)
       .toEqual({
         initial: [ 3, 4 ],
-        value: [ 3, 4 ]
+        value: [ 3, 4 ],
+        _isFieldValue: true
       });
     // reset form to newly initialized values
     stub.props.resetForm();
@@ -1842,7 +1845,8 @@ describe('createReduxForm', () => {
     expect(store.getState().form.testForm.name)
       .toEqual({
         initial: 'Bob',
-        value: 'Bob'
+        value: 'Bob',
+        _isFieldValue: true
       });
     // set value
     stub.props.fields.name.onChange('Dan');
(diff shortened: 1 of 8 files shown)


erikras:
> To make it work well, it's probably required to rewrite whole concept of nested field.

I **_did_** "rewrite the whole concept of nested field". It's called [`v6.0`](https://github.com/erikras/redux-form/tree/v6).

As I stated on #1071, this has nothing to do with #1071 as it was originally reported, as #1071 is about `v6`..

`v5` never officially supported ImmutableJS, except through a hack.

`v5` never had any problems – that I know of (and I, and _many_ others, use it in production) – with serializing the store down to the client on server-side render.

Can you please write a test that would fail before your change, but passes after your change? I'd like to see some evidence for your assertion.

kkwiatkowski:
Yeah, I'll provide such test soon.

> v5 never had any problems – that I know of (and I, and many others, use it in production) – with serializing the store down to the client on server-side render.

It won't fail on server side rendering, it will fail on client side.
**Issue story:**
> I've upgraded React to version 15, with this upgrade I decided also to upgrade redux-form to latest available version.
After upgrade, I tried to submit login form in my application, and I couldn't believe that after upgrade no any data was sent to the server with user credentials.
After some investigation I found that fields data is "saved" to store, but it cannot be retrieved, because "isFieldValue" were failing.
After some code research I realized, it's because during immutable.js's object traversing, it is not possible to enumerate `_isFieldValue` property, so immutable.js is not aware of such property and because of that it's just omitting it. The same goes for JSON.stringify (not enumerable fields are omitted).
Because this field is omitted I am not able to fetch data from store using `getValues` method provided by redux-form package.

kkwiatkowski:
Tests cases added. I hope it's enough.

johanneslumpe:
Definitely need this 👍

erikras:
Looks good. Thanks, @kkwiatkowski! Might not get around to publishing until next week.

erikras:

Sorry about the delay, @kkwiatkowski. 😳

Issue #1066, opened 2016-05-28
v6 Field doesn't rerender when handleSubmit triggered


vslinko:
I have following form:

```typescript
export default class ExampleForm extends React.Component {
  render() {
    return (
      <form onSubmit={this.props.handleSubmit}>
        <Field
          name="name"
          component={(props) => (
            <TextField
              disabled={this.props.submitting}
              value={props.value}
              onValueChange={props.onChange}
            />
          )}
        />
```

Field doesn't updates after form submission and after end of form submission.
Am I doing something wrong?

erikras:
Yes, you are doing something wrong. (it was right before, but now is wrong) [Read this](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.14) and update to 6a14.

vslinko:
But how to pass extra props through `Field`?

erikras:
Yes, through `Field`. Any prop you give to `Field` will be passed on to the `component`.

vslinko:
Hmm. It's ok for JavaScript, but it hard to declare using TypeScript type system.
Why don't use one prop for passing extra props `<Field name="a" component={x} props={{a: 1}} />`?
I think it should be easier for everything: propTypes, typescript, documentation.
Also, what if I need pass `name` property?

erikras:
Seems like it'd be okay to make `'props'` a special prop name that would get destructured into the other props.

erikras:
Published fix as [`6.0.0-alpha-15`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.15).

PR #1078, opened 2016-05-30
Pass props through Field and FieldArray using props prop

vslinko:
ref https://github.com/erikras/redux-form/issues/1066


erikras:
Great PR. 👍  Can you do [the docs](https://github.com/erikras/redux-form/tree/v6/docs/api), too?

vslinko:
@erikras updated

ooflorent:
It looks really good! Maybe you could add a note about _why_ this prop was introduced.

erikras:
Yes, what @ooflorent said. ☝️

vslinko:
I don't know what to add as the reason.

erikras:

rluiten:
I think there may be a small issue here, if its a bug or not I could not say.

I was creating a helper component to output BooleanRadioField with label and error display area.

I had already discovered that I must pass in the property "type" to `Field` so that the logic in processProps() can create the checked property for me.

The `type` field works in either the Field properties or in the 'props' property.

For setting the value of a radio input I have tried both
 ```html
<Field name="option1" component="input" props={{type="radio", value=false}} />
```
and
 ```html
<Field name="option1" component="input" props={{type="radio", value="false"}} />
```
Neither set the value attribute on the input element.

Only this works for actually setting the radio button attribute named "value".
```html
<Field name="option1" component="input" props={{type="radio"}} value={false} />
```

CarsonF:
Is this still needed? I know there are a lot of changes for how props are passed in RC4. I haven't needed it for my TypeScript definition. And IMO it just makes things more confusing.

rluiten:
In my opinion it is not needed, i was just using at one point.

erikras:
📟 @vslinko

vslinko:
I'm not using v6 right now, so I don't have solid opinion on that.

erikras:
Well, @CarsonF is the current `v6` typescript expert, so I'm inclined to go with him.

CarsonF:
@vslinko [Here's](https://github.com/erikras/redux-form/pull/1318#issuecomment-236655775) the example I gave @rluiten. The concept can be applied to any component. The example creates a class so it's reusable but really only the first two lines inside `render()` that are needed if you are doing a one-off.

PR #1091, opened 2016-06-02
fix(readField): Generate correct prefix when reading inside array fields, fixes #1090

### LINKED ISSUE #1090: [v5] addField on nested fields doesn't work at leaf of `first[].second.third[]`
Similar to #630, but with a different property path structure.

`first[].second.third[].addField({ forth: "DOESNT WORK" })`

While most nested structures work, for whatever reason this particular order of them does not. Just like #630, the `fields` array in the `ADD_ARRAY_VALUE` action is empty.

JSBin: http://jsbin.com/wodoxe/edit?js,output
```js
const Form = reduxForm({
  form: 'example-form',
  fields: [
    'first[].second.third[].forth'
  ],
  initialValues: {
    first: [{
      second: {
        third: []
      }
    }]
  }
})(
  ({ fields: { first }, values }) =>
    <div>
      <pre>{JSON.stringify(values)}</pre>
      <button
        onClick={() => first[0].second.third.addField({ forth: 'DOESNT WORK' })}
      >
        Add
      </button>
    </div>
);
```

No errors.

jayphelps:
fixes #1090

The issue was that the prefix passed to `readField` wasn't being generated like it is for object fields. e.g. incorrect: `proposals[0].meta.items[].` correct: `proposals[].meta.items[].`

That is, `pathToHere` *may* actually contain indices for deeply nested stuff so you cannot use it.

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -1573,7 +1573,8 @@ describe('createReduxForm', () => {
         'proposals[].note',
         'proposals[].rooms[].name',
         'proposals[].rooms[].adults',
-        'proposals[].rooms[].children'
+        'proposals[].rooms[].children',
+        'proposals[].meta.items[].name'
       ]
     })(Form);
     const dom = TestUtils.renderIntoDocument(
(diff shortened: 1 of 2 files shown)


ooflorent:
LGTM

erikras:

PR #1100, opened 2016-06-05
Events originating from numeric inputs should return numbers, not strings

whawker:


erikras:
Very interesting. Thanks! I'll fix the linting error.

whawker:
Fantastic, thanks @erikras (and thanks for this project!) When is the next release likely to happen?

erikras:

whawker:
@erikras is this change likely to be released on the v5.2.x branch? Or is it too much of a breaking change?

whawker:
Hi @erikras, any chance this could get be released on v5? Whether that be v5.3.0 or v5.2.6 ?

erikras:

tomazy:
I'm afraid this commit makes validations of numeric fields impossible as anything that is not a valid number entered into an input will be converted to `NaN`. It's impossible to distinguish if the input is empty (`''`) or has some gibberish like "e32eee4e3". 
This breaks `isNumeric()` validator from `revalidate` library. 
It breaks `isRequired()` from the same library too, as `NaN` doesn't equal `null` nor empty string.

erikras:
`NaN` is falsy, though. That's usually how I write my `isRequired()` functions.

The old functionality is still available if you just change your type to `"text"`.

whawker:
Just to point out, if @tomazy is using the `v5.3.0` release, looks the the fix from #1136 isn't included on that branch. Maybe merging in that fix would help?

tomazy:
Oh, sorry. I forgot to mention that I'm using the latest v5.3.1 release.

@erikras 
I'd rather avoid writing my own libraries for validation if I can use existing ones. 

> NaN is falsy, though. That's usually how I write my isRequired() functions.

I will get NaN for "33ee22" too - in this case I want to show a message that the input is invalid, not that it is missing. 

If there is "33ee22" in the input field I would like be able to tell the user: "'33ee22' is not a valid number" but I can't access the original input's value.

#1136  partially fixes the problem IMHO. You can't be consistent here if sometimes a value can be `''`, sometimes `NaN` and sometimes `42`

> The old functionality is still available if you just change your type to "text".

Having type "number" makes more sense in my app. It's 2016 ;)

manubamba:
It adds a new problem that long numbers will not render correctly.

for eg: 12345678901234567 will get converted to 12345678901234568

Issue #1081, opened 2016-05-31
expose the original component for testing purposes


paooolino:
Would it be possible to access the original component as described here?
http://stackoverflow.com/questions/31459720/how-to-test-decorated-react-component-with-shallow-rendering#comment-61693171

I'm applying shallow rendering to a component which is parent of the decorated redux-form component. I get a component of type: "ConnectedForm" even if my original component is called "LoginFormContainer".

src:

    render() {
        return (
            <div><LoginFormContainer /></div>
        )
    }

test:

    const renderer = TestUtils.createRenderer();
    renderer.render(<App isAuthenticated={false} />);
    const output = renderer.getRenderOutput();

    expect(output.props.children.type.WrappedComponent.name).toBe('connectedForm');`

Having a way to get the original wrapped component would be nice, e.g.

    expect(output.props.children.type.WrappedComponent.originalComponent.name).toBe('LoginFormContainer');

andrewmclagan:
Was in the process of upgrading our stack to 6.x.x and noticed that testing is a little more difficult without this. Was intending on writing some testing docs in a PR.

erikras:
Published fix as [`6.0.0-alpha-15`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.15).

PR #1101, opened 2016-06-05
Provide wrapped ref for testing. #1081

erikras:
Addresses #1081.


erikras:

PR #1108, opened 2016-06-07
Update ESLint

ooflorent:
Also bump side dependencies


.eslintrc
```diff
@@ -1,19 +1,18 @@
 {
   "extends": "eslint-config-rackt",
+  "parser": "babel-eslint",
```
ooflorent: Explicitly use Babel for parsing

.eslintrc
```diff
@@ -1,19 +1,18 @@
...
-    "react/jsx-uses-react": 1,
+    "react/jsx-uses-react": 2,
+    "react/jsx-uses-vars": 2,
     "react/jsx-no-undef": 2,
     "react/wrap-multilines": 2
   },
   "plugins": [
     "react"
-  ],
-  "globals": {
```
ooflorent: This global variable is not used

src/reduxForm.js
```diff
@@ -255,6 +255,7 @@ const createReduxForm =
 
           render() {
             // remove some redux-form config-only props
+            /* eslint-disable no-unused-vars */
```
ooflorent: Here we use destructuring to compute `passableProps`. We need this to mute ESLint because we create a lot of unused bindings.

erikras: That's dumb on eslint's part.

ooflorent: This won't be addressed until _Rest/Spread Properties_ would progress to higher stage.
See https://github.com/eslint/eslint/issues/4880#issuecomment-169860909.

PR #1115, opened 2016-06-07
Added normalize prop to Field

erikras:
This accomplishes _most_ of what the `v5` [`reducer.normalize()`](http://redux-form.com/5.2.5/#/api/reducer/normalize) API was providing.

### DIFF
--- a/docs/api/Field.md
+++ b/docs/api/Field.md
@@ -47,10 +47,35 @@ Defaults to `''` to ensure that the input is
 If `true`, the rendered component will be available with the `getRenderedComponent()` method.
 Defaults to `false`. **Cannot be used if your component is a stateless function component.**
 
+#### `normalize : (value, previousValue, allValues, previousAllValues) => nextValue` [optional]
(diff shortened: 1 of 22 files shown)


erikras:
@ooflorent Opinions on this?

ooflorent:
I always disliked how normalization works until this patch. 
Moving this to `Field` was the right move. 

👍  on removing getters! I think you forgot to update the [Instance API documentation](https://github.com/erikras/redux-form/blob/v6/docs/api/Field.md#instance-api).

erikras:
Great.

>  I think you forgot to update the Instance API documentation.

I don't think so. I only removed getters on `ConnectedField`, not on `Field`, which is what the docs address.

ooflorent:
Oh my bad.

erikras:
Published as [`v6.0.0-rc.1`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.1).

PR #1117, opened 2016-06-08
v6 — Remove lodash's `partial` and `partialRight`

ooflorent:
We can substitute these helpers by `Function#bind` and closures.

Removing them ease the configuration of `LodashModuleReplacementPlugin` because it no longer requires `placeholders` to be enabled.


erikras:
Okay. The more native js we can use the better.

erikras:

PR #1120, opened 2016-06-08
Fixed bug where field was not rerendering when sync error changed

### LINKED ISSUE #1024: V6 - Field does not rerender when its sync error changes
I have an array editor, and a validation function.

When I make changes in 1 field, my validation function runs and based on the value of field 1, the validation for field 2 might change (it's not passwords comparison).
When I make changes to field 1, validation is updated only for that field, but when the validation function removes an error for field 2, that error is not removed in the UI.

erikras:
Fixes #1024.

Moved `getSyncError()` from `ConnectedField` to `Field`.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -20,14 +20,6 @@ const createConnectedField = ({
       return !deepEqual(this.props, nextProps)
     }
 
-    getSyncError() {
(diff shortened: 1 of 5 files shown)


erikras:
One of these days I will learn to run `eslint` _before_ I create the PR. 😧

erikras:

Issue #1089, opened 2016-06-01
Errors using v6 standalone builds


eddywashere:
I've been trying to set up a simple codepen/jsbin with the v6 alpha version but I kept running into errors. I had a feeling there was something missing in the standalone umd build but couldn't pin it down to something specific.

I did an `npm run build` from scratch, based off of the `v6.0.0-alpha.14` tag and noticed there were fewer dependencies included in the webpack build. See diff https://www.diffchecker.com/wvbidnqd

I tried out the dist files from that build and sure enough it worked fine.

Examples:

- [v6.0.0-alpha.14](https://npmcdn.com/redux-form@6.0.0-alpha.14/dist/redux-form.js) - [codepen](http://s.codepen.io/eddywashere/debug/pbvZdx) (broken example)
- [v6.0.0-alpha.14-updated](https://d3g7hm1jmpcdis.cloudfront.net/redux-form-v6.0.14.updated.js) - [codepen](http://s.codepen.io/eddywashere/debug/bpXeLP) (working example)

fusepilot:
Just ran into this same problem trying to create a repo on codepen for another issue.

If you downgrade the redux-form to ````https://npmcdn.com/redux-form@6.0.0-alpha.13/dist/redux-form.js```` it won't throw an error, but the text fields then don't accept user input.

jayphelps:
Receiving the same error using UMD build: http://jsbin.com/hawema/1/edit?js,output

```
TypeError: Cannot read property 'length' of undefined
    at arraySome (redux-form.js:6430)
    at redux-form.js:4001
    at a.configureFinalMapState (react-redux.min.js:1)
    at a.computeStateProps (react-redux.min.js:1)
    at a.updateStatePropsIfNeeded (react-redux.min.js:1)
    at a.render (react-redux.min.js:1)
    at ReactCompositeComponentMixin._renderValidatedComponentWithoutOwnerOrContext (12726956_279572999043766_1590663715_n.js:6491)
    at ReactCompositeComponentMixin._renderValidatedComponent (12726956_279572999043766_1590663715_n.js:6517)
    at ReactCompositeComponentMixin.performInitialMount (12726956_279572999043766_1590663715_n.js:6027)
    at ReactCompositeComponentMixin.mountComponent (12726956_279572999043766_1590663715_n.js:5920)
```

I need to report some v6 bugs and discuss feature issues, but without being able to create a JSBin makes it rather awkward to explain.

erikras:
[...]

PR #1136, opened 2016-06-10
fix(type=number): Don't parseFloat empty value strings, they NaN

jayphelps:
#1100 added automatic number parsing of `<input type="number">` fields but if the field is empty it currently applies `parseFloat` to an empty string, which produces `NaN` which could be argued either way as right or wrong, but this has the consequence of making the `<input>` warn in Google Chrome with:

> The specified value "NaN" is not a valid number. The value must match to the following regular expression: -?(\d+|\d+\.\d+|\.\d+)([eE][-+]?\d+)?

This happens on every re-render, once that field has been touched. Using `undefined` also invokes the warning, the only thing that doesn't is the default empty string or a string that is a valid number.

Because of #1089 I can't create a JSBin to reproduce, but here's some code:

```js
console.clear();

const { createStore, combineReducers } = Redux;
const { Provider } = ReactRedux;
const { reduxForm, Field, reducer: form } = ReduxForm;

/////////////////////////////////////////////////

const Form = reduxForm({
  form: 'example-form'
})(
  () =>
    <div>
      <Field name="count" type="number" component="input" />
    </div>
);

/////////////////////////////////////////////////


const store = createStore(
  combineReducers({ form })
);


ReactDOM.render(
  <Provider store={store}>
    <Form />
  </Provider>,
  document.getElementById('app')
);
```

If it's too awkward now the value might be an empty string or a parsed number, I think removing the parsing number behavior altogether is probably the right call. Then the behavior is the same as the browser. Especially because `parseFloat` has [known quirks](http://stackoverflow.com/a/13676265/1770633) that may causes issues in browsers that don't actually restrict you from typing non-number characters in the input (like Chrome does). e.g. they type `'1-dal@41'` and `parseFloat('1-dal@41') === 1`

### DIFF
--- a/src/events/__tests__/getValue.spec.js
+++ b/src/events/__tests__/getValue.spec.js
@@ -101,7 +101,7 @@ describe('getValue', () => {
     }, false)).toBe(false)
   })
 
-  it('should return a number type for numeric inputs', () => {
+  it('should return a number type for numeric inputs, when a value is set', () => {
     expect(getValue({
       preventDefault: noop,
       stopPropagation: noop,
@@ -134,6 +134,15 @@ describe('getValue', () => {
         value: '3.1415'
       }
     }, false)).toBe(3.1415)
+
+    expect(getValue({
+      preventDefault: noop,
+      stopPropagation: noop,
+      target: {
+        type: 'range',
+        value: ''
+      }
+    }, false)).toBe('')
   })
 
   it('should return event.target.files if file', () => {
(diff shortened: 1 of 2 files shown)


erikras:

antoinerousseau:
It also happens to me when I use a decimal separator that is accepted by Chrome but is not valid JS, e.g. if I try to type `12.3` with a French locale, since the separator should be `.`, the fields empties when I type `.` and I get the same alert than you.

gif

PR #1141, opened 2016-06-12
V6 - ShallowEqual need nextState

Darmody:
Because `shallowequal(undefined, null) === false`, should add mssing `nextState` param to `shallowCompare`

I believe some rerender cases are caused by this problem.

### DIFF
--- a/src/Field.js
+++ b/src/Field.js
@@ -16,7 +16,10 @@ const createField = ({ deepEqual, getIn, setIn }) => {
       this.normalize = this.normalize.bind(this)
     }
 
-    shouldComponentUpdate(nextProps, nextState) {
+    shouldComponentUpdate(nextProps, nextState, nextContext) {
+      if (this.context._reduxForm.syncErrors !== nextContext._reduxForm.syncErrors) {
+        return true
+      }
       return shallowCompare(this, nextProps, nextState)
     }
 
@@ -87,7 +90,7 @@ const createField = ({ deepEqual, getIn, setIn }) => {
       return createElement(this.ConnectedField, {
         ...this.props,
         normalize: this.normalize,
-        syncError: this.getSyncError(),
+        syncError: this.getSyncError(this.context),
         ref: 'connected'
       })
     }
--- a/src/Field.js
+++ b/src/Field.js
@@ -16,8 +16,8 @@ const createField = ({ deepEqual, getIn, setIn }) => {
       this.normalize = this.normalize.bind(this)
     }
 
-    shouldComponentUpdate(nextProps) {
-      return shallowCompare(this, nextProps)
+    shouldComponentUpdate(nextProps, nextState) {
+      return shallowCompare(this, nextProps, nextState)
     }
 
     componentWillMount() {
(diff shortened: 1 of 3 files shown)


erikras:
That was intentionally left out because these components are stateless, and therefore `this.state` and `nextState` will always be `undefined`. Have you seen a case where one is `null` and the other is `undefined`?

Darmody:
@erikras Yep, I check it in my project, `this.state` is null.And I think null is the default value of `this.state`?

src/__tests__/Field.spec.js
```diff
@@ -606,61 +606,61 @@ const describeField = (name, structure, combineReducers, expect) => {
     // ----------------------------------------------
     // Uncomment this to confirm that #1024 is fixed.
     // ----------------------------------------------
```
erikras: Why did you comment this out again?

Darmody: Because it failed, I think maybe it has not been fixed correctly.

Darmody: @erikras Oh, I think that syncError is not related to props, so when I avoid the unnecessary rerender, the test failed.

src/Field.js
```diff
@@ -30,6 +31,11 @@ const createField = ({ deepEqual, getIn, setIn }) => {
         this.ConnectedField =
           createConnectedField(this.context._reduxForm, { deepEqual, getIn }, nextProps.name)
       }
+
+      const syncError = this.getSyncError()
+      if (syncError !== this.state.syncError) {
+        this.setState({ syncError })
+      }
```
Darmody: @erikras Is it reasonable?

erikras: No. The results of `getSyncError()` may change over time.

Darmody: I think it is a problem that can not track error properly. 
If field`s validation depend on itself, it is easy to track the error.
If field's validation depend on other field, there is no way to track the error unless make component  rerender every time.

Darmody: https://github.com/erikras/redux-form/pull/1120

this PR works because `Field` component always rerender.

src/Field.js
```diff
@@ -16,8 +16,11 @@ const createField = ({ deepEqual, getIn, setIn }) => {
       this.normalize = this.normalize.bind(this)
     }
 
-    shouldComponentUpdate(nextProps) {
-      return shallowCompare(this, nextProps)
+    shouldComponentUpdate(nextProps, nextState, nextContext) {
+      if (this.context._reduxForm.syncErrors !== nextContext._reduxForm.syncErrors) {
```
Darmody: I believe it works!

Darmody: Oops, it may not...

Darmody:
I apply the commit 4ad06b59041d17144e3e63e01de78cf0230f1267 to tag v6-alpha.15, all validation work fine in my project.

But when I apply it to branch v6, validation broken.

Since I don't know how to fix the problem, closed the PR.

erikras:
I have concluded that you are correct about needing to pass along `nextState`. But that _does_ cause the test that you commented out to fail, although I am not sure why.

Darmody:
@erikras Because  error states are stored in context, whenever error changed, the Field component will not rerender. That makes the test failed.

But I don't know how to fix it..

PR #1154, opened 2016-06-13
Prioritizing custom dispatch over global dispatch

tomkis1:
As [we discussed on twitter](https://twitter.com/tomkisw/status/742364129915240452) I believe that `redux-form` should prioritize `dispatch` which has been passed via props over global `connect`ed dispatch. The implementation is rather trivial and involves simple unit test.

Please, do you feel like this might be an issue? I am open to any improvements / suggestions.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -55,7 +55,6 @@ const createConnectedField = ({
     _reduxForm: PropTypes.object
   }
 
-  const actions = mapValues({ blur, change, focus }, actionCreator => actionCreator.bind(null, name))
   const connector = connect(
     (state, ownProps) => {
       const initial = getIn(getFormState(state), `initial.${name}`) || propInitialValue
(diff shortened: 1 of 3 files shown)


src/ConnectedField.js
```diff
@@ -72,7 +71,10 @@ const createConnectedField = ({
         _value: ownProps.value // save value passed in (for checkboxes)
       }
     },
-    actions,
+    (dispatch, initialProps) => mapValues({ blur, change, focus }, actionCreator => (...args) => {
```
ooflorent: This is really suboptimal. Action creators will be rebound on every render. Maybe we could assert that `dispatch` would never change and ACs once:

```js
(initialDispatch, initialProps) => {
  const dispatch = initialProps.dispatch || initialDispatch
  const boundActions = boundActionCreators(actions, dispatch)

  return () => boundActions
}
```

tomkis1: Well that's a tradeoff, I could imagine some use-cases where user would be forced to change the dispatch over life time of the component. However, what we could rather do, was some memoization.

PR #1178, opened 2016-06-17
Add newer v6 alpha links

mikepugh:


ooflorent:
Thanks! Looks good to me.

ooflorent:
Thanks!

Issue #1140, opened 2016-06-12
[v6] Initializing the form unregisters all the fields


eloytoro:
Dispatching `initialize` unregister all of the form fields, causing errors upon submission.
Also the third parameter `fields` (from the most recent docs) doesn't seem to apply to v6 since redux-form no longer needs to define the form fields in the form config.

Steps to reproduce:

1. Load a form
2. Make sure all fields were registered to the form
3. Dispatch `initialize`
4. Notice that the form state no longer no longer has `registeredFields`
5. Submitting the form causes an error

eliseumds:
The methods `reset` and `destroy` are doing the same. My use case is: I have a "Reset" button that resets the form and then automatically submits it. Because `registeredFields` disappear, we get:

````
Uncaught TypeError: Cannot read property 'map' of undefined
```

[at this line](https://github.com/erikras/redux-form/blob/v6/src/reduxForm.js#L181).

eloytoro:
This appears to be fixed in recent versions, closing it

PR #1179, opened 2016-06-17
[Docs] Remove not relevant to v6 params and description from 'initialize' action creator description

nktssh:
As mentioned in #1140 
> third parameter `fields` doesn't seem to apply to v6 since redux-form no longer needs to define the form fields in the form config.


erikras:
Thanks, @nktssh.

PR #1184, opened 2016-06-20
'checked' attribute should be set when string boolean like "true" is used

kromit:
I am using a key-value map as `intialValues`. The map is coming from server but the server has no knowledge about values so it sends all values as string. 

String `"true"` is a valid value for the `checked` attribute in react so it should also be for redux-form.

### DIFF
--- a/src/readField.js
+++ b/src/readField.js
@@ -115,7 +115,7 @@ const readField = (state, fieldName, pathToHere = '', fields, syncErrors, asyncV
     const initialFormValue = read(`${name}.initial`, form);
     const initialValue = initialFormValue || read(name, initialValues);
     field.name = name;
-    field.checked = initialValue === true || undefined;
+    field.checked = String(initialValue).toLowerCase() === 'true' || undefined;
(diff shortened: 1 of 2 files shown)


erikras:
Sure, seems reasonable.

src/readField.js
```diff
@@ -115,7 +115,7 @@ const readField = (state, fieldName, pathToHere = '', fields, syncErrors, asyncV
     const initialFormValue = read(`${name}.initial`, form);
     const initialValue = initialFormValue || read(name, initialValues);
     field.name = name;
-    field.checked = initialValue === true || undefined;
+    field.checked = String(initialValue).toLowerCase() === 'true' || undefined;
```
ooflorent: I don't like this condition... A more descriptive test would have been better in my opinion, something like:

```js
initialValue === true || initialValue === "true" || undefined
```

erikras:
Yeah, the test has failed, too. I'm repairing.

erikras:

PR #1187, opened 2016-06-20
Fixed problems caused by #1184

erikras:
Reworked how 'checked' is determined to allow string values. Originally started in #1184.

It's a shame that `Boolean('false') === true`. Silly javascript!

This change makes any case of `'true'` or `'false'` convert properly to boolean for checkboxes.

@ooflorent Review?


ooflorent:
lgtm

erikras:

Issue #1201, opened 2016-06-22
V6: Error in code after an async validation results in swallowed Errors (maybe bluebird specific)


q42jaap:
https://github.com/erikras/redux-form/blob/v6/src/reduxForm.js#L226

Let's say I make a typo in my onSubmit() function, causing a TypeError or something similar.
The resulting error causes a rejected promise (this might depend on promise library, bluebird wraps all continuations with tryCatcher).
The listenToSubmit function handles the submit promise (which is a result of having async validation).

The second argument to `.then()` "handles" the error.

I would like to suggest to have a different approach (or something similar which allows binding to `this`):

```
return promise.then(this.submitCompleted, (err) => {
  this.submitCompleted();
  return Promise.reject(err);
})
```

This way the Promise library taking the `then` can handle the rejection as needed.

This works well with bluebird, and I expect it to also work with native promises.

erikras:
That seems reasonable.

erikras:
Addressed in #1203. Will be in next release.

PR #1203, opened 2016-06-22
Always return rejected promise when submit is rejected

erikras:
Addresses #1201.

Also removed `returnRejectedSubmitPromise`. That was put in as a stopgap measure to avoid a breaking change, but I think the promise that is returned from submitting should always be rejected if there was an error.

@ooflorent, please review.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -223,7 +223,10 @@ const createReduxForm =
               return promise
             }
             this.submitPromise = promise
-            return promise.then(this.submitCompleted, this.submitCompleted)
+            return promise.then(this.submitCompleted, err => {
+              this.submitCompleted()
+              return Promise.reject(err)
+            })
           }
 
           submit(submitOrEvent) {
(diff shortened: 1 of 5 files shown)


ooflorent:
LGTM. I never liked `returnRejectedSubmitPromise` 👍

erikras:

src/reduxForm.js
```diff
@@ -223,7 +223,10 @@ const createReduxForm =
               return promise
             }
             this.submitPromise = promise
-            return promise.then(this.submitCompleted, this.submitCompleted)
+            return promise.then(this.submitCompleted, err => {
```
HriBB: As reported in #1266, this line throws `undefined:1 Uncaught (in promise) Object { ...errors }` if there are validation errors.

If I do this
```
return promise.then(this.submitCompleted, err => {
  this.submitCompleted()
  return Promise.reject(err)
}).catch(err => {
  console.log('==> CATCH', err)
})
```
the error is caught. Let's continue discussion in #1266

PR #1217, opened 2016-06-25
Change single quotes in test scripts to escaped double quotes

rluiten:
Allows tests to run on windows platforms.
Switched script "test:watch" to append options to the "test" script target.


ooflorent:
Thanks!

erikras:
Quotes in the npm scripts can be very finicky. Several times a helpful contributor has added or removed the quotes to make it work on Windows and it made half of the tests not run. I will test with this change.

erikras:
Seems okay. 👍

PR #1233, opened 2016-06-28
[v6] Add action setSubmitSuccess

eliseumds:
I got the following use case:

* User fills some forms fields
* A Redux saga catches the CHANGE actions and store the form state in local storage
* If user reloads the page without submitting the form, a saga rehydrates the form state after INITIALIZE
* When form is submitted or reset, local storage is cleared

I am not sure about the nomenclature ("setSubmitSuccess") though. Opinions?


ooflorent:
I'm not very fond of having `stopSubmit()` and `setSubmitSuccess()` called from `handleSubmit`. Maybe the submit lifecycle should be re-thought...

eliseumds:
@ooflorent could `stopSubmit()` be called on sync submits at least? One could then check:

```js
const action = yield take(STOP_SUBMIT);

if (!action.error) {
  // success!
}
```

eliseumds:
@ooflorent @erikras any feedback? I saw `onSubmitSuccess` was introduced but that still wouldn't solve my problem as I can't set it dynamically. This is how I'm expecting to tell a form component that its state will be persisted:

```
@reduxForm(...)
@persistFormState() // should work for both sync and async submits, that's the issue
class MyForm extends Component { ... }
```

erikras:
Sorry to take so long to get back to you, @eliseumds.

I think #1428 did some of this for us. There is now a `SET_SUBMIT_SUCCEEDED` action. Your PR was very helpful in reminding me all the ways that the contribution from #1428 needed to be used. Thank you.

eliseumds:
@erikras aah, awesome! Thanks guys (@scboffspring included). `SET_SUBMIT_SUCCEEDED` should suffice.

oreqizer:
Hi, came here, albeit this issue's closed, because `SET_SUBMIT_SUCCEEDED` is somewhat 'happening' without me knowing what activated it.

I have this [saga](https://github.com/oreqizer/reactizer/blob/develop/src/universal/modules/user/userSagas.js) where I manually call start/stop submit with the expectation of `props.submitting` to be `true` (my loading state depends on it), but `SET_SUBMIT_SUCCEEDED` sets it to `false` immediately after `startSubmit` is called.

Am I doing something wrong?

PR #1244, opened 2016-06-30
[WIP]Only pass initial values to validate when initialValues present

yesmeck:
~~Should fixes #1241~~

Not ready, do not merge.

PR #1061, opened 2016-05-27
Pass initial values to validate function on first render

yesmeck:

PR #1250, opened 2016-06-30
Revert "Pass initial values to validate function on first render"

erikras:
Reverts erikras/redux-form#1061. As requested on #1244.


erikras:
Published as [`v5.3.1`](https://github.com/erikras/redux-form/releases/tag/v5.3.1).

PR #1254, opened 2016-07-01
When there are no registered fields, return an empty list rather than…

### LINKED ISSUE #1231: [V6] Submitting with no registered fields
I have the multistep wizard form which last step has no field, just a plain submit button. 

Let's consider the follow example:

```js
function WizardFirstStep({ handleSubmit }) {
  return (
    <form onSubmit={handleSubmit}>
      <Field name="username" component="input" type="text" />
      <Field name="password" component="input" type="password" />
    </form>
  );
}

export default reduxForm({
  form: 'wizard',
  destroyOnUnmount: false,
})(WizardFirstStep);
```

```js
function WizardSecondStep({ handleSubmit }) {
  return (
    <form onSubmit={handleSubmit}>
      <button type="submit">Submit</button>
    </form>
  );
}

export default reduxForm({
  form: 'wizard',
  destroyOnUnmount: false,
})(WizardSecondStep);
```

And also the main component which changes these steps (like in the examples).

The issues is the following: in the reducer of `UNREGISTER_FIELD` action if the there is no fields `registeredFields` property is [removed](https://github.com/erikras/redux-form/blob/v6/src/reducer.js#L239) and when you try to submit with the removed property [`ReduxForm:getFieldList()`](https://github.com/erikras/redux-form/blob/v6/src/reducer.js#L239)  method will throw an error. 

I think that you shouldn't remove the `registeredFields` property from the state and let it to be an empty array.

One workaround for this is to create a hidden field `<Field name="hidden" component="input" type="hidden" />` just to make that property not to be empty.

clayne11:
… trying to map over nothing

Fixes #1231

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -1566,6 +1566,33 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
 
       expect(decorated.wrappedInstance.props).toEqual(wrapped.props)
     })
+
+    it('should return an empty list if there are no registered fields', () => {
+      const store = makeStore({})
+
+      class Form extends Component {
+        render() {
+          return (
+            <form>
+            </form>
+          )
+        }
+      }
+
+      const Decorated = reduxForm({
+        form: 'testForm'
+      })(Form)
+
+      const dom = TestUtils.renderIntoDocument(
+        <Provider store={store}>
+          <Decorated/>
+        </Provider>
+      )
+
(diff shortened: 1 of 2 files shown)


ooflorent:
I think this change should be made into the store to ensure that `registeredFields` is always an array.
Iterating over it would have no cost.

erikras:
I fixed this already. I'll add the test, though.

ruiaraujo:
A release with that fix would be awesome, I am seeing some crashes in production related to registeredFields being undefined.

erikras:

Issue #726, opened 2016-03-10
redux-form v5 API Proposal


erikras:
Greetings all,

There are a number of feature requests, bugs, and performance issues that have caused me to rethink the design of `redux-form`. I'd like to use this issue as a way to introduce you to some of my thinking, prepare you for what's to come, and get your feedback. Hopefully at least one of the features will make you excited to migrate to `v5` when it comes out.

# `redux-form` `v5` API Proposal

#### Glossary
> __connected__: subscribed to changes in the Redux store with [`connect()`](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options)

## 1. The form should not be connected

When writing the initial API, I provided things that seemed convenient for small forms, like having `this.props.values` be the current state of the values of your form. This required having the wrapping form component to be connected to the Redux store, ___and to re-render on every value change___. This doesn't matter if it's just a `username` and `password` form, but some of you are writing complex CRM systems with potentially hundreds of fields on the page at a time, causing `redux-form` to grind to a halt. #529 

So, what does this mean?

* The convenience functionality of providing `mapStateToProps` and `mapDispatchToProps` is going away. You'll have to use `connect()` directly yourself.
* But if the form isn't connected, then...

## 2. Each field is connected

Each field should subscribe _only_ to the very specific slice of the Redux state that it needs to know how to render itself. This means that `redux-form` can no longer only provide you with props to give to your own inputs.

```js
import React, { Component } from 'react'
import { reduxForm } from 'redux-form'

@reduxForm({
  form: 'login',
  fields: [ 'username', 'password' ]
})
export default class LoginForm extends Component {
  render() {
    const { fields: { username, password } } = this.props
    return (
      <form>
        <input type="text" {...username}/>     // <--- can no longer use this syntax
        <input type="password" {...password}/> // <--- can no longer use this syntax
      </form>
    )
  }
}
```

So what do we do? The destructured props was one of the most elegant things about `redux-form`!

There are __three options__ that I want to support:

### 2.1. On-Demand Field Structure

The `redux-form` user base seems pretty divided on how much they like providing the list of fields for the entire form, so I want to provide away to arbitrarily specify the path to your field at render-time. This can be done with a `<Field/>` component that is given the name of the field in `dot.and[1].bracket` syntax.

```js
import React, { Component } from 'react'
import { reduxForm, Field } from 'redux-form'

@reduxForm({
[...]

PR #1256, opened 2016-07-01
[v6] Fix broken links and typos in docs

fredrikolovsson:
Came across a few things when reading docs. This pull request fixes:

1. Broken links between validation examples, e.g. try to click the "Submit Validation" link in this example on [Github](https://github.com/erikras/redux-form/blob/v6/examples/asyncValidation/src/AsyncValidation.md) and [redux-form.com](http://redux-form.com/6.0.0-alpha.15/examples/asyncValidation/).
2. A couple of typos

Also noticed two more potential changes in the [`reduxForm()`](http://redux-form.com/6.0.0-alpha.15/docs/api/ReduxForm.md/) docs, which I could add to the PR:

screenshot 2016-07-01 16 10 53 copy

* **Blue**: This link seems superfluous and the placement would suggest it relates to all options or options in general, which is not the case.
* **Red**: As mentioned in #726, section 12, the ambition is to have docs navigable on both Github and redux-form.com. In the docs for `reduxForm()`, the link to the async validation example works on Github, but not on redux-form.com. Not sure how to resolve that. A quick fix could be to provide two links (sounds ugly, but perhaps better than something broken), if there is no easy way to restructure the files to allow relative links.

Please let me know if there is something you would like me to add or edit. Thanks.

### DIFF
--- a/docs/api/ReduxForm.md
+++ b/docs/api/ReduxForm.md
@@ -111,7 +111,7 @@ Your `shouldAsyncValidate()` function will be given an object with the following
 
 > The reason to possibly run async validation. It will either be: `'blur'` or `'submit'`, 
 depending on whether an async blur field had triggered the async validation or if submitting the 
-form has triggered it, respecitively.
+form has triggered it, respectively.
 
 > ##### `blurredField : string` [optional]
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -70,7 +70,7 @@ open [`localhost:3030`](http://localhost:3030) in your browser.
 
 ---
 
-### [React-Widegets Examples](react-widgets)
+### [React-Widgets Examples](react-widgets)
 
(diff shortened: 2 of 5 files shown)


erikras:
Thanks for this, @ferdinandsalis!

I'll remove the thing you put in the blue box.

As for the red... Grrr!! So frustrating. I designed the redux-form.com website with the express intention of being able to navigate links the same on GH and on the website, but the darn gh-pages web server is forcing a `/` on the end of my `/ReduxForm.md` url. I'll have to think on it...

Thanks again for your help.

erikras:
Ah, I know. Those links should just be hardcoded to go to redux-form.com. That's where you can actually _see_ the example running, rather than in the GH repo. Should be fixed now on rc1 version of website.

fredrikolovsson:
Sounds good 😃 Cheers!

PR #1258, opened 2016-07-01
Added propNamespace functionality

erikras:
Rarely (?) used functionality from `v5`.

Also specifically whitelisted the props that `redux-form` provides to the wrapped component.


erikras:

Issue #1241, opened 2016-06-29
Validate returning undefined as values


mmahalwy:
Version `5.3.0`

Validate is returning `undefined`, see: 

```js
const validate = (...args) => {
  const errors = {};
  console.log(args);
  // if (!values) return errors;
  //
  // if (!values.email) {
  //   errors.email = 'Required';
  // } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values.email)) {
  //   errors.email = 'Invalid email address';
  // }
  //
  // if (!values.password) {
  //   errors.password = 'Required';
  // }
  //
  // if (fields.includes('passwordConfirmation') && !values.passwordConfirmation) {
  //   errors.passwordConfirmation = 'Required';
  // }
  //
  // if (
  //   fields.includes('passwordConfirmation') &&
  //   values.password &&
  //   values.passwordConfirmation &&
  //   values.password !== values.passwordConfirmation) {
  //   errors.passwordConfirmation = 'Does not match password';
  // }

  return errors;
};

@reduxForm({
  form: 'authentication',
  validate
})
export default class AuthenticationForm extends Component {
  static propTypes = {
    fields: PropTypes.object,
    handleSubmit: PropTypes.func,
    CTA: PropTypes.string.isRequired,
    authError: PropTypes.object,
    submitting: PropTypes.bool
  };

  render() {
    const { fields: { email, password, passwordConfirmation }, handleSubmit, CTA, authError, submitting } = this.props;
    const styles = require('./style.scss');

    return (
      <form className="login-form" onSubmit={(event) => event.preventDefault()}>
        <FormGroup>
          <FormControl {...email} type="email" placeholder="Email" />
          {
            email.touched && email.error &&
            <div className={`text-left ${styles.errorText} text-danger`}>{email.error}</div>
          }
        </FormGroup>
....
```

mmahalwy:
Seems like this has happened right after 5.3.0 was released...

vdh:
I had the same issue with the validate function getting passed `undefined` values with 5.3.0, downgrading to 5.2.5 fixed it for me.

- For me the issue happened on `redux-form/CHANGE` actions, but `redux-form/START_SUBMIT` sent the values through to the validate function like normal.
[...]

PR #1259, opened 2016-07-01
Create a initial form state before creating fields

yesmeck:
This should fix the long living bug #621.

First round
===========

My first try was #1025, I tried pass a default field state to [updateField](https://github.com/erikras/redux-form/blob/v5.3.1/src/updateField.js#L8) so that updateField can read initial value from given state.

But I forgot the the validate function which causes validate function cann't read initial values at first render.

Second round
============

My second try was #1061, I tried pass initial values to validate function. But I forgot a form may have no initial values or only have partial initial values which causes #1241.

Third round
=====================

I thought I can create a initial form state before creating fields, so that updateField and validate function can both read initial values.

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -2804,4 +2804,43 @@ describe('createReduxForm', () => {
     // FAILS
     //expect(lastPrevBarValue).toNotEqual(lastNextBarValue);
   });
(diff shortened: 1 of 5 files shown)


yesmeck:
@erikras Could you create a `v5` label for v5 issue?

erikras:

PR #1265, opened 2016-07-02
Fix file input

HriBB:
This fixes file input in v6

### DIFF
--- a/src/__tests__/createFieldProps.spec.js
+++ b/src/__tests__/createFieldProps.spec.js
@@ -268,6 +268,14 @@ const describeCreateFieldProps = (name, structure, expect) => {
         .toEqual([])
     })
 
+    it('should default value to undefined for file inputs', () => {
+      expect(createFieldProps(getIn, 'foo', {
+        state: empty,
+        type: 'file'
+      }).value)
+        .toEqual(undefined)
+    })
+
     it('should replace undefined value with empty string', () => {
       const result = createFieldProps(getIn, 'foo', {
         state: empty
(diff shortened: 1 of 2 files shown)


niksajanjic:
This doesn't fully fix issue with file input because checking for type happens only if you pass it through Field component. What if I want to statically set file type inside a component that I pass to Field component.

So, this works:

```
<Field name="image" component={FileUpload} type="file" />

const FileUpload = ({ input, type }) =>
  <input type={type} {...input} />
```

This doesn't work and throws errors:
```
<Field name="image" component={FileUpload} />

const FileUpload = ({ input, type }) =>
  <input type="file" {...input} />
```

Either there should also be a check for statical type or documentation should mention this constraint.

niksajanjic:
Actually, the thing is that I need value in my component passed to Field component because my FileUpload component also has preview of the image. That is why I'm not assigning file type on Field component but I'm passing value inside and then making sure I don't assign it to my input field. This way I can statically assign file type and I also have access to value so I don't have to fetch the value from store. Not sure if there's a better way of solving this issue as it is right now.

ahmeier:
in redux-form 7.3.0 the working file upload @niksajanjic suggested does NOT work anymore:
`<Field name="image" component={FileUpload} type="file" />`
`const FileUpload = ({ input, type }) => <input type={type} {...input} />`

results in the following error:

> Uncaught DOMException: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string.

PR #1268, opened 2016-07-02
Fix submit promise error if form is not valid

HriBB:
@erikras not 100% sure about this one, but I believe that returning rejected promise in `listenToSubmit()` is unnecessary and only results in an error. This PR fixes that.


HriBB:
I think this needs some more thought :)

Issue #1249, opened 2016-06-30
[React 15.2.0] Additional props warning


erikras:
It will be interesting to see, when React `15.2.0` is released, how many of these warnings `redux-form` generates. More info and discussion [here](https://github.com/react-bootstrap/react-bootstrap/issues/1994) from [the page the warning links to](https://gist.github.com/jimfb/d99e0678e9da715ccf6454961ef04d1b):

---

> The unknown-prop warning will fire if you attempt to render a DOM element with a prop that is not recognized by React as a legal DOM attribute/property.  You should ensure that your DOM elements do not have spurious props floating around.

> There are a couple of likely reasons this warning could be appearing:

> 1.  Your component is transferring its own props directly to a child element (eg. https://facebook.github.io/react/docs/transferring-props.html).  When transferring props to a child component, you should ensure that you are not accidentally forwarding props that were intended to be interpreted by the parent component.

> 2.  You are using a non-standard DOM attribute, perhaps to represent custom data.  Custom data attributes should always have a `data-` prefix, as described in https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes

> 3.  React does not yet recognize the attribute you specified.  This will likely be fixed in a future version of React.  However, React currently strips all unknown attributes, so specifying them in your React app will not cause them to be rendered.

---

Props like `valid`, `invalid`, `dirty`, and `pristine` might cause warnings for cases where someone is just destructuring the field object into an input, i.e. `<input type="text" {...firstName}/>`.

ooflorent:
I'm affraid this change will lead to a massive amount of issues created here... Unfortunately we can't do much to prevent this. One solution in v6 is to provide a new input component.

```js
const InputProps = [
  // HTML attributes
  "placeholder",
  "type",
  "value",
  // Event listeners
  "onBlur",
  "onChange",
  "onFocus",
]

const InputWrapper = props => 
  React.createElement("input", _.pick(props, InputProps))

export default InputWrapper
```

And to make it the default component of `Field`:

```js
Field.defaultProps = {
  component: InputWrapper,
}
```

ooflorent:
Another workaround is to instrument `component` if it is passed as a `String` (`input` or `select`).

erikras:
Results of first test:

---

[...]

PR #1276, opened 2016-07-04
[React 15.2.0] Moved input props to input structure

erikras:
As discussed on #1249.

This will be a Breaking Change, but there's nothing we can really do about it.


jayphelps:
This PR removed `onUpdate` and didn't include it in the breaking changes BTW

erikras:
Yes, thank you. Added to the release notes.

PR #1277, opened 2016-07-04
Add Action Creators to the primary contents in the getting started guide

FuzzySockets:
This would have been very helpful for me to know up front.  I spent like 6 hours trying to figure out how to 'bind' redux data to the form because I didn't understand that you could trigger form behavior from external components.

In my case, I have a modal window for Terms and Conditions which can either check or uncheck a checkbox.


erikras:
👍 Thanks!

Issue #1284, opened 2016-07-05
[v6 rc1] Revalidation does not clear errors


estaub:
Validation errors are not cleared when the user corrects them.

I'm new with redux-form, so I thought this was my badness, but after seeing the same behavior on http://erikras.github.io/redux-form-material-ui/, I'm not so sure.  

At http://erikras.github.io/redux-form-material-ui/, 

1. click in "Name" field
2. click in "Email" field.  The "Name" field now goes red.
3. click back in "Name" field and enter text.  The "Name" field stays red.
 
Nothing you do (AFAICT) will clear the error message.

In my case, I see that the validation function is returning a correct "errors" object, but the "error" on fields that are absent in that "errors" object are not cleared on the field itself.  So it looks like a redux-form issue.

erikras:
Thanks. This is a known bug I'm working on.

erikras:
This was introduced in 6rc1. It does not exist in 6a15.

erikras:
Should be fixed in [`6.0.0-rc.3`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.3).

brotzky:
@erikras 
Just downloaded 6.0.0-rc.3 and it is indeed fixed! Thank you **so much** for the awesome support. Having the errors within redux state is awesome.

PR #1292, opened 2016-07-06
Moved sync errors into Redux state

### LINKED ISSUE #1282: [v6] Clearing errors?
I've scoured the documentation hoping to find a way to easily clear form errors.  Is there an action creator for this? I'm looking to allow the user to "x" out of an error message at the field or form level.

erikras:
I have resisted this for a very long time, all the way back to Issue #45, but putting the sync errors into the Redux state solves so many problems that I was taking convoluted measures to fix before.

The outer form HOC is in charge of noticing when values change, running sync validation, and then dispatching `UPDATE_SYNC_ERRORS` if and only if the sync errors have changed.

Fixes #1282 #1284


erikras:
You win again, @davidkpiano! 😡  😄

davidkpiano:
Haha, it really does make things so much easier though, I promise!

clayne11:
Definitely excited about this!

chrisbateman:
@erikras One possible downside to this approach is that you can't really use React elements for error messages. For instance:

```javascript
   errors.firstName = <span>A message with some <b>bold text</b></span>;
```

Putting React elements in a redux store can do some pretty bad things for redux performance. If it were just simple HTML - you could use dangerouslySetInnerHTML to render error messages - but that doesn't allow for the use of other reusable React components in the error messages. Any possible alternatives for this case?

tyrsius:
@erikras I am confused, does this require changes to our code, or is the outer HOC behavior you mentioned taken care of by ReduxForm?

clayne11:
It's handled by `redux-form`. You don't have to change anything in your code.

s-coimbra21:
@chrisbateman Create custom input components and use the component prop of Field

like so

```js
<div>
        <input className={cx(className, styles.inputInner)} {...otherProps} />
        {(dirty || touched) && error && <span className={cx(errorClassName, styles.inputhelp)}>{error}</span>}
        {children}
</div>
```

chrisbateman:
@s-coimbra21 In my case - I have some error messages with custom HTML/behavior - and some without.  The custom behavior is usually a popover link (something like "learn more" about how to fix that particular error).

Since I can't pass React components into the error messages anymore - I'm still stuck without a solution.

clayne11:
You can use an object that describes the error. For example

```javasciprt
{
  "type": "popoverError",
  "message": "Something went wrong",
  "popover": "Click here to learn more",
}
```

And then you can render a different component based on the type.

chrisbateman:
@clayne11 Yes, but there are still some drawbacks there:
- the provider of the error message can't control where the popover would fit into the error message
- the reusable input component is now responsible for owning/importing/providing the Popover - rather than allowing the application (which is doing the validating) to provide it (only if it wishes to).
- the provider of the error message still can't provide any other custom HTML (this could still be done if the input component rendered the error with dangerouslySetInnerHTML - but that doesn't solve the above problems)

Basically - just losing a bunch of flexibility with something that was previously very simple to do.

clayne11:
IMO the overall gains in flexibility of being able to structure your UI in any way you want and call the submit action from outside of the form context FAR outweigh the loss in how you can display error messages. There are workarounds for displaying error messages but there aren't really for submitting with a `redux` action. Additionally, having the sync errors in the store fixed several other issues where errors sometimes would not re-render.

PR #1318, opened 2016-07-08
[WIP] Typescript definition for v6

CarsonF:
# Typescript definition for v6 RC 4
This allows users to have the definitions just from installing the repo from `npm`, which is 👍 👍 💯 .

@erikras I don't mind helping maintain this, if you would just ping me before release I can do the definition updates. Or you can do them yourself and I can look over them :) It should be pretty easy to do small tweaks.

This is very close to complete. I just need to test:
- [ ] FieldArrays
- [ ] Reducer plugin
- [x] `validate`'s 2nd parameter (may have been a bug that was fixed)

@blakeembrey Would love your thoughts and verification that I'm doing the react dependency right.

EDIT: Updated for RC4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.iml
 coverage
 node_modules
+typings
 dist
 lib
 es
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,6 +9,7 @@ import {
   SyntheticEvent
 } from 'react';
 import { Dispatch, Reducer, Action } from 'redux';
+import FormEventHandler = __React.FormEventHandler;
 
 type FieldValue = any;
 
@@ -569,17 +570,14 @@ interface WrappedFieldInputProps {
   name: string;
 
   /**
-   * A function to call when the form field loses focus. It expects to
-   * either receive the React SyntheticEvent or the current value of the
-   * field.
+   * A function to call when the form field loses focus.
    */
-  onBlur(eventOrValue: SyntheticEvent | FieldValue): void;
+  onBlur: FocusEventHandler;
 
   /**
-   * A function to call when the form field is changed. It expects to either
-   * receive the React SyntheticEvent or the new value of the field.
+   * A function to call when the form field is changed.
    */
-  onChange(eventOrValue: SyntheticEvent | FieldValue): void;
+  onChange: FormEventHandler;
 
   /**
    * A function to call when the form field receives a 'dragStart' event.
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,954 +1,949 @@
 import { Component, ComponentClass, StatelessComponent, ReactElement, ReactEventHandler, SyntheticEvent } from 'react';
 import { Dispatch, Reducer, Action } from 'redux';
 
-declare namespace ReduxForm {
+type FieldValue = any;
 
-  type FieldValue = any;
+type DataShape = {[fieldName:string]: FieldValue};
 
-  type DataShape = {[fieldName:string]: FieldValue};
+type FormErrors<FormData extends DataShape> = FormData & { _error?: string };
 
-  type FormErrors<FormData extends DataShape> = FormData & { _error?: string };
+/**
+ * Decorator / Form Config
+ */
 
-  /**
-   * Decorator / Form Config
-   */
+export function reduxForm<FormData extends DataShape, P, S>(
+  config: Config<FormData, P, S>
+): FormDecorator<FormData, P, S>;
 
-  export function reduxForm<FormData extends DataShape, P, S>(
-    config: Config<FormData, P, S>
-  ): FormDecorator<FormData, P, S>;
-
-  interface FormDecorator<FormData extends DataShape, P, S> {
-    <T extends (typeof Component)>(component: T): T & Form<FormData, P, S>;
-  }
-
-  interface Config<FormData extends DataShape, P, S> {
-    /**
-     * the name of your form and the key to where your form's state will be
-     * mounted under the redux-form reducer
-     */
-    form: string;
-
-    /**
-     * An adapter function that will render a component based on a string component
-     * type and the props given to a Field. Remember that all you really need to hook
-     * up to your custom component's value and onChange. Defaults to [].
-     *
-     * See Asynchronous Blur Validation Example for more details.
-     */
-    adapter?: (component: string, props: Object) => ReactElement<any>;
-
-    /**
-     * field names for which onBlur should trigger a call to the asyncValidate
-     * function. Defaults to [].
-     *
-     * See Asynchronous Blur Validation Example for more details.
-     */
-    asyncBlurFields?: string[];
-
-    /**
-     * a function that takes all the form values, the dispatch function, and
-     * the props given to your component, and returns a Promise that will
-     * resolve if the validation is passed, or will reject with an object of
-     * validation errors in the form { field1: <String>, field2: <String> }.
-     *
-     * See Asynchronous Blur Validation Example for more details.
-     */
-    asyncValidate?(values: FormData, dispatch: Dispatch<S>, props: P): Promise<any>;
-
-    /**
-     * Whether or not to automatically destroy your form's state in the Redux
-     * store when your component is unmounted. Defaults to true.
-     */
-    destroyOnUnmount?: boolean;
-
-    /**
-     * A function that takes the entire Redux state and returns the state slice
-     * which corresponds to where the redux-form reducer was mounted. This
-     * functionality is rarely needed, and defaults to assuming that the reducer
-     * is mounted under the "form" key.
-     */
-    getFormState?(state: S): any;
-
-    /**
-     * The values with which to initialize your form in componentWillMount().
-     * Particularly useful when Editing Multiple Records, but can also be used
-     * with single-record forms. The values should be in the form
-     * { field1: 'value1', field2: 'value2' }.
-     */
-    initialValues?: FormData;
-
-    /**
-     * Reinitialize the form every time the initialValues prop changes.
-     * Defaults to false.
-     */
-    enableReinitialize?: boolean;
-
-    /**
-     * The function to call with the form data when the handleSubmit() is fired
-     * from within the form component. If you do not specify it as a prop here,
-     * you must pass it as a parameter to handleSubmit() inside your form
-     * component.
-     */
-    onSubmit?: SubmitHandler<FormData, S>;
-
-    /**
-     * A callback function that will be called when a submission fails for whatever reason.
-     */
-    onSubmitFail?: (errors: FormErrors<FormData>, dispatch: Dispatch<S>) => void;
-
-    /**
-     * A callback function that will be called when a submission succeeds.
-     */
-    onSubmitSuccess?: (result: any, dispatch: Dispatch<S>) => void;
-
-    /**
-     * If specified, all the props normally passed into your decorated
-     * component directly will be passed under the key specified. Useful if
-     * using other decorator libraries on the same component to avoid prop
-     * namespace collisions.
-     */
-    propNamespace?: string;
-
-    /**
-     * The use of this property is highly discouraged, but if you absolutely
-     * need to mount your redux-form reducer at somewhere other than form in
-     * your Redux state, you will need to specify the key you mounted it under
-     * with this property. Defaults to 'form'.
-     *
-     * See Alternate Mount Point Example for more details.
-     */
-    reduxMountPoint?: string;
-
-    /**
-     * An optional function you may provide to have full control over when async
-     * validation happens.
-     */
-    shouldAsyncValidate?(params: AsyncValidateCallback<FormData>): boolean;
-
-    /**
-     * marks fields as touched when the blur action is fired. Defaults to true.
-     */
-    touchOnBlur?: boolean;
-
-    /**
-     * marks fields as touched when the change action is fired. Defaults to
-     * false.
-     */
-    touchOnChange?: boolean;
-
-    /**
-     * a synchronous validation function that takes the form values and props
-     * passed into your component. If validation passes, it should return {}.
-     * If validation fails, it should return the validation errors in the form
-     * { field1: <String>, field2: <String> }.
-     * Defaults to (values, props) => ({}).
-     */
-    validate?(values: FormData): FormErrors<FormData>;
-  }
-
-  interface SubmitHandler<FormData extends DataShape, S> {
-    // Promise should be rejected with FormErrors<FormData> but the definition has that as "any" -__-
-    (values: FormData, dispatch?: Dispatch<S>): void | FormErrors<FormData> | Promise<any>;
-  }
-
-  interface AsyncValidateCallback<FormData extends DataShape> {
-    /**
-     * Any existing asynchronous validation errors
-     */
-    asyncErrors?: FormErrors<FormData>;
-
-    /**
-     * true if the form has ever been initialized with initial values
-     */
-    initialized: boolean;
-
-    /**
-     * The reason to possibly run async validation. It will either be: 'blur' or
-     * 'submit', depending on whether an async blur field had triggered the async
-     * validation or if submitting the form has triggered it, respectively.
-     */
-    trigger: 'blur' | 'submit';
-
-    /**
-     * The name of the field that has triggered the async validation. May be undefined.
-     */
-    blurredField?: string;
-
-    /**
-     * true if the form is pristine, false if it is dirty
-     */
-    pristine: boolean;
-
-    /**
-     * true if synchronous validation is passing, false if it is failing.
-     */
-    syncValidationPasses: boolean;
-  }
-
-  /**
-   * Form
-   */
-
-  /**
-   * The following are methods or properties that you can access on an instance
-   * of your decorated form component (outermost component named "ReduxForm").
-   */
-  interface Form<FormData extends DataShape, P, S> extends Component<P, any> {
-    /**
-     * true if the form data has changed from its initialized values. Opposite
-     * of pristine.
-     */
-    dirty: boolean;
-
-    /**
-     * An array of strings representing all the fields in the form.
-     * Mainly useful for testing.
-     */
-    fieldList: string[];
-
-    /**
-     * true if the form has validation errors. Opposite of valid.
-     */
-    invalid: boolean;
-
-    /**
-     * true if the form data is the same as its initialized values. Opposite
-     * of dirty.
-     */
-    pristine: boolean;
-
-    /**
-     * Resets all the values in the form to the initialized state, making it
-     * pristine again.
-     */
-    reset(): void;
-
-    /**
-     * Submits the form. Returns a promise that will be resolved when the form
-     * is submitted successfully, or rejected if the submission fails.
-     */
-    submit(): Promise<any>;
-
-    /**
-     * true if the form passes validation (has no validation errors). Opposite
-     * of invalid.
-     */
-    valid: boolean;
-
-    /**
-     * The current values of all the fields in the form.
-     */
-    values: FormData;
-
-    /**
-     * A reference to the instance of the component you decorated with reduxForm().
-     * Mainly useful for testing.
-     */
-    wrappedInstance: ReactElement<P & FormProps<FormData, S>>
-  }
-
-  /**
-   * These are the props that will be passed to your form component.
-   * Your form component's props can extend this interface.
-   */
-  interface FormProps<FormData extends DataShape, S> {
-    /**
-     * true if any of the fields have been marked as touched, false otherwise.
-     */
-    anyTouched?: boolean;
-
-    /**
-     * A function that may be called to initiate asynchronous validation if
-     * asynchronous validation is enabled.
-     */
-    asyncValidate?: () => void;
-
-    /**
-     * This value will be either:
-     * - false - No asynchronous validation is currently happening
-     * - true - Asynchronous validation is currently running in preparation to submit a form
-     * - a string - The name of the field that just blurred to trigger asynchronous validation
-     */
-    asyncValidating?: string | boolean;
-
-    /**
-     * Destroys the form state in the Redux store. By default, this will be
-     * called for you in componentWillUnmount().
-     */
-    destroy?(): void;
-
-    /**
-     * true if the form data has changed from its initialized values. Opposite
-     * of pristine.
-     */
-    dirty?: boolean;
-
-    /**
-     * A generic error for the entire form given by the _error key in the
-     * result from the synchronous validation function, the asynchronous
-     * validation, or the rejected promise from onSubmit.
-     */
-    error?: string;
-
-    /**
-     * The form name that you gave to the reduxForm() decorator or the prop you
-     * passed in to your decorated form component.
-     */
-    form?: string;
-
-    /**
-     * A function meant to be passed to <form onSubmit={handleSubmit}> or to
-     * <button onClick={handleSubmit}>. It will run validation, both sync and
-     * async, and, if the form is valid, it will call
-     * this.props.onSubmit(data) with the contents of the form data.
-     *
-     * Optionally, you may also pass your onSubmit function to handleSubmit
-     * which will take the place of the onSubmit prop. For example: <form
-     * onSubmit={handleSubmit(this.save.bind(this))}>
-     *
-     * If your onSubmit function returns a promise, the submitting property
-     * will be set to true until the promise has been resolved or rejected.
-     * If it is rejected with an object like
-     * new SubmissionError({ field1: 'error', field2: 'error' })
-     * then the submission errors will be added to each field (to the error
-     * prop) just like async validation errors are. If there is an error that
-     * is not specific to any field, but applicable to the entire form, you
-     * may pass that as if it were the error for a field called _error, and
-     * it will be given as the error prop.
-     */
-    handleSubmit?(event: SyntheticEvent): void; // same as ReactEventHandler
-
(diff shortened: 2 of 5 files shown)


erikras:
While personally a TS virgin, I heartily approve of this PR. Let me know when you feel it is complete.

calvin:
@CarsonF heads up! I'm getting the following error using your tsd.

```
ERROR in /home/user/typings/modules/redux-form/index.d.ts
(834,58): error TS7006: Parameter 'from' implicitly has an 'any' type.

ERROR in /home/user/typings/modules/redux-form/index.d.ts
(834,64): error TS7006: Parameter 'to' implicitly has an 'any' type.
```

bbenezech:
@CarsonF Using '?' for FormProps definitions is a hack that helps with compilation but it won't work with TS2 strictNullChecks.
For example, TS will complain that this.props.fields can be undefined (since the prop is optional in definition).

We should be using correct HOC definition (merge Props) like Redux' `connect`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/react-redux/react-redux.d.ts#L26-L73

It was added with this commit https://github.com/DefinitelyTyped/DefinitelyTyped/commit/206c964ed1f68ec1778c029123dada84f2aa25c1 by @use-script

It forwards Redux's props (like state selections or dispatch) in addition to the decorated component's props.

That way you don't need to make HOC's props optionals in order to pass compilation. 

Unfortunately I don't feel confident enough in my TS skills to provide any kind of help here.

@use-script would you consider helping us? That would be awesome 🎉

CarsonF:
@calvin Yeah those are from the `arrayMove` action creator I just added. @erikras That one isn't documented and I couldn't figure out from source what `from` and `to` parameters where supposed to be. Strings?

----

@bbenezech Ah good point. Ironically I have problems using `react-redux`'s `connect` function as a decorator. TS has declared that decorators cannot mutate the class definition (more info at [TypeScript#4881](https://github.com/Microsoft/TypeScript/issues/4881)). 
```ts
declare type ClassDecorator = <TFunction extends Function>(target: TFunction) => TFunction | void;
```
I've looked at that block of code several times too and it's hard to wrap my head around. I'll see what I can do to fix this though.
I've gotten around this before with `(target: T)  =>  T & DecoratedT`. Although that's still a hack too because it doesn't return `T` only `DecoratedT` but it satisfies the compiler.

erikras:
> That one isn't documented and I couldn't figure out from source what `from` and `to` parameters where supposed to be. Strings?

Integers. The indexes in the array.

bbenezech:
@CarsonF 
I had 2 issues with redux decorators:

1. `connect()(MyComponent)` doesn't require `dispatch; Dispatch<any>` prop and doesn't satisfy it. You need to add it to your props with a ?
I fixed the issue with `export function connect<TOwnProps>(): ComponentDecorator<{dispatch: Dispatch<any>}, TOwnProps>`
I then pass `TOwnProps` like `connect<TOwnProps>()(MyComponent)`

2. The other signature overloads won't work without explicitly type casting the connect function (like above) unless TS can infer TOwnProps with `mapStateToProps(state, props: TOwnProps)`.

For redux-form, we should probably require passing ownProps explicitly and keep ReduxFormProps optionals so that we don't need to redeclare them all in user land.
Unless we Inform user there Component's prop type should extend ReduxFormProp.
Which is better for autocomplete.

rluiten:
@CarsonF I have pulled in your type definition and adjusted my code to match names it mostly looks good.

I have temporarily hacked my BaseFieldProps to allow me to pass id, type and value at moment as I need to be able to pass in 'value' as a minimum this way at the moment due to what I believe may be a bug.

One larger issue a bug in Field maybe ?

1. Currently it appears it is not possible to pass in the 'value' property to Field via the props property.

This works fine for defining members of my radio buttons.
```html
<Field name={name} component="input" id={fieldTrueId} type="radio" value="true" />
```
This does not work at, it breaks because the 'value' field in props is not set on the creatd input element in the DOM. I think this one may be a bug in Field for handling the property.
```html
<Field name={name} component="input" props={{value: 'false', id: fieldFalseId, type: 'radio'}} />
```

If this is a bug

One smaller issue.

2. I think its nicer to have the 'id', 'type', 'value' on BaseFieldProps so they can be passed to Field rather than forced to use the 'props' property. There are probably a long list of other nice to have attributes for different user cases and different components passed to Field so maybe we need to make BaseFieldProps generic ? (not sure how far reaching that change would be)

CarsonF:
@rluiten I think you are having different problems. You shouldn't need to set the value on the Field. You should use the `defaultValue` prop for that field or `initialValues` on the form config.

rluiten:
@CarsonF it is a radio button, the value is the value that is set to the identifier name if the radio button is checked or inversely in redux-form. This value is looked up by redux-form and used with type and the current model to derive the checked property for the control. I don't see a way around it, at the moment the only extra field on baseFieldProps I need without modifying redux-form is 'value'.

rluiten:
@CarsonF I have created a branch in my fork of redux-form with a very simple change to examples/src/SimpleForm in redux-form repo.

I changed it so that `value` property is delivered via the `props` property and the radio button no longer operates correctly. This is why at least at the moment I think we need `value` on BaseFieldProps

It may be a bug in redux-form Field but I had a dig around but had no luck making progress there.
I am happy to raise it as an issue if you think thats the better avenue ?

Here is reference to the commit  that makes Field not work right for radio input buttons.
https://github.com/rluiten/redux-form/commit/cf115873ce3b14289c27ebf20edd72e7c690cd00

I would have liked to run up a codepen, jsfiddle or jsbin or something with a runnable example but it didn't appear feasible at the moment with no readily accessible version of rc3 of redux-form around.

CarsonF:
@bbenezech I've looked again and I don't think this is possible. Yes what you are proposing is more correct, but if it doesn't compile it doesn't matter. See here: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9951

CarsonF:
@erikras I'm trying to address @rluiten's situation. How does Field's `defaultValue` differ from component's `value`?

erikras:
`defaultValue` is the answer to the question: "What should I use as a value if `value` is `undefined`?"

It defaults to `''`, but if your component wanted a `Number`, it could be `0`, or for a `Date`, it could be `new Date()` (today).

So as far as typing is concerned, it can be anything.

CarsonF:
Should've reread the doc :facepalm: Did you intend to have `value` passed into `Field` directly? Seems like it is bypassing redux store.. Are radio buttons an edge case maybe?

CarsonF:
@rluiten You could do this:
```ts
import * as React from 'react';
import { Field } from 'redux-form';

// This is how they say to define react components that have generics
// ...I don't really like it either.
type InputField = new () => Field<React.HTMLAttributes>;
const InputField = Field as InputField;

...
render() {
  return <InputField component='input' type='hidden' />;
}
```
That would work for `input`, `select`, and `textarea`.

erikras:
> Did you intend to have value passed into Field directly? Seems like it is bypassing redux store.. Are radio buttons an edge case maybe?

Bingo. It's only used for radio buttons.

rluiten:
Thanks for having a look at this case, I really don't want a custom type definition when 6 is released.

It appears that InputField as it stands now does not compile for me and I don't understand the error at the spread attribute to fix it at the moment. I would have expected the casting of Field as the new InputField type was the point of the exercise to address types for typescript.

    <InputField component="input" {...this.props} />

The errors is.

    error TS2606: Property 'value' of JSX spread attribute is not assignable to target property.

CarsonF:
@rluiten My example compiles with my definition. Do you not have a `value` property on your `Field` class?

CarsonF:
*notices rc4* @erikras I think you're just trying to give me more work lol

erikras:
Yeah, sorry about that. Better to have a breaking change than a shitty API. *Blah, eggs, blah, omelet.*

rluiten:
@CarsonF I never even thought of the fact I still had value on Input tag, that was it. 
I still don't understand type errors to frequently thanks for that.

CarsonF:
@erikras Trying to address the type of the second parameter, _props_, of the `validate` function passed into `reduxForm()`. Just from runtime debugging I don't know which of these props are "public api" vs "internal/just happen to be passed along". 

I was going to start with my FormProps interface and the "custom" props passed to the component. But I see way more than that...it looks like stuff from `Config` / `Form` class.

Any clues?

erikras:
I'm not 100% sure that I understand the question. Can you list them, and I can check off which ones are "public"?

CarsonF:
```ts
@reduxForm({
  validate: (values, props) => {} // <-- type of props?
})
class MyForm {}
```
Chrome is showing me these:
- anyTouched
- array
- arrayMove
- arrayRemoveAll
- blur
- change
- destroy
- dirty
- dispatch (could be custom / from me)
- focus
- form
- handleSubmit
- initialize
- initialized
- invalid
- onSubmit
- pristine
- registeredFrields
- reset
- setSubmitFailed
- setSubmitSucceeded
- shouldAsyncValidate
- startAsyncValidation
- startSubmit
- stopAsyncValidation
- stopSubmit
- submitFailed
- submitSucceeded
- submitting
- touch
- untouch
- updateSyncErrors
- valid
- validate

erikras:
It looks like you have identified some failures of [the props doc](http://redux-form.com/6.0.0-rc.4/docs/api/Props.md/), along with some masking issues.

- anyTouched ✅ 
- array **missing from doc**
- arrayMove **should not be passed**
- arrayRemoveAll **should not be passed**
- blur **should not be passed**
- change **should not be passed**
- destroy ✅ 
- dirty ✅ 
- dispatch (could be custom / from me) ✅ 
- focus **should not be passed**
- form ✅ 
- handleSubmit ✅ 
- initialize ✅ 
- initialized **should not be passed**
- invalid ✅ 
- onSubmit - I think you passed this one
- pristine ✅ 
- registeredFrields **should not be passed**
- reset ✅ 
- setSubmitFailed **should not be passed**
- setSubmitSucceeded **should not be passed**
- shouldAsyncValidate **should not be passed**
- startAsyncValidation **should not be passed**
- startSubmit **should not be passed**
- stopAsyncValidation **should not be passed**
- stopSubmit **should not be passed**
- submitFailed ✅
- submitSucceeded ✅
- submitting ✅ 
- touch **should not be passed**
- untouch **should not be passed**
- updateSyncErrors **should not be passed**
- valid ✅ 
- validate **should not be passed**

Your TypeScript stuff should only address the ones with a ✅, and I need to mask many of them.

erikras:
See also: [propTypes.js](https://github.com/erikras/redux-form/blob/v6/src/propTypes.js)

rluiten:
I am just  updating to rc.4 and the latest type def.

@CarsonF getting this error at moment.

    ERROR in C:\dev\qk-enhanced\Enhanced.Eylm.MyFamilyLounge\typings\modules\redux-form\index.d.ts
    (340,5): error TS2300: Duplicate identifier 'handleSubmit'.

    ERROR in C:\dev\qk-enhanced\Enhanced.Eylm.MyFamilyLounge\typings\modules\redux-form\index.d.ts
    (342,5): error TS2300: Duplicate identifier 'handleSubmit'.


I am getting one other error to do with a reduxForm() wrapped component but I had that error previously, and had not figured out why yet, so I had not updated to your most recent rc.3 typedef because of it.

Its one of those multiline errors that I really wish I could get a report of these fields conflict or these fields are added or removed :).

CarsonF:
@erikras Ok yeah that's what I was thinking. 
- touch and untouch _are_ listed in the props doc though.
- `onSubmit` was defined in the form config so I think that should be masked as well.

In my `FormProps` I have `asyncValidate` and `asyncValidating` which I did not see in Chrome, that's just because async wasn't enabled in form config right?

Dispatch as well? I haven't mentioned it anywhere else yet..

CarsonF:
@rluiten Derp I knew there was a reason I defined it like I did. One sec

CarsonF:
@rluiten What's the other problem you're having?

rluiten:
This is top level form that includes ReactRouter.RouteComponentProps<>
and FormProps<>

Error is.

    ERROR in ./containers/transition.tsx
    (187,4): error TS2345: Argument of type 'typeof Transition & Form<ITransitionFormViewModel, {}, {}>' is not assignable to parameter of type 'ComponentClass<{ stateValue: string; transition: ITransitionFormViewModel; initialValues: ITransi...'.
      Type 'typeof Transition & Form<ITransitionFormViewModel, {}, {}>' is not assignable to type 'StatelessComponent<{ stateValue: string; transition: ITransitionFormViewModel; initialValues: ITr...'.
        Type 'Form<ITransitionFormViewModel, {}, {}>' is not assignable to type 'StatelessComponent<{ stateValue: string; transition: ITransitionFormViewModel; initialValues: ITr...'.
          Type 'Form<ITransitionFormViewModel, {}, {}>' provides no match for the signature '(props?: { stateValue: string; transition: ITransitionFormViewModel; initialValues: ITransitionFormViewModel; editingServices: Object; } & {}, context?: any): ReactElement<any>'

And it it comes from the last line....

    export default connect(mapStateToProps, {
        fetchTransition,
        saveTransition,
        leaveTransitionForm,
        saveTransitionGuardianComplete,
    })(transitionForm);

Not sure if you can mind read me enough to help, I can probably paste more type info if think it might help, I suspect its going to be bear to track down.

CarsonF:
@rluiten Where does `RouteComponentProps` come from?

rluiten:
Type def for the wrapped component parameters.

    export interface TransitionProps extends
        ReactRouter.RouteComponentProps<TransitionListRouteParams, {}>,
        FormProps<VM.ITransitionFormViewModel, {}> {
        stateValue: string;
        transition: IMinimumTransitionFormViewModel; // not needed with redux-form ?
        fetchTransition: fetchTransitionPromise;
        saveTransition: saveTransitionPromise;
        editingServices: VM.ISupportServicesViewModel[];
        leaveTransitionForm: typeof leaveTransitionForm;
        saveTransitionGuardianComplete: saveTransitionGuardianCompletePromise;
    }

With typings from "react-router": "registry:dt/react-router#2.0.0+20160621215300",

CarsonF:
Yeah...again decorators have problems not being able to mutate and with the issue around TS #8674. Not really a problem I can solve. Try `connect<any, any, any>(`

rluiten:
This hides the error, just be nice if I didn't have to, but for now I am not spending more effort on it.

    export default connect(mapStateToProps, {
        fetchTransition,
        saveTransition,
        leaveTransitionForm,
        saveTransitionGuardianComplete,
    })(transitionForm as any);

erikras:
@CarsonF Here you go. [A complete list of the props](https://github.com/erikras/redux-form/blob/v6/src/__tests__/reduxForm.spec.js#L74-L138).

CarsonF:
Perfect. Looks like I'm just missing the `array` one. Is that just for FieldArray's or Field's too?

erikras:
> Looks like I'm just missing the array one. Is that just for FieldArray's or Field's too?

Not sure what you mean. Those props are for the component decorated by `reduxForm()(MyForm)`.

CarsonF:
Ah derp. Of course. Still early for me 😅

erikras:
☕

CarsonF:
@erikras `array` functions in props are documented. Are their function parameters the same as their action creator counterparts (without the form name param)?

erikras:
Oops. Forgot to push the docs. [There](http://redux-form.com/6.0.0-rc.4/docs/api/Props.md/)

CarsonF:
@erikras They don't return functions do they?

erikras:
@CarsonF No, they _are_ functions. They are bound action creators so will return nothing.

erikras:
Although the doc notation is pretty confusing. I'll fix...

CarsonF:
Used like this, right?
```js
this.props.array.removeAll('foo'); // returns void
```

erikras:
🎉 Bingo!

CarsonF:
@erikras Would you mind confirming #1342?

Errorific:
Playing with this briefly, I don't think this exports anything to cover 'redux-form/immutable'. I'm going to take a look in the next few days at what needs to be different for that interface, but I suspect not much.

Also doesn't work with the definitelytyped/types-2.0 version of the react typings (https://github.com/DefinitelyTyped/DefinitelyTyped/tree/types-2.0/react) but thats a problem for a few months further down the track once typescript 2.0 is out and those new bindings get merged in. Just seems to be some react interfaces got changed to generic types.

Errorific:
After some brief investigation I think all thats needed for redux-form/immutable to work is to add [immutable.d.ts](https://gist.github.com/Errorific/7cbbd3775b5bc37c4c3f46c76cac2e84) that re-exports the same bindings. The one thing that should probably change in the immutable typings is `DataShape` should be an Immutable.Map rather than a normal one, but I think the existing type will match an Immutable.Map. I haven't tested any of these assumptions properly, I need to run up a cleaner example project that doesn't have any of the typescript 2.0 stuff mixed in.

waynebrantley:
I took simple form and tried to get it to work using your latest typings.
https://github.com/erikras/redux-form/blob/v6/examples/simple/src/SimpleForm.js

- `placeholder` it says is not valid (everywhere it is used)
- `type` it says is not valid (everywhere it is used)
- `value` it says is not valid (only used in the radio type, which I think you identified above is an edge case - but as of now, will not compile - so needs handled in some fashion)
- `id` it says is not valid (used on employee field)
- `reduxForm` call at the bottom, error on the parameter value of SimpleForm.  Error is TS2345: Argument of type '(props: any) => Element' is not assignable to parameter of type 'typeof Component'.
  Type '(props: any) => Element' provides no match for the signature 'new <P, S>(props?: P, context?: any): Component<P, S>'    This last issue is probably a lack of support for functional stateless component syntax?

CarsonF:
@waynebrantley `Field` has a generic type, if you a wanting a `input` element, then check out the example [above](https://github.com/erikras/redux-form/pull/1318#issuecomment-236655775).

`reduxForm` also has generics. I'll see if I can write up an example

CarsonF:
@Errorific Yeah I've been thinking about Immutable as well. I have some concerns but I'll play around with it first and see if I can get it to behave correctly. Can you test for me once I have something?

CarsonF:
@waynebrantley Yeah you are. `placeholder`, `type`, `value`, and `id` are all defined in [React.HTMLAttributes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/react/react.d.ts#L1912).

I understand wanting to be able to take and use the examples directly though. It's a balance. I'm trying to find it. I don't want those HTMLAttributes properties being code-completed for me for a component from Material UI, since it doesn't accept those.

waynebrantley:
@CarsonF So you are saying I have to define this new InputField type just so I can use this?  Seems pretty messy - I guess that is a typescript thing?  Seems like you would have the Field definition in your typings implement the HtmlAttributes interface.  (In Material UI, you would know not to use them - but they ARE valid in Redux-Form so I think they should be there)

Also, what about the `ReduxForm` call at the bottom?

CarsonF:
@waynebrantley Yes you are right. I was a bit uncomfortable with it to begin with, and hearing the feedback from you and a couple others gave me motivation to work with it some more. I think I have a solution. Commits inbound!

BTW Sorry if I came off short. I appreciate your input! 👍

erikras:
Making more work for you, @CarsonF. #1449 #1554 😍

CarsonF:
@erikras What's up?

huan086:
To support forms that are stateless component, I had to change the FormDecorator to

```ts
interface FormDecorator<FormData extends DataShape, P, S> {
  <T extends (typeof Component)>(component: T | StatelessComponent<P>): T & Form<FormData, P, S>;
}
```

To suppress errors such as `Generic type \'ReactEventHandler\' requires 1 type argument(s).`, I added `<any>` behind those generic types

huan086:
`Fields` is missing from definition

```ts
interface BaseFieldsProps {
  /**
   * An array of string paths, in dot-and-bracket notation, corresponding to a value
   * in the form values. It may be as simple as 'firstName' or as complicated
   * as contact.billing.address[2].phones[1].areaCode.
   *
   * Required but made optional so interface can be used on decorated components.
   */
  names?: string[];

  /**
   * A Component, stateless function, or string corresponding to a default
   * JSX element.
   *
   * Required but made optional so interface can be used on decorated components.
   */
  component?: ComponentConstructor | "input" | "select" | "textarea",

  /**
   * If true, the rendered component will be available with the
   * getRenderedComponent() method. Defaults to false. Cannot be used if your
   * component is a stateless function component.
   */
  withRef?: boolean;

  /**
   * Formats the value from the Redux store to be displayed in the field input.
   * Common use cases are to format Numbers into currencies or Dates into a
   * localized date format.
   */
  format?: Formatter;

  props?: Object;

  /**
   * Parses the value given from the field input component to the type that you want
   * stored in the Redux store. Common use cases are to parse currencies into Numbers into
   * currencies or localized date formats into Dates.
   */
  parse?: Parser;
}

export class Fields<FieldCustomProps> extends Component<BaseFieldsProps & FieldCustomProps, {}> {
  /**
   * true if the current value is different from the initialized value,
   * false otherwise.
   */
  dirty: boolean;

  /**
   * The names prop that you passed in.
   */
  names: string[];

  /**
   * true if the current value is the same as the initialized value,
   * false otherwise.
   */
  pristine: boolean;

  /**
   * The current value of the field.
   */
  values: { [name: string]: FieldValue };

  /**
   * Returns the instance of the rendered component. For this to work, you must
   * provide a withRef prop, and your component must not be a stateless function
   * component.
   */
  getRenderedComponent(): Component<{[name: string]: WrappedFieldProps} & FieldCustomProps, any>;
}
```

waynebrantley:
@erikras @CarsonF 
Why did this get closed??

CarsonF:
I'm not sure. I'm still working on this. Just having a busy few weeks.

aliatsis:
any progress here?

LaurensRietveld:
@CarsonF: just trying out your defs. Would you like me to post any issues in this thread, or submit issues in your fork. Simple issue: the `Selector` interface returns an object, where it can also return a single value  (see http://redux-form.com/6.0.5/examples/selectingFormValues/)

CarsonF:
@LaurensRietveld Post them here. I'll try to take a look at that. 

I've just been busy lately. Still haven't forgotten about this.

grovesNL:
I have the same issues as @waynebrantley, we should be able to use the examples directly without wrappers. The props should be determined by an intersection of `component` props and `Field` props ideally.

joewood:
Any progress with this? Can what's been done so far be merged and raise new tickets for immutable etc...

bbenezech:
My fork: https://gist.github.com/bbenezech/8ebbced86c0f357d09e4000635eea275
Complete 100% typesafe redux-form example: https://gist.github.com/bbenezech/528c7057c62300d41f51b212f0a92ad2

If someone need a FieldArray example, I can help too, I have some in my codebase.

Let's write and exchange some examples through gist, TS with redux is non-trivial, and RF6 with the inversion of control was specially challenging for me.

The TS definition should be merged, it works well, thanks a lot @CarsonF 🎊

joewood:
That's great - @bbenezech could you add the same declarations for "redux-form/react-native" module name too?

bbenezech:
@joewood Sorry I've never used RN, I don't know what it implies and I cannot test it.

MichaelBitard:
@bbenezech nice job, I was using @CarsonF commit for my typings.json `github:CarsonF/redux-form/index.d.ts#0bea952` and now I'll try to use yours. I you have examples with dynamic field arrays with initial values I'm really interested!

bbenezech:
@MichaelBitard the typings are really similar: I just added some API that was missing and that I needed. All credits to @CarsonF 

To push new values, I use 
```javascript
this.props.fields.push({ firstName: 'Sherlock', lastName: 'Holmes' })
// To hydrate I don't do anything special
<PatientForm
  handleSave={this.handleSave}
  initialValues={{ users: [{ firstName: 'Sherlock', lastName: 'Holmes' }] }}
  form="users"
/>
```
I am not sure I understand the issue and how it is related to typings?

zakdances:
Doesn't seem like this has been merged yet.

rluiten:
Lol, seen em.
I recently replaced most of the bulbs in my house (like recently 2 weeks back) with led's.
Still uhmming about the flouro's in kitchen, and maybe flouro's in other parts of house.

I hate CFL's so had only a couple of them.

IRus:
@zakdances do you figure out what happens with definitions?

IRus:
Sooo, after all we doesn't have definitions?

sawatsky:
You removed the onSubmitSuccess and onSubmitFail handlers?
They are still in the documentation: http://redux-form.com/6.2.1/docs/api/ReduxForm.md/

>  **onSubmitFail : Function [optional]**
> 
> A callback function that will be called when a submission fails for whatever reason. It will be called with the following parameters:
> 
> errors : Object
> 
> The errors that caused the submission to fail.
> 
> dispatch : Function
> 
> The Redux dispatch function.
> 
> submitError : Error
> 
> The error object that caused the submission to fail. If errors is set this will be most likely a SubmissionError, otherwise it can be any error or null.
> 
> **onSubmitSuccess : Function [optional]**
> 
> A callback function that will be called when a submission succeeds. It will be called with the following parameters:
> 
> result : Object
> 
> Any result that onSubmit has returned
> 
> dispatch : Function
> 
> The Redux dispatch function.

bjmin:
is this issue completed?

CarsonF:
We're waiting for DefinitelyTyped people to merge my PR. I don't think Eric wants to maintain the definition here.

PR #1345, opened 2016-07-14
Don't overwrite user input on fields not present in initialValues when  `overwriteOnInitialValuesChange: false`

### LINKED ISSUE #1112: "overwriteOnInitialValuesChange: false" still overwrites user input on fields not present in the initialValues
Using 5.2.5, "overwriteOnInitialValuesChange: false" does preserve user input on fields that are present in the initialValues when they change.

However, if a field is *absent* from the initialValues (`initialValues[field] === undefined`), then it does get overriden (losing user input) when the initialValues change.

In src/initializeState.js :

```js
const makeEntry = (value, previousValue, overwriteValues) => {
  return makeFieldValue(value === undefined ? {} : {
    initial: value,
    value: overwriteValues ? value : previousValue
  });
};
```
when `value` is `undefined`, the `overwriteValues` param is ignored, and the field gets back to just `{}` even if there was a `previousValue`

tomazy:
Fixes #1112

### DIFF
--- a/src/__tests__/createReduxForm.spec.js
+++ b/src/__tests__/createReduxForm.spec.js
@@ -2255,7 +2255,7 @@ describe('createReduxForm', () => {
 
     const Decorated = reduxForm({
       form,
-      fields: [ 'firstName', 'lastName' ],
+      fields: [ 'firstName', 'lastName', 'instrument' ],
       overwriteOnInitialValuesChange: false
     })(Form);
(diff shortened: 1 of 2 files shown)


erikras:
Published as [`v5.3.2`](https://github.com/erikras/redux-form/releases/tag/v5.3.2). Cheers! 🍺

PR #1359, opened 2016-07-17
fix #1358

### LINKED ISSUE #1358: Uncaught TypeError: Cannot assign to read only property 'value' of object '#<Object>'
This error occurs in `normalizeField`: 
```js
finalField.value = normalizer(finalField.value, previousState && previousState[field.key] && previousState[field.key].value, values, previousValues); 
```
So what can I do so solve it?

flosse:


erikras:
Released as [`v5.3.3`](https://github.com/erikras/redux-form/releases/tag/v5.3.3).

PR #1369, opened 2016-07-19
event with value key shouldnt be special case

### LINKED ISSUE #1355: Value shouldn't look for a prop within the object
Im having issues creating complex value types that instead of holding strings hold objects/bools/arrays or any other thing.
But this really puzzled me https://github.com/erikras/redux-form/blob/v6/src/events/getValue.js#L41
I had to read the source to figure out what was going on.
Turns out calling `onChange` with an object that has a `value` key will use the value of that key as the new value, which is very cryptic.
Please remove this

eloytoro:
Closes #1355

### DIFF
--- a/src/events/__tests__/getValue.spec.js
+++ b/src/events/__tests__/getValue.spec.js
@@ -18,13 +18,13 @@ describe('getValue', () => {
     expect(getValue('dog', false)).toBe('dog')
   })
 
-  it('should unwrap value if non-event object containing value key is passed', () => {
-    expect(getValue({ value: 5 }, true)).toBe(5)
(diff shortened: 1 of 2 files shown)


eloytoro:
hmm tests passed dont know why it reports that it didnt

erikras:
This looks good. I will evaluate when I get back from vacation next week.

erikras:
FYI: Your test problem was a linting error with a stray semicolon.

erikras:
Published in [`v6.0.0-rc.4`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.4).

PR #1370, opened 2016-07-19
[v6] DOC broken immutable example

fabriziomoscon:
I am not able to understand the immutable example, so I decided to run it locally... just to find out that it was not using the `reducer.js` file

So I decided to amend the example, but I receive a JS error in console:
```
_immutable.reducer.validation is not a function
```
I thought it might be a good idea to PR my changes to have something to fix.

Eventually I am interested in an example with immutable and initialValues to use in a react-native app.

Note: To run the example locally a user must `npm install` the root of the project not just the individual example.


erikras:
Thanks!

Issue #1088, opened 2016-06-01
Errors in createOnDragStart and createOnDrop?


stevengoldberg:
I've been seeing rare errors popping up in my logs from within redux-form and I'm wondering if anyone else has seen these or can offer any insight. I have not been able to repro these myself.

There's this one from `createOnDrop`:
```js
Error: Invalid argument. 
~/redux-form/lib/events/createOnDrop.js?fe65 in Anonymous function at line 9:0

7.  var createOnDrop = function createOnDrop(name, change) {
8.   return function (event) {
9.     change(name, event.dataTransfer.getData(_createOnDragStart.dataKey));
10.  };
11. };
```

And this one from `createOnDragStart`:
```js
Error: Unexpected call to method or property access.
~/redux-form/lib/events/createOnDragStart.js?eaa1 in Anonymous function at line 7:0

4. var dataKey = exports.dataKey = 'value';
5. var createOnDragStart = function createOnDragStart(name, getValue) {
6.  return function (event) {
7.    event.dataTransfer.setData(dataKey, getValue());
8.  };
9. };
```

These occurred with v. 5.2.3, and always with IE11.

ooflorent:
In a few cases `dataTransfer` might not be available. Accesses to `getData` or `setData` would then raise and error.

stevengoldberg:
I looked into this a bit more and [this seems a likely answer to me](http://stackoverflow.com/questions/26213011/html5-dragdrop-issue-in-internet-explorer-datatransfer-property-access-not-pos).

>getData() and setData() attribute must be called exactly "text"

The first param to `setData` and `getData` specifies format, and `redux-form` is passing `"value"`.

erikras:
Fix published in [`v6.0.0-rc.4`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.4).

prevostc:
Still have this issue on v6.4.2

When I initiate a drag on an input I get a createOnDragStart.js:9 Uncaught TypeError: Cannot read property 'setData' of undefined(…)

I don't know what might help to debug this.

PR #1373, opened 2016-07-19
Fix drag and drop on IE

alurim:
Bumped into https://github.com/erikras/redux-form/issues/1088 and the solution to stop these errors (and to unbreak drag and drop on IE) is to just change the dataKey to 'text. 

The [stack overflow question](http://stackoverflow.com/questions/26213011/html5-dragdrop-issue-in-internet-explorer-datatransfer-property-access-not-pos) mentioned in the referenced issue suggests this as well FWIW.

### DIFF
--- a/src/events/__tests__/createOnDrop.spec.js
+++ b/src/events/__tests__/createOnDrop.spec.js
@@ -12,15 +12,19 @@ describe('createOnDrop', () => {
   it('should return a function that calls change with result from getData', () => {
     const change = createSpy()
     const getData = createSpy().andReturn('bar')
+    const preventDefault = createSpy()
     createOnDrop('foo', change)({
-      dataTransfer: { getData }
+      dataTransfer: { getData },
+      preventDefault
     })
     expect(getData)
       .toHaveBeenCalled()
       .toHaveBeenCalledWith(dataKey)
     expect(change)
       .toHaveBeenCalled()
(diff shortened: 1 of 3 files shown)


erikras:
Still works on other browsers?

alurim:
@erikras I've tested Safari, Chrome, and Firefox and they all look good. My initial commit stopped the error from firing but didn't fix the fact that dropping into a field wasn't working (because I was testing dropping text that matched the field name which was the only thing that was working). I discovered the following problem:

We're calling `change` with `name` when that function is already bound to the correct field's name (for example the [`createOnChange`](https://github.com/erikras/redux-form/blob/v6/src/events/createOnChange.js#L5) function is calling `change` correctly). 

I fixed that problem by removing `name` from the args of `change` (the code could be cleaned up a bit to avoid passing it in but I didn't want to mess with your library more than I needed to for this PR). 

This uncovered another issue. Now we're finally calling `change` correctly on drop, but some browsers do this drop action for you and will end up firing a change event. This results in the input having the value dropped twice. To avoid this from occurring I added a preventDefault on the event which seems to do the trick.

erikras:
Published in [`v6.0.0-rc.4`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.4). Thanks again!

alurim:
No worries, thanks for merging the PR. One less error in our logs!

PR #1404, opened 2016-07-26
Small doc update registeredFields - I believe this field has changed I don't know when it changed

rluiten:
Just noticed it comparing documentation to a logged value.

### DIFF
--- a/docs/api/ReduxForm.md
+++ b/docs/api/ReduxForm.md
@@ -189,10 +189,6 @@ component.
 
 `true` when the current form values are different from the `initialValues`, `false` otherwise.
 
-#### `fieldList : Array`
-
-An array of strings representing all the fields in the form. Mainly useful for testing.
-
 #### `invalid : boolean`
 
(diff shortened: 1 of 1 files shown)


erikras:
This was semi-intentionally left undocumented, but I guess it probably should be. So thanks.

olslash:
@erikras, is this still correct?

https://github.com/erikras/redux-form/blob/6401f982cc08c030a4440ee95d29c50865454521/src/reduxForm.js#L687

screen shot 2017-02-21 at 2 51 11 pm

gustavohenke:
[If it's not documented, then it doesn't exist](https://medium.com/@fagnerbrack/library-internals-do-not-exist-1ff50a30e3b0).

olslash:
The PR removed fieldList and added registeredFields -- I see fieldList but registeredFields doesn't exist as document

gustavohenke:
> I see fieldList

You see it because, well, it does exist. But the behavior should not be relied upon, *because it's not documented*.

> registeredFields doesn't exist as documented

It's also not documented.

PR #1412, opened 2016-07-28
Array.findIndex() polyfill

### LINKED ISSUE #1326: [V6] call to Array.findIndex breaks IE
V6 seems to use [findIndex](https://github.com/erikras/redux-form/blob/v6/src/reducer.js#L236) which needs a [polyfill](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) in some browsers.

I'd be happy to create a documentation PR if i'm not missing something here.

erikras:
Fixes #1326

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
   },
   "homepage": "https://redux-form.com/",
   "dependencies": {
+    "array-findindex-polyfill": "^0.1.0",
     "deep-equal": "^1.0.1",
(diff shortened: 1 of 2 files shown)


erikras:
Published in [`v6.0.0-rc.4`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.4).

PR #1422, opened 2016-07-29
Fix errors getting lost

jasontwong:
Why:

* Non `SubmissionError` errors like `ReferenceError` are silenced and
  makes debugging difficult inside submission logic.

This change addresses the need by:

* Throw the error again if it's not a `SubmissionError`.
* The error is thrown in `setTimeout` for a `Promise` so we get the real
  error as opposed to an uncaught rejection.

### DIFF
--- a/src/handleSubmit.js
+++ b/src/handleSubmit.js
@@ -15,7 +15,10 @@ const handleSubmit = (submit, props, valid, asyncValidate, fields) => {
       try {
         result = submit(values, dispatch)
       } catch (submitError) {
-        const error = submitError instanceof SubmissionError ? submitError.errors : undefined
+        if (!submitError instanceof SubmissionError) {
+          throw submitError
+        }
+        const error = submitError.errors
         if(onSubmitFail) {
           onSubmitFail(error, dispatch)
         }
@@ -32,6 +35,9 @@ const handleSubmit = (submit, props, valid, asyncValidate, fields) => {
             return submitResult
           }, submitError => {
             const error = submitError instanceof SubmissionError ? submitError.errors : undefined
+            if (!error) {
+              setTimeout(() => throw submitError)
+            }
             stopSubmit(error)
             if(onSubmitFail) {
               onSubmitFail(error, dispatch)


sknoslo:
Someone please correct me if I'm wrong, but I think this is probably intended behavior. You should handle other error types *before* redux-form even sees it. From the [docs](http://redux-form.com/6.0.0-rc.3/docs/api/SubmissionError.md/):

``` jsx
<MyForm onSubmit={values =>
  ajax.send(values) // however you send data to your server...
    .catch(error => {
      // how you pass server-side validation errors back is up to you
      if(error.validationErrors) {
        throw new SubmissionError(error.validationErrors)
      } else {
        // what you do about other communication errors is up to you

>>>>>>> // you could log the error, display an alert, etc...
      }
    })
}/>
```

jasontwong:
I'm trying to catch javascript errors that are supposed to bubble up to the stable running thread... for example:

```es6
<MyForm onSubmit={values =>
  ajax.send(values) // however you send data to your server...
    .catch(error => {
      // how you pass server-side validation errors back is up to you
      const { bar } = this.foo
      if(error.validationErrors) {
        throw new SubmissionError(error.validationErrors)
      } else {
        // what you do about other communication errors is up to you

>>>>>>> // you could log the error, display an alert, etc...
      }
    })
}/>
```

In the case above, no errors are ever reported and there is no way to know that you're submission failed unless you have the `onSubmitFail` prop defined. And even if you do have it defined, all you get in the error is `undefined` which has no meaning.

src/handleSubmit.js
```diff
@@ -32,6 +35,9 @@ const handleSubmit = (submit, props, valid, asyncValidate, fields) => {
             return submitResult
           }, submitError => {
             const error = submitError instanceof SubmissionError ? submitError.errors : undefined
+            if (!error) {
+              setTimeout(() => throw submitError)
+            }
             stopSubmit(error)
```
erikras: Doesn't using `setTimeout()` make your error uncatchable? Why not just return it, so code paying attention to the promise can see it?

jasontwong: The `setTimeout()` solution is slightly opinionated. With this solution, if some kind of unexpected error like `TypeError`, `ReferenceError` and other system errors, this will actually give you file name and line number of the original error. If you just do a regular throw/catch through the promise you lose that information. I'm happy to conform to whatever standard the team feels is best.

erikras: Okay. So you're going for log-ability rather than catch-ability.

jasontwong: Yes, that's correct. It would work just as if there was an error in a plain javascript function.

PR #1426, opened 2016-07-30
Use the `for` attribute for label elements

ivarni:
Assistive technologies makes good use of `for` in order
to help people who depend on screenreaders.

I believe in making the web an inclusive place and I think libraries like this promoting best practices is a good place to start. Including the `for` attribute in the examples increases the chance that developers who copy and paste from the docs will include it in their own markup.

Your PR, should you choose to accept it, will add the `htmlFor` attribute to `<label>` elements in the examples, increasing the chance that other developers will include it in their codebase.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -47,15 +47,15 @@ class ContactForm extends Component {
     return (
       <form onSubmit={handleSubmit}>
         <div>
-          <label>First Name</label>
+          <label htmlFor="firstName">First Name</label>
           <Field name="firstName" component="input" type="text"/>
         </div>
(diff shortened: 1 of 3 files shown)


ivarni:
It seems it will also delete some trailing whitespace. If you were using it for anything I'm sorry about that.

erikras:
👍

chmanie:
Sadly this won't work like this. The `for=` attribute is given the [corresponding element id](https://developer.mozilla.org/de/docs/Web/HTML/Element/label) of the linked input field. All the input fields in the examples lack ids so we might want to add these as well.

Issue #853, opened 2016-04-24
Feature Request: submitSucceed prop


aight8:
Because when submitFailed is false it's not necessary submitted and valid.

aight8:
**dump**
Easy use case:
I wan to show a success message when the user successful submitted a form. When submitFailed don't necessary means it's submitted too. Is there a redux-form way to solve this? (without set an own state)

ooflorent:
Have you tried using `onSubmitSuccess`? It was added [v5.1.0](https://github.com/erikras/redux-form/releases/tag/v5.1.0). It is not a boolean flag but a callback that gets invoked once the submit has succeeded.

aight8:
Its not in v6. But beside that, why there is no submitSuccess property.
There is on submitFail, on  submitSuccess callback but only a submitFailed property. Why isn't the submitSucess property exists too?

ooflorent:
Feel free to submit a PR to add it! I'll happily review it!

kevzettler:
is this the same idea as https://github.com/erikras/redux-form/issues/94#issuecomment-224716048 ?

gustavohenke:
Done thanks to #1428.

Happy new year, guys!

PR #1428, opened 2016-07-30
[V6] Add submitSucceeded prop (related to issue #853)

scboffspring:
Hi,

This pull request adds the submitSucceeded prop proposed and asked in issue #853.
It also includes a `setSubmitSucceeded` which isn't used yet, but can be useful in the future.

Thanks for this great library, v6 is really awesome !


erikras:
Wow! Excellent work, @scboffspring! Perfect PR. 💯

erikras:
Well. Did you not leave out where it is actually called,  in `handleSubmit()`?

scboffspring:
That's what I found "weird". `setSubmitFailed` was only called part of asyncValidation, and not part of the doSubmit
(here https://github.com/erikras/redux-form/blob/v6/src/handleSubmit.js#L54 and here https://github.com/erikras/redux-form/blob/v6/src/handleSubmit.js#L64)

The failure is otherwise handled part of the STOP_SUBMIT reducer that will check _error fieldErrors.

niksajanjic:
Flag submitSucceeded is very helpful when you just want to show a message that your form got submitted successfully. But after successfully submitted form and flag set to true I would like to have that flag set back to false once you touch any part of the form again (similar to how error prop works set by SubmissionError _error flag, if you touch any part of the form error disappears).

May I know why isn't it working like that?

Imagine profile page which is full of fields of your own data. All those fields are going to be prepopulated with data that already exists in your store. Once you change that data and update it you get a success message thanks to sumbitSucceeded flag which becomes true. Now, if you decide you want to change something again and update for the second time, flag stays true all the time and user has no clue whether that flag is from the first time or the second.

The way I fixed it is by using submitSucceeded flag along with submitting flag. Basically, I only show success message if submitSucceeded flag is true but submitting is false and it works this way. But I would really love to remove success message once you touch a form again as it is obvious you are trying to do another form submission. And this can't be done with dirty/pristine flag because if you touch any field and then change it back to what it was before, success message will disappear at first and then get back once you change value to it's initial.

rjhilgefort:
Why hasn't this made it into the docs yet? I wasted time because this flag wasn't documented. Thanks for the work!

gustavohenke:
@rjhilgefort 

> Why hasn't this made it into the docs yet?

What are you talking about? `submitSucceeded` is documented since ever:
- http://redux-form.com/6.0.0-rc.4/docs/api/Props.md/#-submitsucceeded-boolean-
- http://redux-form.com/6.1.0/docs/api/Props.md/#-submitsucceeded-boolean-
- http://redux-form.com/6.2.0/docs/api/Props.md/#-submitsucceeded-boolean-
- http://redux-form.com/6.4.3/docs/api/Props.md/#-submitsucceeded-boolean-
- http://redux-form.com/6.6.0/docs/api/Props.md/#-submitsucceeded-boolean-

> I wasted time because this flag wasn't documented.

The main feature of this PR *is* documented, you're probably just looking at the wrong place.
And please use wiser words the next time, otherwise people may get mad at you.

rjhilgefort:
@gustavohenke My apologies. Thanks for the work, love the library!

Seems to just have been an SEO issue pulling up the most popular version of redux form from a google search and and then navigating there within. I hadn't noticed the URL for the version (seems to be no other prominent call out in any other place.

While nice to have doc instances for all versions, without a prominent display of the version on the page, it can add confusion and noise. I doubt I'm the only one to make this mistake. I think prominently displaying a version tag below the Redux Form logo would be extremely helpful.

Again, thanks for the work on this tool- it's made my life much easier and is generally a pleasure to interface with.

Luckygirlllll:
Is anyone can show example how to use submitSucceeded now?

TimOgilvy:
It doesn't look like this has made it into the docs yet. Is it in the current stable release?

I have a feeling I'm going to extend this FR further with a conversation about field specific feedback - I need to resolve whether my API or my front-end will be providing feedback on password quality first.

Would be nice to see this conversation progress to include more thinking about affirming positive feedback.  I realise I'm getting into the realms of advocacy here, but I would like us to all think about how to "help Gran change her password"...  maybe it's an edge opinion but good user forms IMO should be affirming when entry is correct. (without being so full of green popups that they get annoying)

PR #1447, opened 2016-08-03
Pruned extraneous props from HOC

erikras:
The props documented and passed to the wrapped form component was done sloppily, as pointed out in https://github.com/erikras/redux-form/pull/1318#issuecomment-237075918. This PR cleans them up to only officially documented props.


erikras:

PR #1449, opened 2016-08-03
pure option

edvinerikson:
Adds `pure` option.

Needed if a parent component re-renders and a child component fetches state from context / any other source that isn't checked in the `shouldComponentUpdate`.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -88,6 +88,7 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
         'initialized',
         'invalid',
         'pristine',
+        'pure',
         'reset',
         'submitFailed',
         'submitSucceeded',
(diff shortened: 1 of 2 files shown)


erikras:
This defeats much of the performance optimization of `v6`, and will result in your entire form rerendering on every single keypress in your form.

Also, I'm unsure about the use of the word `pure` for this. How about `alwaysUpdate`?

edvinerikson:
I took the `pure` option from Redux. We can't rely on that every component will be pure, for example I use react-intl which uses context to get translations but rely on that the `IntlProvider`component re-renders the tree. If we short-circuit the re-render somewhere in the middle the components underneath won't get the new translations.  

Adding a `pure` option is the easiest and safest way to ensure that we are compatible with data that get's mutated. 

Another option (will break if using mutation) is to also diff context the same way we diff props.
Adding both a option to turn of shouldComponentUpdate and add a diff for context would probably be good.

CarsonF:
@edvinerikson I ran into the same problem. What I did was created a Field decorator that passed the intl context in as props. That way they were checked in `reduxForm`'s logic. 

Unless I'm missing something isn't this a more performant solution?

erikras:

Issue #1462, opened 2016-08-04
[v6 rc4] 'Values' not passed to props of wrapped component


jsdmc:
Values should be passed to wrapped component.

jsdmc:
Ah, ok. 
Didn't knew about this change. So, now only use selectors to get values from state?
Probalbly docs should be updated.

jsdmc:
My fault, docs are fine. 
So for anyone who will read this. Values can be accessed either using provided selectors to read from state, or accessing `ownProps` in `connect((undefined, { values }) => ({ values }))`

erikras:
No worries. We all make mistakes like this.

PR #1463, opened 2016-08-04
added 'values' to props that passed to Wrapped comp

jsdmc:


jsdmc:
closed because of #1462

PR #1465, opened 2016-08-04
[v6] When enableReinitialize is true, keep the value of dirty fields to avoid stomping on user edits.

hathawsh:
The new enableReinitialize feature is good, but it can replace edits made by the user. This patch slightly changes the meaning of enableReinitialize so that dirty fields keep the value set by the user, while pristine fields get reinitialized. What do you think?


erikras:
First of all, great PR; try thorough, although you left out updating the action creator docs.

It's not clear to me that this is expected behavior. In my mind, if I reinitialize something, I'd expect it all to get overwritten. So at the very least, `keepDirty` should default to `false`.

I'm curious as to what your use case is for this, because my imagination is failing me. When would you want to keep the users's changes?

hathawsh:
Thanks for looking at this!

In an app I'm building, we've decided we want most forms to auto-save. That means the user types a little in a field, a timeout expires, and the form quietly submits while the user continues to type. The server may take some time to respond (especially when the Internet connection isn't good) and the user could easily type a lot before the submit process finishes; the last thing we want to do is lose their work. This idea applies generally to any situation where the form has live updates or continues to be editable after pressing the submit button.

The name of the flag "enableReinitialize" is a bit of a misnomer here. I don't really want to "reinitialize" the form; what I want is live updates of the initial values, and I want the form to sort out what it means to apply live updates. Sometimes it means resetting a field value and sometimes it doesn't. Maybe we should add a "liveUpdate" flag?

I'll wait for your feedback before making any more changes to this PR.

src/reduxForm.js
```diff
@@ -119,7 +119,7 @@ const createReduxForm =
             if (nextProps) {
               const { enableReinitialize } = this.props
               if ((enableReinitialize || !nextProps.initialized) && !deepEqual(this.props.initialValues, nextProps.initialValues)) {
-                this.props.initialize(nextProps.initialValues)
+                this.props.initialize(nextProps.initialValues, true)
```
erikras: I'm pretty sure this should default to `false`. Perhaps there could be a `keepDirtyOnInitialize` config parameter or something?

erikras:
That makes a lot of sense. 👍

hathawsh:
OK, I'll add a "keepDirtyOnReinitialize" flag along with documentation. Although I'd like a simpler name like "live", the interaction between "enableReinitialize" and "live" would probably be confusing to users, so I think I'll stick to a long but clear name.

hathawsh:
The keepDirtyOnReinitialize option is now ready. The important commit is this one:

https://github.com/erikras/redux-form/pull/1465/commits/810529af3c2b74d1d192899acc6f1e2aa2e78bf1

I merged the v6 branch to make sure everything still works together.

erikras:
Thanks, @hathawsh!

erikras:

PR #1476, opened 2016-08-05
Use redux-validate for examples

ashtonwar:
This is **not** a serious pull request. I wanted to:

* Draw attention to a recent rewrite of [redux-validate](https://github.com/ashtonwar/redux-validate), it adds several features including async & deep key support.
* Make sure redux-validate works well with redux-form.

I think the changes to the `fieldArrays` example are particularly interesting, the code seems a lot easier to understand with the addition of redux-validate.

### DIFF
--- a/examples/fieldArrays/src/validate.js
+++ b/examples/fieldArrays/src/validate.js
@@ -1,48 +1,14 @@
-const validate = values => {
-  const errors = {}
-  if(!values.clubName) {
-    errors.clubName = 'Required'
-  }
-  if (!values.members || !values.members.length) {
-    errors.members = { _error: 'At least one member must be entered' }
-  } else {
-    const membersArrayErrors = []
-    values.members.forEach((member, memberIndex) => {
-      const memberErrors = {}
-      if (!member || !member.firstName) {
-        memberErrors.firstName = 'Required'
-        membersArrayErrors[memberIndex] = memberErrors
-      }
-      if (!member || !member.lastName) {
-        memberErrors.lastName = 'Required'
-        membersArrayErrors[memberIndex] = memberErrors
-      }
-      if (member && member.hobbies && member.hobbies.length) {
-        const hobbyArrayErrors = []
-        member.hobbies.forEach((hobby, hobbyIndex) => {
-          if (!hobby || !hobby.length) {
-            hobbyArrayErrors[hobbyIndex] =  'Required'
-          }
-        })
-        if(hobbyArrayErrors.length) {
-          memberErrors.hobbies = hobbyArrayErrors
-          membersArrayErrors[memberIndex] = memberErrors
-        }
-        if (member.hobbies.length > 5) {
-          if(!memberErrors.hobbies) {
-            memberErrors.hobbies = []
-          }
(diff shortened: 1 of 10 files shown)


examples/fieldArrays/src/validate.js
```diff
@@ -1,48 +1,14 @@
...
+const validate =
+  _validate(
+    [ 'clubName', 'members[].firstName','members[].lastName', 'members[].hobbies[]' ],
+    'Required')
+  .then(
+    'members<._error>',
+    members => (!members || !members.length) && 'At least one member must be entered')
+  .then(
+    'members[].hobbies<._error>',
+    hobbies => hobbies && hobbies.length > 5 && 'No more than five hobbies allowed')
```
erikras: Yes, this example is particularly intriguing.

erikras:
This is very interesting, @ashtonwar. The power of `redux-validate` was not obvious to me from its readme. You need to put some more words in there about each of the code examples. 😄

While, as you noted, it does not make sense for the standard examples to use a third party validation library, it _would_ make sense for you to copy some or all of the examples from this repo over to an `examples` folder in your repo. I would be more than happy to include a "Does this code look disgusting? Check out `redux-validate`!" comment in my examples, especially the `FieldArray` one.

ashtonwar:
Linking to redux-validate in the examples would be really cool. I'll probably copy across the examples sometime next week & setup an interactive github.io demo based on them like you have.

In regards to improving the readme: I'm not sure which words to use or the order to put them in. It'd be awesome if you could open a PR with some revisions.

It occurred to me if if you're using field arrays a lot it might be helpful to define some higher-order validation functions like this (not sure whether it's suitable for documentation):

```js
const moreThan = (ammount) => (value, key) => {
  if (isNaN(Number(value))) return 'Not a number'
  if (Number(value) <= ammount) return
  // convert 'members.0.hobbies._error' to 'hobbies'
  key = key.split('.').slice(-2)
  key = key[1] === '_error' ? key[0] : key[1]
  return `No more than ${stringify(ammount)} ${pluralize(key, ammount)} allowed`
}
```

Not 100% clear on what you meant by "Cool! It seems oriented to mostly Required validation. README would benefit from some < or > validations as well. It's non-obvious" on Twitter.

Did you mean:

1. Include more key manipulation examples with <>'s and {}'s?
2. Examples that show expressing errors in terms of other errors (e.g: if height *or* weight is invalid then so is BMI)?

2 Could be done like:
```js
const validate = values => {
  let errors = _validate(['title', 'content'])(values)
  errors =
    _validate({title: title => title.replace('required', 'needed')})
      .then(null, error => error)(errors)
  return errors
}
validate() = {title: 'Title is needed', content: 'Content is required'}
```

With a `next` helper that could become:

```js
_validate('title').then('content').next('title', title => title.replace('required', 'needed'))
```

erikras:
By < and >, I meant `if (min > max) { errors.min = 'Too big' }`; I wasn't referring to your use of `<` and `>` in your rule notation.

erikras:
I'm gonna close this, but I'm glad it's around to reference. Thanks.

PR #1486, opened 2016-08-08
UNREGISTER_FIELD should not remove fields that don't exist

### LINKED ISSUE #1460: [v6 rc4] Wizard form fails validation after a few steps.
Hi!

Using the technique from the rc.4 Wizard Form Example, I've run into some strange issues, where validation stops working after some pages. I've tried swapping the order of the pages, and they work fine, as long as they are moved to earlier steps. I use disabled={pristine || submitting || invalid} on the submit buttons, but when the bug occurs the form says it's valid, even though it should fail and not be able to submit.

I have compared the code for a few hours, and it seems identical to the sample, except for more steps (5) and other Fields (radio buttons, checkboxes etc.)

I'll try and give more information on request, but unfortunately I don't have time to make a complete public test case at this moment.

I think the issue was also on rc.3, but alas I cannot easily switch back to test, because of all the changes that happened in rc.4.

Best Regards

jervi:
`findIndex()` returns -1 for non existing values. When feeding -1 to
the `splice()` method, the first value of the `registeredFields` array
is erroneously removed.

Fixes #1460.

### DIFF
--- a/src/__tests__/reducer.unregisterField.spec.js
+++ b/src/__tests__/reducer.unregisterField.spec.js
@@ -21,6 +21,22 @@ const describeUnregisterField = (reducer, expect, { fromJS }) => () => {
     expect(state)
       .toEqual(initialState)
   })
+  
+  it('should do nothing if the field is not registered', () => {
+    const state = reducer(fromJS({
+      foo: {
+        registeredFields: [ 
+          { name: 'bar', type: 'field' }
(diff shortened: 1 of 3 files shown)


erikras:
Good catch! Could you devise a test that fails before the fix and passes after the fix?

erikras:

PR #1489, opened 2016-08-08
[v6] Fixed the pristine/dirty state of fields that get reinitialized to an empty value.

hathawsh:
Here's a simple fix. I noticed that if a field has a non-empty initial value, then I clear the field and reinitialize the form with an empty string, the field should become pristine, but does not (as of 6.0.0-rc.4). This happens because the code treats an empty string as no value at all and falls back to the original initial value. This pull request fixes the problem by expecting 'undefined' to be the only value that indicates the initial value is missing. Test included.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -73,7 +73,8 @@ const createConnectedField = ({
   const connector = connect(
     (state, ownProps) => {
       const formState = getFormState(state)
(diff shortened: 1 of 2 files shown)


erikras:
Very interesting bug. It took me a while to recreate it and verify, but you explained it perfectly.

Great PR! 👍

PR #1492, opened 2016-08-08
Fixed splice methods and added tests for ImmutableJS splice.

### LINKED ISSUE #1490: [6.0.0-rc.4] arraySplice doesn't work properly
Calling `arraySplice` with the plain structure doesn't work properly. If you pass a `removeNum` and a `value`, then the values get removed but nothing gets added. This is the logic:

```javascript
const splice = (array = [], index, removeNum, value) => {
  const copy = [ ...array ]
  if (removeNum) {
    copy.splice(index, removeNum)   // removing
  } else {
    if(index < copy.length) {
      copy.splice(index, 0, value)  // adding
    } else {
      copy[index] = value           // outside range, so just set it
    }
  }
  return copy
}

export default splice
```

In the first branch it should be `copy.splice(index, removeNum, value)`.

The logic for `ImmutableJS` isn't correct either:

```javascript
  splice: (list = List.isList(list) || List(), index, removeNum, value) => {
    if(removeNum) {
      return list.splice(index, removeNum)  // removing
    }
    if(index < list.size) {
      return list.splice(index, 0, value)   // inserting
    }
    return list.set(index, value)         // inserting after end of list
  }
```

When adding using this method it will first remove a value then override the remaining value in that position.

clayne11:
Splice methods didn't use to add the value you passed them into the
value array. Fixed this logical issue for both plain and immutable
structures.

Fixes #1490.

### DIFF
--- a/src/__tests__/FieldArray.spec.js
+++ b/src/__tests__/FieldArray.spec.js
@@ -206,7 +206,7 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
       })
       expect(props.fields.removeAll).toBeA('function')
     })
-    
+
     it('should provide swap', () => {
       const props = testProps({
         values: {
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -20,10 +20,11 @@ const createReducer = structure => {
   const rootKeys = [ 'values', 'fields', 'submitErrors', 'asyncErrors' ]
   const arraySplice = (state, field, index, removeNum, value) => {
     let result = state
+    const nonValuesValue = value != null ? empty : undefined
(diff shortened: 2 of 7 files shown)


clayne11:
Hey @erikras any chance you can have a look at this? I added tests for the logic I added and all the tests are passing.

I'm going to have to publish a fork today or tomorrow if this doesn't get merged and published because I need this feature for my production app.

I built my branch and copied over the build into my app and I've been testing with it as well and it's working as intended.

erikras:
Good work. Thanks! 👍

clayne11:
Thanks!

erikras:

PR #1508, opened 2016-08-10
Allow Fields to have an "_error" assigned to them

### LINKED ISSUE #1107: V6 _error on object does not prevent submitting
I have a nested object for witch I generate an `_error` field in the errors object (sync validation).

I'm using this to be able to have validation on individual fields in the object itself and for the nested object.
(let's suppose it's comparing passwords and wanting to set the error on the `passwords` field:
```js
{
  passwords: {
    password: '',
    passwordConfirm: ''
  }
}
```

I want an error on `passwords.password` *and* on `passwords`: 
```js
errors = {
  passwords: {
    _error: 'passwords do not match',
   password: 'password too weak'
  }
}
```
(this example is a bit contrived)

`_error` just works for `Field`, not only for `FieldArray`. But the `_error` does not prevent submitting.
also, `valid` gives the 'wrong' warning.

I'm using `<Field name="passwords" component={DisplayError}>` to display the error on the nested object (adding a Field passwords._error will not work because it does not display the error).

using v6.0.0-alpha.15

clayne11:
Fields can have object values, in which case we need a way to set an
error at the object level. The library will pass "_error" to the Field as
the "error" prop to be displayed, but isValid wasn't respecting it as an error.

Also reused `isValid` selector in the `reduxForm` `connect` function so that we're not duplicating logic.

Fixes #1107

### DIFF
--- a/src/__tests__/hasError.spec.js
+++ b/src/__tests__/hasError.spec.js
@@ -110,6 +110,21 @@ const describeHasError = (name, structure, expect) => {
         expect(hasError(field, null, null, error)).toBe(false)
       }
     })
+
+    it('should return true if a Field that has an object value has an _error', () => {
+      const field = fromJS({ name: 'foo', type: 'Field' })
+      const plainError = {
+        foo: {
+          _error: 'An error'
+        }
+      }
+
+      expect(hasError(field, plainError)).toBe(true)
+
+      const error = fromJS(plainError)
+      expect(hasError(field, null, error)).toBe(true)
+      expect(hasError(field, null, null, error)).toBe(true)
+    })
   })
 }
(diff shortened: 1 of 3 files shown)


erikras:
Nice work. 👍

erikras:

erikras:
@clayne11, dude... Reducing with string concatenation? 😎 

diff for range 2016-08-24 11-00-03

clayne11:
Wow - I must have been half asleep when I wrote that... a little embarrassing.

erikras:
We all do stupid 💩 like that, and I respect your skills enough to joshingly throw it in your face. 😄

PR #1509, opened 2016-08-10
Pass "submitting" as prop to ConnectedField and ConnectedFieldArray

### LINKED ISSUE #1122: [v6] Fields don't update in response to `submitting` flag
The `ConnectedField.shouldComponentUpdate` looks like this:

```javascript
shouldComponentUpdate(nextProps) {
  return !deepEqual(this.props, nextProps)
}
```

The problem is that the `submitting` flag gets set in `this.context._reduxForm`. This essentially means that Fields don't update in response to the submitting flag, and so if you want to disable / enable fields based on the submitting state, the Field will not currently update.

We need to add context to `shouldComponentUpdate` and check the `submitting` flag.

```javascript
shouldComponentUpdate(nextProps, nextState, nextContext) {
  return this.context._reduxForm.submitting !== nextContext._reduxForm.submitting ||
    !deepEqual(this.props, nextProps)
}
```

This same change also needs to be made in `ConnectedFieldArray`.

clayne11:
This will force the component to re-render when submitting so that we
can disabled / enable components based on submit state.

Fixes #1122

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -76,6 +76,7 @@ const createConnectedField = ({
       const initialState = getIn(formState, `initial.${name}`)
       const initial = initialState === undefined ? propInitialValue : initialState
       const value = getIn(formState, `values.${name}`)
+      const submitting = getSyncError(getIn(formState, 'submitting'))
       const syncError = getSyncError(getIn(formState, 'syncErrors'))
       const pristine = value === initial
       return {
@@ -85,6 +86,7 @@ const createConnectedField = ({
         pristine,
         state: getIn(formState, `fields.${name}`),
         submitError: getIn(formState, `submitErrors.${name}`),
+        submitting,
         syncError,
         value,
         _value: ownProps.value // save value passed in (for checkboxes)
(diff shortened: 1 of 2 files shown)


erikras:
Where does it end up, though? It gets passed to the component? Should it not be under `meta`?

erikras:
There are a few things wrong with this, like using `getSyncError()`, but I'm going to merge and repair them on the main branch.

clayne11:
Alright. Good point about where the props should go. I didn't fully think it through. Thanks for merging this in! Looking forward to the next release. Should be getting close to a stable release I imagine.

erikras:

brotzky:
When is submit state updating?

Before I was passing a `submitting` prop to my button to display a loading animation for async calls. Now the `submitting` prop on the button is being stripped and there is only `meta.submitting`, but `meta.submitting` doesn't seem to be updating for asynchronous calls.

Redux Actions:
`FORM_SUBMIT_REQUEST` ==> submitting: true (show animation)
`FORM_SUBMIT_SUCCESS` ==> submitting: false (hide animation)

In the mean time I've renamed my prop to `isSubmitting` to avoid the conflict with `meta.submitting`

Issue #1511, opened 2016-08-10
[v6-v4] Unable to tell which async field you are validating 


KaboomFox:
In my asyncValidate function, I want to know what field I am validating. For instance if I am validating that the username is available  I want to call an API to validate to just validate the username. However if I am checking if the email is taken I want to call the API to check the email. Right now I cannot tell which field has been blurred. It would be nice if the asyncValidate function had blurredField parameter or if it was pass in with the props.

JMacLulich:
@KaboomFox yes this would be useful.

Asynchronous operations are typically long running by default, i.e. server side validation so if you have 3 or 4 of these per form and you run all of them even for fields that haven't changed that's a waste of time.

erikras:
Fix published in [`v6.0.0-rc.5`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-rc.5).

Issue #980, opened 2016-05-16
async validate multiple feilds


upq:
hello
I have 2 feilds that I want to do async validate for

```js
asyncBlurFields: [ 'username', 'email' ],
```

but I want each one to call a different function they both call `asyncValidate` [following this example](http://redux-form.com/5.2.3/#/examples/asynchronous-blur-validation?_k=7yebxo) 

I asked that question in the discord channel as suggested but I got nothing
however I then tried this hack:

```js
const asyncValidate = (values, dispatch , props) => {
  console.log(props.form);
  return new Promise((resolve, reject) => {
    if (props.form._active === 'username'){
      (async function() {
        const resp = await fetch('/graphql', {
          method: 'post',
          headers: {
            Accept: 'application/json',
            'Content-Type': 'application/json',
          },
          body: JSON.stringify({
            query: '{allUsers(username:"' + values.username + '"){username}}',
          }),
          credentials: 'include',
        });
        const { data } = await resp.json();
        if (data.allUsers.length > 0) {
          reject({ username: 'That username is taken' });
        }else {
          resolve()
        }
      }());
    } else if (props.form._active === 'email') {
      (async function() {
        const resp = await fetch('/graphql', {
          method: 'post',
          headers: {
            Accept: 'application/json',
            'Content-Type': 'application/json',
          },
          body: JSON.stringify({
            query: '{allUsers(email:"' + values.email + '"){email}}',
          }),
          credentials: 'include',
        });
        const { data } = await resp.json();
        console.log(data);
        if (data.allUsers.length > 0) {
          reject({ email: 'That email is taken' });
        }else {
          resolve()
        }
      }());
    }
  })
}
```

Everything is fine but when I reject the promise when email is taken the error for the username disappears and vice versa.

How can I do that elegantly without bugs?

sckoh:
+1 
I'm using asyncValidate for multiple fields too, username and tel number.
I'm using the same approach as yours, when my telNumber is in error state.
Then, when I blur username field and got a resolve promise for username field,
The error state of telNumber field is gone.

skiant:
In case you're still having this issue, it seems you need to reject the promise with every error.
In our case, I've used Lodash to extract all errors from the `props.form` object.

I push any error the validation might find in this `currentErrors` object and reject the promise with said object if it's not empty.

KaboomFox:
in V6 this will become an issue again because props.form will only return the current form name.

peteruithoven:
[...]

Issue #817, opened 2016-04-15
How do I know which field has blurred in the `asyncValidate` function?


techird:
I just can't find a proper way to do it. I don't want all my async fields being validated when there is just one field blurred.

erikras:
Hmm.. I don't think that's made available to you. It's available to your form, via the `asyncValidating` prop, but not to your async validation function. It was assumed that you would probably only be validating a few fields, and they might depend on other fields, so you'd need all the form values.

techird:
@erikras Yes, I agree with you. However,  in some scenario I need this, expecially I want to validate a field list in async way.

Coobaha:
@techird  as a workaround, you can get private field with asyncValidated from form reducer

techird:
@Coobaha That's a way out. And I am also asking for a better way that doesn't hacks.

chelzwa:
Perhaps asyncValidate could receive the field which was blurred in addition to all the form values? I can definitely work around this, but it would be very helpful to know which field was blurred when running async validators.

erikras:
> Perhaps asyncValidate could receive the field which was blurred in addition to all the form values? I can definitely work around this, but it would be very helpful to know which field was blurred when running async validators.

That seems very reasonable.

kristian-puccio:
don't we already have this?
```
function validate(values, dispatch, props) {
const triggeredBy = props.form._active;
}
```

erikras:
@kristian-puccio is correct. That is available, but it feels a little hacky.

chelzwa:
@erikras Honestly, although props.form._active definitely does feel weird, it also does work pretty well for my use case.

thomasmery:
Hi there,

as I am relying on the `props.form._active` hack to get which field just got blurred I'm facing a behaviour where `props.form._active` is `undefined` every second field that gets blurred

would someone tell me if they have seen this or if it rings a bell about something I might be doing wrong?

thnks

mjward:
I'm also getting `undefined` when accessing `props.form._active`

KaboomFox:
V6 doesn't have this private variable.

KaboomFox:
[...]

PR #1539, opened 2016-08-15
expose blurredField in asyncValidate

KaboomFox:
This addresses some pain points as outlined in #980 #817 #1511 and allows you to call different apis for different blurred fields.

### DIFF
--- a/docs/api/ReduxForm.md
+++ b/docs/api/ReduxForm.md
@@ -34,10 +34,10 @@ Defaults to `[]`.
 > See [Asynchronous Blur Validation Example](http://redux-form.com/6.0.0-rc.4/examples/asyncValidation/)
 for more details.
 
-#### `asyncValidate : (values:Object, dispatch:Function, props:Object) => Promise<undefined, errors:Object>` [optional]
+#### `asyncValidate : (values:Object, dispatch:Function, props:Object, blurredField:String) => Promise<undefined, errors:Object>` [optional]
 
-> a function that takes all the form values, the `dispatch` function, and the `props` given
-to your component, and 
+> a function that takes all the form values, the `dispatch` function, the `props` given
+to your component and the current blurred field, and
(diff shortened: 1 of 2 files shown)


peteruithoven:
But this isn't the whole solution? If I understand #980 it explains that the resulting error of the asyncValidate will override the existing errors, this means you can't run two validations at the same time. (Which can easily happen on a bad connection for example). Ideally we have a way to handle multiple async validation errors and a way to track the state of multiple async validations (to show spinners for example). 
That's why I [proposed](https://github.com/erikras/redux-form/issues/980#issuecomment-239815732) adding the validation properties to the Field components.

KaboomFox:
`const asyncValidation = (values, dispatch, props, currentField) => {
  let oldErrors = props.asyncErrors || {};
  if (currentField === 'username') {
    return dispatch(checkAvailableUsername(values.username)).then((newErrors) => {
      let combo = combineErrors(newErrors, oldErrors, currentField);
        if (combo) {
          reject(combo);
        } else {
          resolve();
        }
      });
    });
  }
  if (currentField === 'password') {
    return dispatch(checkPasswordQuality(values.password)).then((newErrors) => {
      let combo = combineErrors(newErrors, oldErrors, currentField);
        if (combo) {
          reject(combo);
        } else {
          resolve();
        }
      });
    });
  }
};

function combineErrors(newErrors, oldErrors, currentField) {
  let oE = oldErrors;
  delete oE[currentField];
  return assign(oE, newErrors);
}`

Here is an example to #980 issue

erikras:
This seems okay. It's a pretty minimal change. But @peteruithoven's per-field suggestion also has merit.

erikras:

aJoohongKim:
@KaboomFox Thanks for the example. I found out that I need return a default Promise when I submit the form.

return new Promise((resolve) => {
    resolve();
  });

Issue #1546, opened 2016-08-16
[v5] - Pass parent props into custom `handleSubmit()`


baronswindle:
I'd really like to be able to do something like this:

```
function handleSubmit(values, dispatch, props) {
  return () => {
    const { policyType } = props;
    return createUserWithPolicy({
      policyType,
      user: { ...values }
    })
      .catch(error => Promise.reject({ _error: error.message }));
  };
}
```

I reckon it's possible to do this by making `policyType` a form field and managing its value at the level where it would change using redux-form field properties, but it would be worlds easier to just pass parent props into the `handleSubmit()` function. It'd be pretty easy to do; I've already made the change on my personal fork that I'm using in a project. So if nobody has any objections, I'd happily make the PR.

Let me know what you guys think.

naw:
Have you considered generating your handleSubmit function with a factory that receives `policyType`?

```javascript
function handleSubmitFactory(policyType) {
  return (values) => {
    // createUser...
  }
}

<form onSubmit={handleSubmitFromReduxForm(handleSubmitFactory(props.policyType))}>
```

baronswindle:
@naw There definitely are other ways to do it, and your suggestion seems like a fine one. Still, I think this could be an improvement to `redux-form` and make the API more consistent. For example, `validate()` and `asyncValidate()` functions receive the component's props as an argument, so it seems reasonable that the `handleSubmit()` function should as well.

erikras:
Released @baronswindle's fix in [`v5.3.3`](https://github.com/erikras/redux-form/releases/tag/v5.3.3).

PR #1548, opened 2016-08-16
pass props to `handleSubmit()` function

baronswindle:
#1546


erikras:
Released as [`v5.3.3`](https://github.com/erikras/redux-form/releases/tag/v5.3.3).

PR #1556, opened 2016-08-17
Fixed remount-reinitialize bug

### LINKED ISSUE #1503: When destroyOnUnmount is set to false, values are getting reset
I have a form (using rc4) with destroyOnUnmount set to false. When my component switches to a modal that calls the change action creator, the form calls redux-form/INITIALIZE and I lose all my values on the form.  It used to work in previous versions of redux-form. Is this a bug?

erikras:
Fixes #1503.


erikras:

PR #1560, opened 2016-08-18
Fix sync form wide error

### LINKED ISSUE #1533: [v6-rc4] Form-wide error object not defined even when using _error key on synch validate
Hi Guys,

I've been going over and over the documentation and source trying to find what I'm doing wrong with this, but I have't had any luck.

I'm trying to return a `_error` object from my synchronous validation function so I can access it as the `error` property during my form render.

The documentation here: http://redux-form.com/6.0.0-rc.4/docs/api/Props.md/ (specifically for the item `error`) says I should assign a value to the key `_error`. I'm also following the example for submit validation here: http://redux-form.com/6.0.0-rc.4/examples/submitValidation/ which shows the use of the `_error` field when throwing a `SubmissionError `.

Except, I continue to get `undefined` as the value of `error`.

For example:

```javascript

function synchValidate(values) { 
  
   const errors = { _error: {} }
   errors._error = "This is an error" 

   return errors
}

function render() { 

  const { error } = this.props
  
  console.log(error)
}
```

I'm expecting that the string "This is an error" is output when the render function of the form is called but I'm not getting it.

I'm probably missing something obvious, any insights would be great.

If you need a more detailed example I'm happy to produce  

Thanks, Jason.

erikras:
Fixes #1533.

### DIFF
--- a/src/__tests__/reducer.updateSyncErrors.spec.js
+++ b/src/__tests__/reducer.updateSyncErrors.spec.js
@@ -29,6 +29,34 @@ const describeUpdateSyncErrors = (reducer, expect, { fromJS, setIn }) => () => {
         }))
   })
 
+  it('should update form-wide error', () => {
+    const state = reducer(fromJS({
+      foo: {
+        values: {
+          myField: 'value',
+          myOtherField: 'otherValue'
+        }
+      }
+    }), updateSyncErrors('foo', {
+      myField: 'myField error',
+      _error: 'form wide error'
+    }))
+    expect(state)
+      .toEqualMap(setIn(fromJS({
+        foo: {
+          values: {
+            myField: 'value',
+            myOtherField: 'otherValue'
+          },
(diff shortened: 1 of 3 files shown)


erikras:

JMacLulich:
@erikras I think this fix is still buggy.

I've been trying to use rc5 and this is the behaviour I'm seeing:

 - I now get the `error` object exposed when I set it via the `validate` function.

BUT

 - It is not removed from the store when the `validate` function no longer returns `_error` field.

The `error` field is still sitting in the store for the form and is being exposed as a property to the form; so I erroneously use it.

So, I still don't think I can still use this :(

I also can't use the `invalid` field because even when the `_error` object is set the `invalid` field is set to `false`.

JMacLulich:
Also, I'm not sure where to file this, this seemed like the best place for it but I can file a new issue for it.

Issue #1567, opened 2016-08-18
[6.0.0-rc.5] Unexpected behaviour when passing dispatch as prop to <Field />


brotzky:
I'm not able to pass dispatch directly into the `Field` component. It appears to be stripping any prop with the name `dispatch`. I'm not sure if this was on purpose, but the behaviour was surprising to me and not very intuitive.

A simplified version of what doesn't work:
```JavaScript
render (
  const fieldProps = {
    dispatch, // passing dispatch into Field, along with other props
    ...
  };

 return <Field name="phone" component={Phone} {...fieldProps} />;
)


-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_


export const Phone = props => {
  console.log(props.dispatch) // undefined
  // ^^^^^^^^ dispatch is being stripped from props passed into <Field />

  return (
      <input
        {...props.input}
        type="tel"
        name="tel" />
  );
};
```

How I had to fix it:

```JavaScript
  const fieldProps = {
    reduxDispatch: dispatch,
   //  ^^^^^ renaming dispatch to something else so it doesn`t get stripped
    ...
  };
````

erikras:
Ha! Yeah, I guess Field should just go ahead and pass `dispatch` to all of its components. Might as well.

erikras:
Closing this. Addressed by #1569. You can use your workaround for now.

brotzky:
Thanks @erikras, ended up going back to rc4 :) thanks for the updates.

erikras:
Published in [`v6.0.1`](https://github.com/erikras/redux-form/releases/tag/v6.0.1).

PR #1569, opened 2016-08-19
Added dispatch to props for Field

erikras:
Addresses #1567.


erikras:

Issue #1579, opened 2016-08-22
[v6 rc5] form-specific change action is not bound anymore


mihhailk:
I have quite common use-case: some fields are updating on another field change. As I have understood, in order to achieve that we should use change action creator that is injected as prop to form component. 
I got it working in rc4 version, but after updating to rc5 it seems that change is not a prop anymore.
[v6 rc4] working code:

```
Form.propTypes = {
  change: PropTypes.func.isRequired,
  dispatch: PropTypes.func.isRequired
}
```

then in Form component: 
```
updateField(field, value) {
  const {dispatch, change} = this.props;
  dispatch(change(field, value));
}
```

The same code in rc5 results in an warning:
`Failed prop type: Required prop `change` was not specified in Form`

Sidenote: docs say that change should already be bound to dispatch (if I understand "actions [...] are already bound to dispatch" correctly), but it seems that in rc4 one should still dispatch it manually. Please correct me if I am wrong

erikras:
This did change as part of the introduction of the `Fields` component. I will investigate. 🔍

andrepcg:
I was too manually changing the value of some fields and it was very useful in specific situations.

erikras:
Fixed in #1588. Will post again here when it has been released. Thanks.

erikras:
Published in [`v6.0.1`](https://github.com/erikras/redux-form/releases/tag/v6.0.1).

PR #1588, opened 2016-08-23
Replaced blur and change bound actions as form props

erikras:
No reason _not_ to provide these.

Addresses #1579.


erikras:

PR #1591, opened 2016-08-23
Fixed immutable array reducer bug

### LINKED ISSUE #1571: input field loses focus on first form input character
redux-form v6.0.0-rc.5 but your examples are working perfect. created a separate app with creact-react-app but still acted the same

### LINKED ISSUE #1574: [v6] Redux-Form doesn't play well with ImmutableJS when using FieldArray
I've created a full example with information about this problem here: [redux-form-failing-cases/immutable-array/](https://github.com/Reanmachine/redux-form-failing-cases/tree/master/immutable-array)

Here's a brief of the problem:
----

It seems that `form.fields.<name>` is being accidentally created as a `Map` instead of a `List` with relation to the `onBlur` or `onFocus` calls trying to set `touched` for an `FieldArray` when using ImmutableJS

*Additionally, the `value` for a field in a `FieldArray` field is defaulting to `Map()` which serializes out to the input as `"Map{}"`.*

Here's the description copied from the repository:
----

**Disclaimer:** I am new to `redux-form` and may very well be doing something incorrect, but I've tried
to follow the information available in the most recent documentation & some of the more important tickets
related to the issue:

* [[v6] Can't pass custom props to Field component #1397](https://github.com/erikras/redux-form/issues/1397)
* [Poll: Field component props structure #1425](https://github.com/erikras/redux-form/issues/1425)

I made this repository mostly because it was easier to illustrate the problem as I'm not intimately
familliar with the `redux-form` codebase.

**I suspect the problem is related to [this line](https://github.com/erikras/redux-form/blob/v6/src/reducer.js#L39) 
in `reducer.js`**. I believe that calling `setIn` here may be the problem as the object at `${key}` (in this case 
`fields.things` doesn't exist, so when it creates `things` it does what is described in the 
[immutableJS documentation](https://facebook.github.io/immutable-js/docs/#/Map/setIn):

> Returns a new Map having set value at this keyPath. If any keys in keyPath do not 
> exist, **a new immutable Map will be created** at that key.

This seems to be triggered when the `onFocus` and `onBlur` get fired as they need this object to
specify if it was `touched` or not.

erikras:
Allowed pushing, unshifting, and inserting of `undefined` into a field array. Previously, after a `push()`, the value passed to your input would be `{}` instead of `''`.

Made arrays be `Immutable.List` when using `redux-form/immutable`. Fixes #1574.

### DIFF
--- a/src/__tests__/FieldArray.spec.js
+++ b/src/__tests__/FieldArray.spec.js
@@ -777,45 +777,461 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
       // field array NOT rerendered
       expect(renderFieldArray.calls.length).toBe(1)
     })
-  })
 
-  it('should work with Fields', () => {
-    const store = makeStore({
-      testForm: {
-        values: {
-          foo: [ 'firstValue', 'secondValue' ]
+    it('should create a list in the store on push(undefined)', () => {
+      const store = makeStore({})
+      const renderField = createSpy(props => <input {...props.input}/>).andCallThrough()
+      const renderFieldArray =
+        createSpy(({ fields }) => (<div>
+          {fields.map(field => <Field name={field} component={renderField} key={field}/>)}
+          <button onClick={() => fields.push()}>Add Dog</button>
+        </div>)).andCallThrough()
+      class Form extends Component {
+        render() {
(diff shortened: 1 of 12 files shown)


erikras:

Issue #1341, opened 2016-07-13
[V6] Field defaultValue not initialized in Store, nor passed to onSubmit


ebwill:
I would think that at the very least field defaults should be passed to the `onSubmit` handler to reflect the actual values of the fields. If not, it places an unnecessary burden on the `onSubmit` handler to be aware of the default values, in the case that they are needed (not every API allows sparse updates).

ebwill:
To clarify, I am seeing this behavior on v6 rc.2. Not sure if it exists in v5.

olslash:
+1 @erikras I'm trying to figure out a good way around this issue -- right now the only thing defaultValue does is overwrite `props.input.value` in the case that no value is given (defaulting to empty string)

I'd like to see default values be more integrated into the data flow. They should be in the store just like if they were set by onChange, that way we don't have to explicitly check and re-populate the store state with them in a submit handler.

staskorz:
+1. Not only it's important for onSubmit(), but also for validate().

alyrik:
+1. It's become initialized in Store only after touching and blurring input.

ktaras:
+1. Have the same on v6-rc.5. Any thoughts how to make a hotfix?

erikras:
Thanks guys. Will investigate. 🔍

msaglietto:
Same is happening to me on rc.4

This is not really is going to apply to all (because I did it just for one field) but what I did as a workaround is on componentWillReceiveProps check for the initialValues of that field and if is not set dispatch a change with the default value .. 
```
componentWillReceiveProps(nextProps) {
  if (!nextProps.initialValues || !nextProps.initialValues.state) {
      this.props.dispatch(this.props.change('state', 'inactive'));
    }
}
```

If you are on rc.5 it seams initialValues are not longer on the props so you have to map the field to a prop or something like that

erikras:
> If you are on rc.5 it seams initialValues are not longer on the props so you have to map the field to a prop or something like that

@msaglietto, are you referring to #1570?

msaglietto:
@erikras yes but no,  because I am not using mapStateToProps to get initialValues on rc.4 ... I did not tried to get initialValues using mapStateToProps on rc.5 ..

What I mean on rc.5 it may be something like mapStateToProps of 'state' .. to know if it was set .. and replace nextProps.initialValues.state by nextProps.state

erikras:
[...]

PR #1593, opened 2016-08-23
Nixed defaultValue 👻

erikras:
RIP `defaultValue`. It was causing too much confusion. Addresses #1341.

### DIFF
--- a/docs/api/Field.md
+++ b/docs/api/Field.md
@@ -36,12 +36,6 @@ be as simple as `'firstName'` or as complicated as
 A `Component`, stateless function, or string corresponding to a default JSX element.
 See the [Usage](#usage) section below for details.
 
-#### `defaultValue : any` [optional]
-
-A value to be used if the current value for this field is `undefined` in the Redux Store.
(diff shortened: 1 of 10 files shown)


erikras:

olslash:
I found that the old behavior was useful in a couple places, so i made a little decorator to imitate it (just this morning, not battle tested 😄 ). Hope this helps someone!

```javascript
export const withDefaultValue = (val) => (WrappedComponent) => {
  return (props) => {
    const defaultProps = {
      ...props,
      input: {
        ...props.input,
        value: props.input.value === '' ? val : props.input.value
      }
    };

    return <WrappedComponent { ...defaultProps } />;
  };
};
```

then 

```javascript
@withDefaultValue
class MyCustomInput extends Component {
  render() {
    return <input ...etc >
  }
}

class MyCustomInputField extends Component {
  render() {
    return <Field component={ MyCustomInput } { ...this.props } ...etc />
  }
}
```

erikras:
@olslash That's a perfect example of why I ❤️ javascript/React/functional programming.

PR #1594, opened 2016-08-23
Fixed bug where FieldArray would not rerender on add/remove

### LINKED ISSUE #1592: [v6] FieldArray not re-rendering after add/remove
The recent commit to fix the FieldArray re-rendering bug (b4fc32dade0c011c79487e0fb0f0c8775e5f6059) had the side effect that the FieldArray doesn't re-render after adding/removing fields

erikras:
Fixes #1592.


erikras:

PR #1598, opened 2016-08-23
Passthrough the initialValues prop to the enhanced component.

### LINKED ISSUE #1570: [v6rc5] initialValues are not propagated
```
@connect(mapStateToProps)
@reduxForm(...)
export default class extends Component {
  render() {
    console.log(this.props.initialValues) <-- undefined
    ...
  }
}

function mapStateToProps(state){
  return { initialValues: { ...} }
}
```
However in v6rc4 worked fine

ruiaraujo:
Fixes #1570.

### DIFF
--- a/docs/api/Props.md
+++ b/docs/api/Props.md
@@ -167,6 +167,10 @@ that as if it were the error for a field called `_error`, and it will be given a
 determined by comparing the current data with these initialized values.
 This is a bound action creator, so it returns nothing.
 
+#### `initialValues : Object`
+
+> The same initialValues object passed to `reduxForm` to initialize the form data.
+
 #### `invalid : boolean`
 
 > `true` if the form has validation errors. Opposite of `valid`.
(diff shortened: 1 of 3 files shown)


erikras:
With docs update and everything! ❤️!

In my `v6` issue fixing rampage this morning, I was subconsciously skipping  #1570 because I didn't understand just how simple it was. Thanks for this.

ruiaraujo:
Thanks for redux-form and for merging!

erikras:

oomathias:
I'm using redux-form 6.5.0, and on first render initialValues prop is undefined.
I should say, that initialValues are populated from the state.
Even inside `render() or componentDidMount()` `this.props.initialValues === undefined`, but the form is correctly initialised with the values.
If I trigger another render, values are correctly populated.

I need to access these initialValues on first render, to generate some `<options />`

Should I open another issue?

Issue #1603, opened 2016-08-23
[v6-rc5] Sync validation does not trigger redux-form/UPDATE_SYNC_ERRORS when validation passes


ntucker:
I'm using the form wide _errors. When validation fails, it works fine, but when the sync errors object no longer has contents, the UPDATE_SYNC_ERRORS is not called, so the form still has errors prop with outdated errors.

erikras:
Very well explained, thank you. Will investigate. 🔍

erikras:
Hmm.. This very simple example is working for me.

![screencast 2016-08-24 09-39-01](https://cloud.githubusercontent.com/assets/4396759/17922333/bee442d0-69de-11e6-96e6-51f1eab610a1.gif)

```jsx
import React from 'react'
import { Field, reduxForm } from 'redux-form'

const validate = values => {
  const errors = {}
  if (!values.username) {
    errors._error = 'Missing username'
  }
  return errors
}

const SyncValidationForm = (props) => {
  return (
    <form>
      <div>
        <Field name="username" component="input"/>
      </div>
      <div>Error: {props.error || 'None'}</div>
    </form>
  )
}

export default reduxForm({
  form: 'syncValidation',  // a unique identifier for this form
  validate                 // <--- validation function given to redux-form
})(SyncValidationForm)
```

ntucker:
Maybe this has to do with immutable?

```javascript
import { Field, reduxForm, SubmissionError } from 'redux-form/immutable'

function validateOrder(values, props) {
  const { account, position } = props
  const errors = {}

  const cost = parseFloat(values.get('quantity')) * parseFloat(values.get('price'))
  const fundsAvailable = account ? parseFloat(account.buying_power) : 0
  if (fundsAvailable < cost) {
    errors._error = 'You don\'t have enough funds!'
  }
  return errors
}

// in the component Order
render() {
  return (
    <div>
      {error && <div className="error">{error}</div>}
      ..
    </div>
  )
}

//the wrapper
reduxForm({ form: 'discovery/order', validate: validateOrder })(Order)
```

JMacLulich:
@erikras I'm actually seeing the same issue.

I reported this issue here: 

but I guess that was the wrong forum.

https://github.com/erikras/redux-form/pull/1560#issuecomment-241392573

Basically, I can set _error but I can't clear it when I no longer return it from synch validation.

erikras:
[...]

PR #1606, opened 2016-08-24
Form-wide _error wasn't flipping valid flag

erikras:
May be related to #1603, which is what I was looking into when I discovered this bug.


erikras:

Issue #1614, opened 2016-08-24
v6.0.0-rc.5 Autofill or alternative


itaylorweb:
Hi,

I have an order form where the user can add multiple products (FieldArray).
Each product added has a SelectField (product) and 2 TextFields. (qty, price).
When the user selects a product I need the price to auto populate with the default value from the product.

How do I handle this in v6.0.0-rc.5

```jsx
import React, { Component } from 'react';
import RaisedButton from 'material-ui/RaisedButton';
import FlatButton from 'material-ui/FlatButton';
import MenuItem from 'material-ui/MenuItem';
import ActionDelete from 'material-ui/svg-icons/action/delete';
import ContentAdd from 'material-ui/svg-icons/content/add';
import IconButton from 'material-ui/IconButton';
import Table from 'material-ui/Table/Table';
import TableHeaderColumn from 'material-ui/Table/TableHeaderColumn';
import TableRow from 'material-ui/Table/TableRow';
import TableHeader from 'material-ui/Table/TableHeader';
import TableRowColumn from 'material-ui/Table/TableRowColumn';
import TableBody from 'material-ui/Table/TableBody';
import Dialog from 'material-ui/Dialog';
import FloatingActionButton from 'material-ui/FloatingActionButton';

import { reduxForm, Field, FieldArray, formValueSelector } from 'redux-form';
import { connect } from 'react-redux';

import {
  SelectField,
  TextField
} from 'redux-form-material-ui';

import _ from 'lodash';

import Loading from '../General/loading';

import { fetchCustomers } from '../../actions/customer';
import { fetchProducts } from '../../actions/product';
import { updateQuote, fetchQuote } from '../../actions/quote';

class EditQuoteForm extends Component {
  constructor() {
    super();
    this.renderProducts = this.renderProducts.bind(this);
    this.handleCancel = this.handleCancel.bind(this);
    this.handleCloseError = this.handleCloseError.bind(this);
    this.handleCloseSuccess = this.handleCloseSuccess.bind(this);
    this.onSubmit = this.onSubmit.bind(this);
    this.renderCustomers = this.renderCustomers.bind(this);
    this.renderQuoteProducts = this.renderQuoteProducts.bind(this);
    this.renderSummary = this.renderSummary.bind(this);
  }

  static contextTypes = {
    router: React.PropTypes.object
  };

  componentWillMount() {
    this.props.dispatch(fetchCustomers());
    this.props.dispatch(fetchProducts());
[...]

PR #1620, opened 2016-08-25
Implemented autofill feature from v5

erikras:
Now _THIS_ is really the last outstanding feature that `v6` needed to do everything that `v5` did.

#1614


erikras:

PR #1652, opened 2016-08-30
Remove deprecated reducer.normalize details from Reducer.md in documentation

kevinzwh:
Remove old details from V6 documentation about the normalizer function which no longer exists in reducer.js

The details also included a broken link to the old ReducerNormalize.md which no longer exists as well. Since this information is deprecated now and there is a new way to normalize form values ([normalizing functions](http://redux-form.com/6.0.1/examples/normalizing/)), it's worth removing from the docs now.


erikras:
Thanks! 👍

PR #1654, opened 2016-08-30
fix submit form after failing

### LINKED ISSUE #1641: Can only submit once?
# Problem Description

I'm doing async submit validation as described [in the doc](http://redux-form.com/6.0.1/examples/submitValidation/).

If my `onSubmit` rejects with an error, I cannot submit the form again.

It also does it in the official doc!

I input "John":

image

Then add the correct password ("redux-form") and get the same error:

image

Whereas if I input "John" "redux-form" after refreshing the page:

image

Is this expected? That's kinda weird...

# Workaround

So I ended up doing:

```javascript
reduxForm({
    form: 'login',
    destroyOnUnmount: false,
    validate: () => ({}),
})
```

because `handleSubmit` does not run submit if the form is not `valid`:

```javascript
// lib/handleSubmit.js
if (valid) {
    var doSubmit = function doSubmit() {
      var result = void 0;
      try {
        result = submit(values, dispatch);
      } catch (submitError) {
        var error = submitError instanceof _SubmissionError2.default ? submitError.errors : undefined;
        setSubmitFailed.apply(undefined, _toConsumableArray(fields));
        if (onSubmitFail) {
          onSubmitFail(error, dispatch);
        }
        return error;
      }
  // etc.
}
```
My interpretation of the issue is that the doc is wrong when it says that the default value of the `validate` function is `() => ({})`:

image

I really like this library and I'd be glad to submit a fix if you're interested and my interpretation of the issue is correct!

In my opinion just having the correct default value for `validate` should be enough.

ortex:
Closed #1641.

### DIFF
--- a/src/__tests__/reducer.change.spec.js
+++ b/src/__tests__/reducer.change.spec.js
@@ -207,8 +207,7 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
         foo: {
           values: {
             myField: 'different'
-          },
-          error: 'some global error' // unaffected by CHANGE
+          }
         }
       })
   })
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -2298,10 +2298,10 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
 
       input.calls[0].arguments[0].input.onChange('bar')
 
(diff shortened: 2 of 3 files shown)


erikras:
Nice work. 👍

erikras:

PR #1667, opened 2016-08-31
Improved performance of getIn()

smirea:
`getIn` seems to be the highest low hanging fruit when it comes to performance

Optimized for the most common use-case: path length of 1, 2 or 3 items

On my chrome it shows ~5x improvement

jsPerf is down, so here's a benchmarkJS fiddle:
https://jsfiddle.net/9vqm86eu/1/

### DIFF
--- a/src/structure/plain/__tests__/getIn.spec.js
+++ b/src/structure/plain/__tests__/getIn.spec.js
@@ -33,6 +33,31 @@ describe('structure.plain.getIn', () => {
     expect(getIn(state, 'foo.bar[1].dog')).toBe(42)
   })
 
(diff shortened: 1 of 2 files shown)


erikras:
> highest low hanging fruit

😆

Good thinking. You've added a lot of branches that the test coverage monster is angry about. It's tedious, but could you...

```bash
$ npm run test:cov
$ open coverage/lcov-report/index.html
```

...and add the 7 or 8 test cases to test each branch?

smirea:
good point, added

erikras:
Excellent work, @smirea! Thanks!

See if you can spot any more fruit. :-)

erikras:

PR #1671, opened 2016-08-31
Fix build:es output by replacing babel-preset es2015 with es2015-no-commonjs

NotWoods:
Currently the jsnext:main build of redux-form contains CommonJS code instead of ES module code.  
Removing the `transform-es2015-modules-commonjs` babel plugin fixes the issue and outputs the correct format.


erikras:

PR #1673, opened 2016-08-31
isPristine does not differentiate between null and 0

alexhayes:
I have a situation where a number field gets set to `0` with an initial value of `null` and the form and field don't get marked as dirty.

Note that because it's a number field `0` in this case is of type `number` not `string` which `isPristine` doesn't seem to accommodate for.

In the case of `'0'` (ie.. string) `isPristine` behaves as expected already.

If this is the desired behaviour are there any suggestions for a work around?


erikras:

Issue #1648, opened 2016-08-29
Update initialValues example documentation


Sir4ur0n:
In [initializeFromState example](http://redux-form.com/6.0.1/examples/initializeFromState/) there is no mention of `enableReinitialize` and it says the only mean of reloading `initialValues` is to emit an `INITIALIZE` redux-form action.

I think it wouldn't hurt to update this example to mention the `enableReinitialize` prop, together with an example, considering the number of issues open about this topic.

JakeLin1024:
Haha, it would be great too to add an example that shows how ```enableReinitialize``` works with ```keepDirtyOnReinitialize````

kevinzwh:
Just opened a PR to address this! Hopefully adding this mention will help

NeXTs:
Does anybody has an idea what's going on at [initializeFromState example](http://redux-form.com/6.0.2/examples/initializeFromState/)?

How to write it without duck? How it would look with common separated redux actions/reducers?
What means name of this action? `redux-form-examples/account/LOAD`
`redux-form-examples` does this part affects something internal in redux-form or it may be whichever? 
`account` once again, it belongs to internal redux-form values for exactly this form(what is unlikely because form name is 'initializeFromState') or it's reducer name where redux-form will take initial values for this form?

klk1010:
I'm in the use `const` in ES6 whenever possible camp (that debate is a separate topic, for further reading see this [article](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75?swoff=true#.ooez1povi) ). The [documentation](http://redux-form.com/6.1.1/examples/initializeFromState/) for setting the initial state of a form uses `let` instead of `const` for the react component `InitializeFromStateForm`, which just made me feel dirty. As an alternative, I used `compose` to combine `redux-form` and `connect`, which allowed me to use `const` with `connect`, and `redux-form` to initialize my forms from data which is already in my store. I also used a custom input component. In the full code, I broke `renderInput` out into its own component for ease of reuse, but included it here for simplicity. For anyone that may share my sentiment: 
```
import React from 'react';
import { connect } from 'react-redux';
import { compose } from 'redux';
import { Field, reduxForm } from 'redux-form';

const renderInput = ({ input, meta: { touched, error }, ...custom }) => (
  <MyTextInput
    errorText={touched && error}
    {...input}
    {...custom}
  />
);

const MyForm = ( props ) => {
  const { handleSubmit } = props;
  return (
[...]

PR #1681, opened 2016-09-02
Add missing documentation on `enableReinitialize` in InitializeFromState example docs

kevinzwh:
This issue is mentioned in https://github.com/erikras/redux-form/issues/1648.

Currently, the example documentation refers to the method of dispatching `INITIALIZE` as if it is the only method to reinitialize the form with new pristine values. `enableReinitialize` & `keepDirtyOnReinitialize` are missing from this section.

This change adds documentation about `enableReinitialize` & `keepDirtyOnReinitialize` to the Initialize From State example as another option.

### DIFF
--- a/examples/initializeFromState/src/InitializeFromState.md
+++ b/examples/initializeFromState/src/InitializeFromState.md
@@ -10,9 +10,11 @@ reducer. To get those values into your `redux-form`-decorated component, you wil
 `connect()` to the Redux state yourself and map from the data reducer to the `initialValues`
 prop.
 
-You may only initialize a form component _once_ via `initialValues`. If you wish to change the 
-"pristine" values again, you will need to dispatch the `INITIALIZE` action yourself (using the 
-action creator provided by `redux-form`).
+By default, you may only initialize a form component _once_ via `initialValues`. You can also
+provide a `enableReinitialize` prop set to `true` to allow the form the reinitialize every time
+the `initialValues` prop changes. If you wish to change the "pristine" values again, you will 
+need to dispatch the `INITIALIZE` action yourself (using the action creator provided by 
+`redux-form`).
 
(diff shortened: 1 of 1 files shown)


skvale:
I think I misunderstood the intent of `keepDirtyOnReinitialize`.
from your unit tests, a dirty state of 
```javascript
   deep: {
     foo: "bar"
  }
```
will not be written over by re-initializing with
```javascript
   deep: {
     foo: "baz"
  }
```
But it appears that if I start with 
```javascript
   deep: {
     foo: "bar"
  }
```
and re-initialize with 
```javascript
   aNewField: {
     something: "value"
  }
```
the form state of `deep: { foo: "bar" }` has been cleared.  Is there a way to re-initialize state without wiping out preexisting values?  If this isn't clear I can make a code-pen to illustrate it.

kevinzwh:
@skvale 
When `keepDirtyOnReinitialize` and `enableReinitialize` are set, the form can reinitialize with new "pristine" values, but won't write over **dirty** values in the form that the user has inputted.

Given your description, `deep: { foo: "bar" }` shouldn't be cleared unless you changed initialValues to an object without `deep: { foo: "bar" }`

For example, if you pass your form component with initialValues:
```es6
let initialValues={ deep: { foo: "bar" } }

<ReduxFormExampleComponent
  initialValues={ initialValues }
/>
```

Then change `initialValues` as you described:
```es6
  initialValues={ aNewField: { something: "value" } }
```

This change will cause the form to **lose** `deep: { foo: "bar" }` because you have not restated them in the new initialState and thus the new pristine value will be an empty field

To retain `deep: { foo: "bar" }` after reinitialization, the new initialValues should look like this:
```es6
  initialValues={ 
    aNewField: { something: "value" },
    deep: { foo: "bar" }
  }
```

PR #1684, opened 2016-09-02
Implemented custom meta props. Fixes #1634

### LINKED ISSUE #1634: [v6] Support custom meta props?
Use-Case: apart from the existing flags (dirty, touched, ...) i'd like to implement my own custom state `highlighted` to globally highlight fields based on interaction. 

I've tried using a separate `highlighted: {}` state in the form, but mimicking the tree anywhere else is clumsy. Changing the value of the input is also not ideal (aka from `foo: String` to `foo: {value: String, highlighted: Bool}`) since that requires changing the signature of the model, which, more often than not, is based on a database object

Ideally the interaction would be:

```
combineReducers: {
    form: form.plugin((state, action) => {
        if (action.type !== 'highlight') return state;
            return {
                    ...state,
                    fields: {
                        ...state.fields,
                        [action.meta.field]: {
                            ...state.fields[action.meta.field],
                            highlighted: action.value,
                        },
                    },
                };
    })
}
```

and in my component:

```
<Field name='foo' component={render} />

render = ({input, meta: {highlighted}}) => 
     <div style={highlighted ? 'fancy' : null}><input {...input} /></div>
```

but that's not available since the meta props are hand-picked and hardcoded.

**TL; DR:** what do you think of adding `...state` on line https://github.com/erikras/redux-form/blob/master/src/createFieldProps.js#L68 to enable user defined meta props

smirea:
Ref: https://github.com/erikras/redux-form/issues/1634

This allow creating own meta properties for fields via the reducer:

```js
combineReducers({
    form: form.plugin((state, action) => {
        if (action.type !== 'redux-form/CHANGE') return state;
        return setIn(state, 'fields.' + action.meta.field + '.myCustomProp', true);
    })
});
```

results are passed to the `meta` object:

```jsx
<Field name='foo' component={renderFancy} />

const renderFancy = ({input, meta: {myCustomProp}}) => 
    <input type="text" {...input} style={myCustomProp ? {color: 'red'} : null} />
```

### DIFF
--- a/src/__tests__/createFieldProps.spec.js
+++ b/src/__tests__/createFieldProps.spec.js
@@ -107,6 +107,35 @@ const describeCreateFieldProps = (name, structure, expect) => {
       expect(submittingResult.meta.submitting).toBe(true)
     })
 
+    it('should pass along all custom state props', () => {
+      const pristineResult = createFieldProps(getIn, 'foo', {
+        value: 'bar'
+      })
+      expect(pristineResult.meta.customProp).toBe(undefined)
+      const customResult = createFieldProps(getIn, 'foo', {
+        value: 'bar',
+        state: {
+          customProp: 'my-custom-prop'
+        }
+      })
+      expect(customResult.meta.customProp).toBe('my-custom-prop')
+    })
+
+    it('should not override canonical props with custom props', () => {
+      const pristineResult = createFieldProps(getIn, 'foo', {
+        value: 'bar'
+      })
+      expect(pristineResult.meta.customProp).toBe(undefined)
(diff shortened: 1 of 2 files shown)


erikras:
I wonder if this should be documented somewhere. It seems like a pretty obscure "feature".

Good job implementing it, though!

erikras:

marcselman:
In the current redux-form version you can only write plugin for specific forms.
I use dynamic form names so I don't see how I can add a plugin to that.
Any other way I can access the main redux-form reducer or add custom meta properties any other way?

PR #1689, opened 2016-09-03
Fix broken npm example script + add scripts & documentation for running each example locally

kevinzwh:
I found that there was a broken npm script inside of package.json:
`"example:simple": "node examples/simple/server.js"`

This script doesn't do anything currently because server.js no longer exists in this version and it does not tell npm to install dependencies for that example folder.

To fix this, I went ahead and added a programmatic example script that can be run for any example directory:
`"example": "npm --prefix ./examples/$form install && npm --prefix ./examples/$form start"`

Then reused this script to create shorthand scripts for each example directory, such as:
`"example:simple": "form=simple npm run example`
(You can still manually run "example" by doing `form=simple npm run example`, but this doesn't feel very intuitive. Also by explicitly writing out these commands in package.json, it will help somebody who might be looking for such a command to run a specific example.)

With these new npm example scripts, I updated the documentation for this in examples/README.md with more detail on how to get examples run locally.

In addition, a section on how to run each specific example locally is added to each individual example page to help anybody who might need direction on how to run it locally, since there was no instruction on those pages previously.

**Summary of changes proposed:**

- Remove old broken example:simple script

- Add new working npm example scripts for each example in package.json

- Update examples/README.md with more detailed instructions on how to run the examples locally

- Add a new section to each example page markdown file with a snippet on how to run that specific example locally.

- Small style change: unified sub-header usage to only H2 on example markdown files, since before they were a mix between H3 & H2, without too much difference in emphasis between sections.

Let me know what you think @erikras !


erikras:
Wow, I have no idea where that `example:simple` entry came from. I always just `cd` into the example dirs manually to run them. But if you've tested this and it's working, it seems okay to me.

Issue #1688, opened 2016-09-03
Bug with SET_SUBMIT_SUCCEEDED and startSubmit/stopSubmit?


taylorhakes:
I have code essentially that does this
```js
MyComponent = ReduxForm.reduxForm({
  form: 'simple',
  onSubmit(data, dispatch) {
    dispatch(ReduxForm.startSubmit('simple'));
    setTimeout(() => dispatch(ReduxForm.stopSubmit('simple')), 2000)
  }
})(MyComponent);
```

I know that you can return a Promise from onSubmit, but I am not using Promises. This type of code is even documented here http://redux-form.com/6.0.2/docs/faq/WebsocketSubmit.md/.

The problem is that `submitting` is not true for the 2 seconds as it should be. It is always `false` because this line https://github.com/erikras/redux-form/blob/master/src/reducer.js#L301 causes submitting to be set to `false` as soon as `onSubmit` finishes. That line of code appears to be unnecessary (for sync onSubmit) because this is the line that calls it https://github.com/erikras/redux-form/blob/e6c5d79482e1954d9fc360a95ba8811e32a37809/src/handleSubmit.js#L45 . It sets submitting to `false`, but it was never set to `true`.

I think the correct fix is to remove that line https://github.com/erikras/redux-form/blob/master/src/reducer.js#L301. It is not necessary for sync onSubmit and the Promise onSubmit already has this line https://github.com/erikras/redux-form/blob/e6c5d79482e1954d9fc360a95ba8811e32a37809/src/handleSubmit.js#L29 I am willing to create a pull request if you agree.

Here is a jsbin reproducing the issue. If you uncomment to commented out line and comment the first dispatch, you will see it works correctly.

http://jsbin.com/jonipafida/edit?js,output

erikras:
Fix published in [`v6.0.3`](https://github.com/erikras/redux-form/releases/tag/v6.0.3).

PR #1692, opened 2016-09-03
Changed setSubmitSucceeded to not touch submitting state

taylorhakes:
This fixes https://github.com/erikras/redux-form/issues/1688. I added a test case that was failing previously for sync onSubmit and added a test case for Promise onSubmit.


erikras:

PR #1706, opened 2016-09-06
isPristine now differentiates between null, 0 and NaN 

alexhayes:
I realise I just submitted #1673 for something very similar just a few days ago but I've found another edge case.

This PR addresses issues where a HTML5 number field is left empty which is fed through as `NaN`.

I could have modified the condition on line 24 of isPristine.js but I felt that it would be easier to read as is.

### DIFF
--- a/src/__tests__/isPristine.spec.js
+++ b/src/__tests__/isPristine.spec.js
@@ -76,4 +76,16 @@ describe('isPristine', () => {
     tryBothWays(null, 0, false);
   });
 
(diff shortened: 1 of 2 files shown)


alexhayes:
Note I've now also added a check as follows;

```javascript
...
} else if (Number.isNaN(initial) && Number.isNaN(data)) {
    return true;
...

alexhayes:
I'm going to resubmit this as another PR with further changes in the near future.

PR #1715, opened 2016-09-07
Add meta.touched to FieldArrays

AsaAyers:
With this PR, if you view the `fieldArrays` example and submit before doing anything, it will display the error for not adding any members.

selection_037

I can't fix the CSS for this because even when running a local example it links out to http://redux-form.com/6.0.2/bundle.css. 

In fixing this it exposed another bug. If you add any members, then you remove all members you can't get the `.touched` property, so the error never shows up again. Once you add a member `form.fieldArrays.fields.members` is converted from an object to an array. Later when the `TOUCH` action happens it can't add the `.touched` property because it's still an array even after the last member is gone. I spent about an hour tracking this down and trying to figure out how to fix it, but I don't know if I can fix it without potentially breaking other things. 

What do you think the right approach is for fixing that? Some things that seem like the could work:

1. `ARRAY_REMOVE` deletes `form.fieldArrays.fields.members` when the last member is removed.
2. `ARRAY_REMOVE` converts `form.fieldArrays.fields.members` to an object when the last member is removed.
3. `TOUCH` should be smarter and check for the empty array and convert it to `{ touched: true }` when it encounters it.

### DIFF
--- a/examples/fieldArrays/src/FieldArraysForm.js
+++ b/examples/fieldArrays/src/FieldArraysForm.js
@@ -12,10 +12,11 @@ const renderField = ({ input, label, type, meta: { touched, error } }) => (
   </div>
 )
 
-const renderMembers = ({ fields }) => (
+const renderMembers = ({ fields, meta: { touched, error } }) => (
   <ul>
     <li>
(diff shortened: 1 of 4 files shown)


erikras:

PR #1716, opened 2016-09-07
Support for ImmutableJS setIn array paths

Reanmachine:
## Problem

ImmutableJS's `setIn` (and thereby `updateIn` / `deleteIn` / etc...) do not seem architected to work with a mixture of `Map` and `List` objects. The call assumes that an "unset" value can only be one thing, and only has one resolution path for when that value is empty. (Namely, add it as an empty map) [See Immutable-js Source](https://github.com/facebook/immutable-js/blob/master/src/Map.js#L802)

This leads to the issue found in #1574 where we get the infamous error:

    Uncaught TypeError: list.splice is not a function

When using immutable-js with `FieldArray`. The focus / blur events are calling `setIn` on `fields.fieldName[0].touched` etc... which causes the `fieldName` part of the tree to be instantiated as `Map` instead of a `List`.

Actual | Expected
------------ | -------------
```{ "fields": { "fieldName": { "0": { "touched": true } } } }``` | ```{ "fields": { "fieldName": [ { "touched": true } ] } }```

## Solution

We need to pre-seed a mixed type path like `fields.thing[0].visisted` with empty `List` objects at each array breakout point. This allows immutable to simply update the existing `List` with a call to `set()` instead of trying to instantiate it as a `Map`

I have broken the `structure/immutable`'s `setIn` method out [into it's own file](https://github.com/Reanmachine/redux-form/blob/reanmachine/immutable-array-setIn/src/structure/immutable/setIn.js) to allow us to take some liberties with it.

## Extra

Along with this PR I've included regression tests for the `setIn` behaviors I fixed as well as some tests for some of the expected vanilla `setIn` behavior the library depends on.

### DIFF
--- a/src/structure/immutable/__tests__/setIn.spec.js
+++ b/src/structure/immutable/__tests__/setIn.spec.js
@@ -1,0 +1,121 @@
+import expect from 'expect'
+import { fromJS, Map, List } from 'immutable'
+import setIn from '../setIn'
+
+describe('structure.immutable.setIn', () => {
+  it('should handle undefined', () => {
+    let result = setIn(new Map(), undefined, 'success')
+    expect(result).toEqual('success')
+  })
+  it('should handle dot paths', () => {
+    let result = setIn(new Map(), 'a.b.c', 'success')
+
+    let a = result.get('a')
+    expect(a).toExist('a missing')
+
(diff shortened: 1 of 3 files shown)


erikras:

PR #1722, opened 2016-09-07
fixes #1704 `Uncaught (in promise)`

### LINKED ISSUE #1704: 6.0.2 AsyncValidation fails with `Uncaught (in promise)` on failed validation
here is a code from my validator: (works in 5.3.3)
```javascript
export function imageSize(fieldName, config) {
  if (!config) throw new Error('Must provide config to imageSize.');

  return (values) => {
    const file = values[fieldName];
    if (!file) return Promise.resolve();

    return loadImage(file)
      .then((image) => {
        if (!isImageInSize(image, config)) {
          throw `image must be ${config.width}x${config.height}.`;
        }
      })
      .catch((error) => {
        throw { [fieldName]: error };
      });
  };
}
```

In 6.0.2 I have the following error on failed validation:
apprime

vitalis:
fixes #1704
removes reject from `listenToSubmit` cause no one handles this reject

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -255,9 +255,8 @@ const createReduxForm =
               return promise
             }
             this.submitPromise = promise
-            return promise.then(this.submitCompleted, err => {
+            return promise.then(this.submitCompleted, () => {
               this.submitCompleted()
-              return Promise.reject(err)
             })
           }
 


erikras:

smiller393:
This bug still appears to be happening even in the current version (6.5)

PR #1734, opened 2016-09-09
Reuse connected ConnectedField, ConnectedFields and ConnectedFieldArray

huan086:
Redux `connect()` returns a new class. Currently, each `ConnectedField`, `ConnectedFields` and `ConnectedFieldArray` is connected to redux, thus creating many classes. This change connects `ConnectedField`, `ConnectedFields` and `ConnectedFieldArray` only once.

In a field array, if we would like to preserve state of our component, we need to use a key that identifies the element, instead of array index. However, doing so, `Field`, `Fields` and `FieldArray` detects that it is now bound to a new field name and discards the old connected class and `connect` again. React sees that the class is different, thus discards the old virtual DOM, along with the `component` and its state.

By using `mapStateToProps(state, ownProps)` and `mapDispatchToProps(dispatch, ownProps)`, we can [dynamically map according to the current ownProps.name or ownProps.names](https://github.com/reactjs/react-redux/issues/265#issuecomment-174344345), thus `ConnectedField`, `ConnectedFields` and `ConnectedFieldArray` only needs to be connected once. React will see the same class and reuse the virtual DOM, preserving our `component` and its state.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -3,18 +3,13 @@ import { connect } from 'react-redux'
 import createFieldProps from './createFieldProps'
 import plain from './structure/plain'
 
-const createConnectedField = ({
-  asyncValidate,
-  blur,
-  change,
-  focus,
-  getFormState,
-  initialValues
-}, { deepEqual, getIn }, name) => {
+const propsToNotUpdateFor = [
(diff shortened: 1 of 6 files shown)


erikras:

Issue #1708, opened 2016-09-06
Option to keep submit errors on change


orfjackal:
In my current project we have high accessibility requirements (e.g. blind people must be able to use the site) and a thing that has been reported as problematic is the disappearance of form validation error messages before the form is submitted.

Currently redux-form is hard-coded to clear the errors always on field change. I'd like to change that so that `reduxForm` can take a `clearErrorsOnChange: false` parameter which will disable this feature. I think that only the `submitError` property should be affected by this parameter, and not `asyncError`. What do you think? Would `clearSubmitErrorsOnChange` or some other name be better?

I have a prototype of the change already working locally. Would you like me to make a pull request? Should I make it for v5 and/or v6? We're still using v5 for now, so I can make a pull request for both versions.

orfjackal:
I've created a pull request for v6. I also have a branch for v5, but that doesn't handle _error, so I'm not making a pull request for it, considering that we'll anyways sooner or later migrate to v6.

erikras:
Fix published in [`v6.0.3`](https://github.com/erikras/redux-form/releases/tag/v6.0.3).

PR #1735, opened 2016-09-09
Option to keep submit errors on change

orfjackal:
See #1708

The change to the `if (valid)` check in handleSubmit.js is still a bit of an hack. I think the cleanest solution would be to call `validate` there, similar to `asyncValidate`.


erikras:
It's API surface area creep, but it seems reasonable and the PR is excellent.

mattdknapp:
The persistentSubmitErrors feature is something that I am really keen on. I was wondering when this might be published to npm?

erikras:

PR #1741, opened 2016-09-11
Improved performance of getIn()

huan086:
Rewrite recursive rest spread getInPath into simple for loop for 10x and 5x speed improvement in Firefox and Chrome respectively.

I've removed the special case for paths of length 1, 2 and 3 added in https://github.com/erikras/redux-form/pull/1667. That resulted in a 1% and 2% decrease in performance in Firefox and Chrome respectively. If the 1% to 2% is important, I'll add the special cases back.

Benchmark
https://jsfiddle.net/huan086/tbjebaz8/8/

Recursive rest spread is a bad idea, as it is creating a lot of new arrays and putting a strain on garbage collection.

### DIFF
--- a/src/structure/plain/getIn.js
+++ b/src/structure/plain/getIn.js
@@ -1,13 +1,5 @@
 import { toPath } from 'lodash'
 
-const getInWithPath = (state, first, ...rest) => {
-  if(!state) {
-    return state
-  }
-  const next = state[first]
-  return rest.length ? getInWithPath(next, ...rest) : next
-}
-
 const getIn = (state, field) => {
   if (!state) {
     return state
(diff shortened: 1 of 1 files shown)


erikras:
> If the 1% to 2% is important, I'll add the special cases back.

I think 1-2% performance is worth the _slightly_ larger bundle. Go ahead and add them back.

huan086:
Adding the special cases didn't change anything. Seems like the tests are affecting the performance of each other. If I run the new getIn first, it outperforms the old getIn.

https://jsfiddle.net/tbjebaz8/13/

Most likely GC is kicking in to collect the arrays created by the toPath calls, causing the 1 to 2% difference.

erikras:

PR #1762, opened 2016-09-14
Improve dirty handling for checkboxes

bbenezech:
`dirty(undefined, false)` should be `false`. It was `true`.

Intuitively, clicking twice on a check-box should return it to
pristine, which is false atm (unless the checkbox is initialized with
`false`).

Same behavior as `empty` and `undefined` for a string.

Also decreases code complexity. If both `value` and `initial` are falsey,
return “equal true”.

Comes with tests, both `immutable` and `plain` structures are covered.

I would be very pleased to see it shipped. :hatching_chick:

### DIFF
--- a/src/structure/immutable/__tests__/deepEqual.spec.js
+++ b/src/structure/immutable/__tests__/deepEqual.spec.js
@@ -275,5 +275,17 @@ describe('structure.immutable.deepEqual', () => {
       }
     }), true)
   })
+
+  it('should treat false and undefined as equal', () => {
+    testBothWays(fromJS({
+      a: {
+        b: false
+      }
+    }), fromJS({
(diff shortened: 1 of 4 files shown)


erikras:

erikras:

erikras:
Revert published in [`v6.2.0`](https://github.com/erikras/redux-form/releases/tag/v6.2.0).

erikras:
⚠️ __IMPORTANT__ ⚠️  As of [`v7.0.0`](https://github.com/erikras/redux-form/releases/tag/v7.0.0), `undefined` and `false` are no longer equal. [Read more](https://github.com/erikras/redux-form/releases/tag/v7.0.0).

PR #1767, opened 2016-09-15
Bound change and blur to dispatch

### LINKED ISSUE #1696: [v6] change and blur form props are not bound to dispatch
The `change` and `blur` props are not bound to dispatch as the docs suggest.

See here: https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L443

For now have to dispatch manually.

erikras:
Fixes #1696 and its dupes.


erikras:

PR #1774, opened 2016-09-15
fix checkox behavior

bbenezech:
Follow up to #1762
When a checkbox is initialized with undefined or is not initialized,
`false` should be treated as `undefined` and removed from the value
tree.
It follows the convention of text fields in redux-form (‘’ value is
removed when initial is `undefined`) and follows web form checkbox
original behavior (a checkbox is checked or ‘absent’).
Of course it won’t affect forms where checkboxes are initialized with
`false` (double-clicking them will bring back `false` instead of
`undefined`, on par with text fields in redux-form)


erikras:

travisbloom:
@erikras This change made it so if a checkbox is initialized with `true`, setting it to false actually sets it to `undefined`. It is no longer possible to set a checkbox to `false`

erikras:
⚠️ __IMPORTANT__ ⚠️  As of [`v7.0.0`](https://github.com/erikras/redux-form/releases/tag/v7.0.0), `undefined` and `false` are no longer equal. [Read more](https://github.com/erikras/redux-form/releases/tag/v7.0.0).

PR #1787, opened 2016-09-16
Change 'homepage' schema

h4:
Package.json contains wrong schema for homepage option. This option is used by npm-check utility for example and when you click on link in console you got security warning in your browser.

If you don't have plans to move site to https, then it will be nice to unify site links.


erikras:
Actually, I've been meaning to enable SSL, so this was a good reminder. Should be up within 24 hours. Thanks.

PR #1789, opened 2016-09-16
Fixes #1785. Ensures component is updated if props length changes.

### LINKED ISSUE #1785: Field holding onto previous props?
This is demonstrated in [this JSBin](https://jsbin.com/reparab/3/edit?html,js,output). I have a toggle button that, when clicked, should essentially change the `Field` between:

```
<Field
  name="foo"
  component="input"
  type="text"
/>
```

and 

```
<Field
  name="foo"
  component="textarea"
/>
```

When clicking the toggle button, you'll notice it doesn't immediately switch between an input and a textarea. However, if you then click on the input or textarea, it will then usually switch to what it should be. It's like props aren't being propagated correctly until the following render cycle.

If I either never set `type` or I always set `type` (to either `"text"` or `""`) then the input/textarea will immediately toggle when clicking the toggle button as expected.

Anyway, something funky is going on. I apologize if it turns out to just be a fundamental misunderstanding of React props or something similar.

Aaronius:
Fixes #1785.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -20,7 +20,7 @@ const createConnectedField = ({ deepEqual, getIn }) => {
     shouldComponentUpdate(nextProps) {
       const nextPropsKeys = Object.keys(nextProps)
       const thisPropsKeys = Object.keys(this.props)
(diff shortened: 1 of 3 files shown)


erikras:
Yep, looks like this was introduced in #1734.

PR #1795, opened 2016-09-17
Fix #1677: Add `name` argument to `parse` & `format` props for Field + Fields

### LINKED ISSUE #1677: [v6] parse and format should be called with field name
I have `Fields` that rely of fields of different type. For example, a time selector consisting of hour, minute (number) and am/pm (string).

All the fields trigger the same `parse` and `format`, giving no way to differentiate the number fields from the string fields

kevinzwhuang:
Description
----

Currently, `parse` and `format` props are called with only the field `value`.:
>`format : (value) => formattedValue` [optional]
>`parse : (value) => parsedValue` [optional]

As described in #1677, there are cases where it would be beneficial for it to be called with the field `name` such that the interface for both would become:
>`format : (value, name) => formattedValue` [optional]
>`parse : (value, name) => parsedValue` [optional]

It would allow for fields of the same component that rely on different types of values to be differentiated by `name` in the value lifecycle through `parse` & `format`.

Changes Proposed
----

- Add field `name` as a second input to `parse` & `format`
- Update tests for `parse` & `format` for both Field & Fields components with new `name` argument
- Update docs for Field & Fields to explain how `parse` & `format` props function with access to the field `name`.

---
Fixes #1677

### DIFF
--- a/docs/api/Field.md
+++ b/docs/api/Field.md
@@ -38,10 +38,13 @@ be as simple as `'firstName'` or as complicated as
 A `Component`, stateless function, or string corresponding to a default JSX element.
 See the [Usage](#usage) section below for details.
 
-#### `format : (value) => formattedValue` [optional]
+#### `format : (value, name) => formattedValue` [optional]
 
 Formats the value from the Redux store to be displayed in the field input. Common use cases are 
 to format `Number`s into currencies or `Date`s into a localized date format.
+
+`format` is called with the field `value` and `name` as arguments and should return the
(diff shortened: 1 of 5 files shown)


yched:
Concur. The parse() & format() props of the Fields component seem pretty useless as is.

Since afaict Fields is the recommended way to do do dependant fields, then at the moment you cannot have a field that both :
- depends on another field (e.g disabled if some other field is empty)
- and has its values massaged in and out with format() / parse()

erikras:

PR #1816, opened 2016-09-19
Fix `name` prop not being passed down to children

m14t:
In commit 89e183f, additional properties were extracted from `this.props`, which in turn removed them from the `...rest` variable.

This commit manually adds the `name` prop back into the generated lists of props.

I'm not 100% sure that this fixes all of the things that were changed in that commit, but it does fix the `FieldArray` `component` not getting this property.


erikras:

Issue #1817, opened 2016-09-19
Reducer plugin can't see previous state


rennat-eb:
We have codependent form fields each other (start and end times) so we are using a plugin to change the end time when the start time changes. We're doing this using `reducer.plugin` and listening to redux-form's build in `CHANGE` event.

When this plugin runs, it receives the result of the built-in reducer without access to the previous values.

This seems like a trivial fix by adding an additional argument to plugin functions `previousState` or including previous values in `CHANGE` action's meta information.

Thoughts?

erikras:
Fix published in [`v6.1.0`](https://github.com/erikras/redux-form/releases/tag/v6.1.0).

PR #1822, opened 2016-09-20
Adds a third argument when calling reducer plugins

rennat:
This argument is the state before the redux form and other plugins reduce the action to the state provided as the first argument.

This fixes issue #1817

### DIFF
--- a/src/__tests__/reducer.plugin.spec.js
+++ b/src/__tests__/reducer.plugin.spec.js
@@ -138,10 +138,10 @@ const describePlugin = (vanillaReducer, expect, { fromJS, deleteIn, getIn, setIn
     
     const plugin = (state, action, startingState) => {
       if (action.type === CHANGE && action.meta.field === 'cat') {
-         let result = state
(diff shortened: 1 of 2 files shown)


src/__tests__/reducer.plugin.spec.js
```diff
@@ -115,6 +118,53 @@ const describePlugin = (vanillaReducer, expect, { fromJS, deleteIn }) => () => {
...
+          cat: { touched: false },
+          lastCat: { touched: false }
+        }
+      }
+    })
+
+    // this plugin will change the value we are after so we can confirm we get the real starting state
+    const intermediatePlugin = (state) => setIn(state, 'values.cat', 'zed')
+    
+    const plugin = (state, action, startingState) => {
```
rennat: The name `startingState` was chosen because it made sense to me at the time but I expect core maintainers to have a better name for this argument.

erikras: `prevState` would be more React-y. It _is_ the previous state before this action, is it not?

erikras:

PR #1844, opened 2016-09-22
Add ability to use function in enableReinitialize

zhigang1992:
This is not final yet.

Just want to know what you guys think.

The use case for this is:

I've a, b, c field in the form, I only want to reload the form if initialValues for c changed. If a or b changed. 
Don't reload.

I'm aware of `keepDirtyData`, but still if we went for re-init, we lost all the meta data and error too.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -116,7 +116,12 @@ const createReduxForm =
           }
 
           initIfNeeded(nextProps) {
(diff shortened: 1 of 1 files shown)


erikras:
Seems fine. Needs a test and the docs updated.

erikras:
Closing due to inactivity.

PR #1589, opened 2016-08-23
Put custom props in proper place in Fields

erikras:
The custom props weren't going into the right place in the props passed to `component`. Fixes #1583.

Issue #927, opened 2016-05-08
v6 how to expose input ref for focus() and blur() with the new Field Component?


ms88privat:
I need to call focus() and blur() programmatically on my input field. But with the new Field wrapper, I don't see an easy way to get access to the ref of the input field. I think the Field Component should implement the methods focus() and blur() if possible.

kristian-puccio:
Yes would be very nice to have

erikras:
To clarify, you mean...
```js
focusMyField() {
  this.refs.myFieldRef.focus()
}

render() {
  return (
    <div>
      <Field name="myField" ref="myFieldRef" component="input"/>
      <button onClick={this.focusMyField.bind(this)}>Focus</button>
    </div>
  )
}
```
??

ms88privat:
@erikras yes

eliseumds:
Ye, makes sense. But we may just need a method to access the rendered field component:

```js
focusMyField() {
  this.refs.myFieldRef.getRenderedComponent().focus()
}

render() {
  return (
    <div>
      <Field name="myField" ref="myFieldRef" component="input"/>
      <button onClick={this.focusMyField.bind(this)}>Focus</button>
    </div>
  )
}
```

Otherwise we might call these methods that are not available on the final component (custom ones).

erikras:
I've implemented `getRenderedComponent()`. It will be published in the next alpha release.

erikras:
Published as [`v6.0.0-alpha.8`](https://github.com/erikras/redux-form/releases/tag/v6.0.0-alpha.8).

ms88privat:
👍 thanks

erikras:
@ms88privat Thanks for bringing my attention to [react-native-stateless-form](https://github.com/danielweinmann/react-native-stateless-form). Very interesting project.

ArmorDarks:
Just in case, there is another method for doing this for stateless components: https://github.com/facebook/react/issues/4936#issuecomment-179917137

Noitidart:
> I've implemented getRenderedComponent(). It will be published in the next alpha release.
Hi there @erikras - I am trying to access the ref from the `onSubmitFail` in my react-native app. However I cannot access that. Is there some way to get the ref to the node in `onSubmitFail`? My goal is to scroll to the `<TextInput />` element.

PR #1845, opened 2016-09-22
Fix ConnectedField withRef/getRenderedComponent() regression for text components

Pisi-Deff:
Regression caused by src/ConnectedField.js changes in PR #1589, partly breaking feature added for #927 

`custom` is now initialized before props.ref is assigned. This means that the ref prop will not reach a text-based component like `input`.

The fix is trivial, just need to assign ref as a property to `custom` instead of to `props`.
Being in `custom`, it will end up in both text and React components.

Testing the change locally seems to indicate that getRenderedComponent() works properly again.


erikras:

PR #1851, opened 2016-09-22
arrayPush should handle a null leaf value

marbemac:
The constructs being used in argument assignment only account for undefined values, for example the array property below:

```
const splice = (array = [], index, removeNum, value) => {
```

If the array value is null, it will not be assigned to be an empty array. This causes any code that assumes array will be an actual array (and not null) to break.

I've added a test to demonstrate the bug, and a simple fix. Basically, do the assignment outside of the arguments, so that null values are handled properly, and list.count() and array.length don't fail.

This might affect other areas of the codebase. If the foo = bar arg assignment construct is being used anywhere, the code in that function has to be protected against null values (or don't use that arg construct, just do default arg assignment in the function itself). 

Thanks for redux-form, loving it :).

### DIFF
--- a/src/structure/plain/splice.js
+++ b/src/structure/plain/splice.js
@@ -1,4 +1,6 @@
-const splice = (array = [], index, removeNum, value) => {
+const splice = (array, index, removeNum, value) => {
+  array = array || []
+
   if (index < array.length) {
     if (value === undefined && !removeNum) { // inserting undefined
       const copy = [ ...array ]
--- a/src/__tests__/reducer.arrayPush.spec.js
+++ b/src/__tests__/reducer.arrayPush.spec.js
@@ -83,6 +83,26 @@ const describeArrayPush = (reducer, expect, { fromJS }) => () => {
         }
       })
   })
+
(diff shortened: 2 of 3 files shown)


src/structure/plain/splice.js
```diff
@@ -1,4 +1,6 @@
-const splice = (array = [], index, removeNum, value) => {
+const splice = (array, index, removeNum, value) => {
+  array = array || []
+
```
erikras: How is this better than the default value?

erikras:
Oh, I see. This is the "`null` doesn't trigger default values, but `||` does" issue that was [discussed on Twitter today](https://twitter.com/matzatorski/status/778868456569630720).

marbemac:
Yeah, it's very annoying - I'm sure most people expect null to trigger a default value in argument assignment. It's bitten me a few times.

marbemac:
I didn't audit the entire codebase - do you know offhand if there are other places that this might have an impact on?

ruiaraujo:
@erikras Can we get a release with this?

erikras:

Issue #1895, opened 2016-10-03
[v6.0.5] Proposal: make createOnChange() and createOnBlur() maintain cursor position


timhwang21:
See #1396 for example (image reproduced here). 

![](https://cloud.githubusercontent.com/assets/544097/17120238/fbff2aac-5298-11e6-8726-fe5c593b59ba.gif)

Given that this is an "issue" with React, I'm not sure that redux-form should bother with dealing with this. However, it is possible for redux-form to fix this behavior, which I think would lead to better UX.

timhwang21:
Closing this -- upon thinking about it, this doesn't seem like a relevant concern for a form library. If people disagree they can reopen.

Issue #1396, opened 2016-07-25
[v6] Normalization caret position is lost


smirea:
Not sure if this should specifically be handled by redux-form, but the phone example with normalization loses the caret position when typing/deleting from the middle of the value

http://redux-form.com/6.0.0-rc.3/examples/normalizing/

![normalization](https://cloud.githubusercontent.com/assets/544097/17120238/fbff2aac-5298-11e6-8726-fe5c593b59ba.gif)

brotzky:
I've had the same issue on my normalized input fields. I'm not sure if it's redux-form, react, or the normalization library.

DTwigs:
I am having this same problem.  I believe it is caused by normalize replacing the inputs value which bumps the cursor to the end.

erikras:
React is the one in charge of the cursor position (and it's a hard problem), so this might not be fixable, I'm afraid. Certainly not by this library.

jampy:
This makes normalization really problematic, IMHO.

Is there any way to delay normalization until BLUR, meaning that the input text does not normalize until the user leaves the field?

brotzky:
Why wouldn't you want to normalize right away? Makes entering the input for users much easier if it's formatted automatically.

jampy:
@brotzky, right, but the behavior of the caret in certain situations (as described in this issue) is a **very** bad UX and a no-go IMHO. 

Any solution very welcome...

smirea:
since opening this i've come to the conclusion that these are the pro-s and cons of using normalization. for the vast majority of cases a user either types the entire string at once or deletes the entire thing and re-types it in which case this problem does not resurface. 
I personally chose to ignore the other group, but if you want, you can implement a combination of realtime + blur normalization based on the caret position

```jsx
class NormalizedInput extends React.Component {

    state = { value: '' }

    constructor (props) {
        super(props);
        this.state.value = props.input.value;
    }

    componentWillReceiveProps (nextProps) {
        if (nextProps.input.value === this.props.input.value) return;
        this.setState({value: nextProps.input.value});
    }

    onChange (event) {
        const {target} = event;
        const {value} = target;
        this.setState({value});
        if (target.selectionStart !== value.length) return;
        this.props.input.onChange(event);
    }

    render () {
        const {input: {value, onChange, ...input}} = this.props;
        return <div>
            <input type='text' {...input} value={this.state.value} onChange={::this.onChange} />;
            <div>props.input.value: {this.props.input.value}</div>
[...]

PR #1896, opened 2016-10-03
Maintain cursor position on parse/normalize (#1895)

timhwang21:
References #1895 and #1396.

* Modifies createOnChange() and createOnBlur() to handle
  moving the caret to the appropriate place when the value
  is changed
* Set to only work for input types that support .selection
* Can probably be cleaned up -- this is just a POC for now

### DIFF
--- a/src/events/createOnBlur.js
+++ b/src/events/createOnBlur.js
@@ -12,7 +12,7 @@ const createOnBlur =
       let distanceFromEnd = -1
       let validInputTypes = [ 'text', 'search', 'url', 'tel', 'password' ]
       // get distance between caret and end of original value
-      if (isEvent(event) && validInputTypes.includes(event.target.type)) {
+      if (isEvent(event) && validInputTypes.indexOf(event.target.type) !== -1) {
         distanceFromEnd = value.length - event.target.selectionEnd
       }
 
--- a/src/events/createOnBlur.js
+++ b/src/events/createOnBlur.js
@@ -1,4 +1,5 @@
 import getValue from './getValue'
+import isEvent from './isEvent'
 import isReactNative from '../isReactNative'
 
 const createOnBlur =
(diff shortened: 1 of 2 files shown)


timhwang21:
Closing this -- upon thinking about it, this doesn't seem like a relevant concern for a form library. If people disagree they can reopen.

andresgutgon:
Ok, but If it's not responsibility of redux-form and I use it, what could be the solution?

To me, as a user of a form, the jumping caret looks like an annoying bug :/

timhwang21:
@andresgutgon the code in the PR works, mostly. Take a look, you may be able to better adapt it.

Basically, I added a check to see if field value gets changed by `normalize` or `parse`. If so, I (sort of hackishly) asynchrously move the cursor back to where it should be.

Unfortunately I also don't know how to write tests for such code since I haven't worked with this sort of DOM-related testing!

andresgutgon:
Ok, I think I'm doing something wrong:

I can see on simple form example caret is not jumping:
http://redux-form.com/6.0.5/examples/simple/

Also cursor is not jumping on vanilla React:
http://codepen.io/anon/pen/rrPwvY

My input doen't have `normalize` prop. What I'm doing wrong?

timhwang21:
^ read through the [example on normalization](http://redux-form.com/6.0.5/examples/normalizing/).

andresgutgon:
Yes, I'm understand what normalization is doing. What I mean is that I'm not using `normailize` prop and I see the jumping caret.

```javascript
<Field component="input" name="my_field" />
```

timhwang21:
Maybe comment on one of the existing issues as your issue isn't really relevant to this PR? It's hard to help without seeing some code, or a clearer description of the problem.

andresgutgon:
Agree, sorry for the noise. I think I'm messing with redux state making unecessary re-renders. 

Thanks for your help :)

PR #1843, opened 2016-09-22
fix asynch validation uncaught promise

elodszopos:
We're at the end of the promise chain already. There's already a rejection coming from above (either by `throw` or `reject`).

Hope this helps.

Fixes #1297.

PS: Thanks for all the awesome @erikras and all others involved.

PR #1930, opened 2016-10-08
Make handleSubmit doesn't submit when asyncValidation fails

### LINKED ISSUE #1929: [6.1.0] Form has been submitted even when async validation fails
In #1843, this change was made: https://github.com/elodszopos/redux-form/commit/73d60f8a75512e8b0fe16d3884f10f6994be8bd9

The test passes because [it still returns a rejected promise](https://github.com/erikras/redux-form/blob/f6e1fb9316f6201f259d0ce761908618513f2e0c/src/__tests__/handleSubmit.spec.js#L103)

Also, this test needs to be rewritten, because even if you return a resolved promise it passes (`catch` doesn't run).

diegohaz:
I've changed the code in a way that preserves changes from #1843 

Fix #1929

### DIFF
--- a/src/__tests__/handleSubmit.spec.js
+++ b/src/__tests__/handleSubmit.spec.js
@@ -108,7 +108,7 @@ describe('handleSubmit', () => {
         throw new Error('Expected to fail')
       })
       .catch(result => {
-        expect(result).toBe(undefined)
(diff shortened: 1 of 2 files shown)


erikras:

PR #1942, opened 2016-10-10
Sub-form support

Zagitta:
Greetings, in this PR I've implemented support for nested forms/sub-forms. What this basically does is prefix the name of Field, Fields and FieldArray when wrapped in a SubForm component.
For example the following: 
```
<form>
  <Field name="amount" component="input" type="text"/>
  <SubForm name="buyer">
    <Field name="name" component="input" type="text"/>
    <SubForm prefix="address">
      <Field name="street" component="input" type="text"/>
    </SubForm>
  </SubForm>
</form>
```
will result in the following 3 fields being registered: `["amount", "buyer.name", "buyer.address.street"]`.
This feature is desirable because it allows reusing standard components multiple times in the same form. An example use-case of this is an order form where there's a buyer and a recipient which both contain the same fields and ideally can reuse the same react component for both. With this feature it's simply a question of wrapping the 2 `<Party/>` instances like this: 
```
<SubForm name="buyer"><Party/></SubForm>
<SubForm name="buyer"><Party/></SubForm>
```

If there's interest in adding this feature to redux-form I would be happy to write tests and documentation for the feature and fix any other potential issues you might have with the code.

Best regards, Simon


erikras:
Wow, this is simpler than I thought it would be. Great job! 👍 

When I have thought in the past about making nested forms, I thought that maybe the actual form component could check its context to see if it was inside an outer form, and cede control (validation, submission, etc.) to the outer form, rather than have a separate `SubForm` component. That way an individual form could stand by itself OR be encapsulated inside a master form, however that might not work very well with relation to displaying submit buttons, etc.

Correct me if I'm wrong, but what you've written here is really a "field prefixer" intermediary, right? This is just my initial reaction, but there's something about the name `SubForm` that seems wrong, because it's not a form itself. Why don't we call it something like `FieldGroup` or `FormSection`? ...or maybe `SubForm` is fine.

> If there's interest in adding this feature to redux-form I would be happy to write tests and documentation for the feature and fix any other potential issues you might have with the code.

"Yes please" to the offer of tests and docs. This looks great!

Zagitta:
I agree having a separate `SubForm` component is a tad ugly but to me it seems like the best balance between ease of use, implementation complexity and composability. I'm not so sure about the multiple `form` approach, not only would it technically be invalid HTML but also complicated to implement. To me it also smells a bit like an anti-pattern by needing to connect dumb components (the components responsible for each sub-part of the form) with redux-form just to make them nest nicely. 
And yes you are indeed correct, `SubForm` is just there to modify the prefix of child fields. I think `FormSection` is a much more suitable name :+1: 

I will started writing test and docs then :smile:

erikras:
> it seems like the best balance between ease of use, implementation complexity and composability

Agreed! 👍 

> I think `FormSection` is a much more suitable name.

Perfect. 🎉

Zagitta:
Okay I've added some tests to Field and Fields but I'm not entirely sure how to test FieldArray. Do you have any suggestions? 
Furthermore I'm not entirely sure how to add a new spec file for FormSection, is it necessary to implement the same `describeX =  (...) => { }` setup that exists in Field, Fields and FieldArray or is that only for components that are impacted by immutable.js? 

Sorry for the many questions, I'm afraid I'm not expert in creating tests :slightly_smiling_face:

erikras:
No need to apologize.

> is it necessary to implement the same `describeX = (...) => { }` setup that exists in `Field`, `Fields` and `FieldArray` or is that only for components that are impacted by immutable.js?

Even though all `FormSection` is doing is string manipulation to prefix the name, it will be affecting the `Field`'s inside it, so I think it might be worth setting up the `describeX` setup like `Field` et. al.

erikras:
Tests you need:

* a test just like [this one](https://github.com/erikras/redux-form/blob/68c18af88710669178ca03d1ae69c465a8a4cfb5/src/__tests__/reduxForm.spec.js#L1314-L1353), but with the fields registering properly from inside a `FormSection`.
* a form with a `Field` inside a `FormSection` that shows that calling `onChange` updates the value at the proper depth in the store. You check that [like this](https://github.com/erikras/redux-form/blob/49d0bc58f96dc355b397576f5838dcd2bcb63c3f/src/__tests__/Fields.spec.js#L948-L965).
* Ditto for `FieldArray`, maybe just testing a `push()`

Zagitta:
I believe that should be everything :slightly_smiling_face: 
Turns out there was one critical issue where Fields rendered inside a FieldArray would prefix the field name twice so I added a simple regex to the prefixName function checking if the field name ends in square brackets with one or more digits between them (`/\[\d+\]$/`).

If you have any more requests or issues with the PR let me know and I'll fix them.

erikras:
Fantastic! Great job. 👍

erikras:
@Zagitta Just noticed your awesome surname. ❤️ 🇩🇰

aislanmaia:
I'm using redux-form v.6.1.0 and getting this:
"TypeError: Super expression must either be null or a function, not undefined"

My component code:

```
import React, { Component } from 'react';
import { FormSection, Field } from 'redux-form';
import Autocomplete from 'react-toolbox/lib/autocomplete';
import Input from 'react-toolbox/lib/input';
import Dropdown from 'react-toolbox/lib/dropdown';


class ResponsibleFormSection extends FormSection {
  static defaultProps = {
    name: "responsible"
  }

  ...
```

I'm trying to follow the docs that says it is possible to inherit directly from FormSection.

Zagitta:
A version containing `FormSection` has not yet been published to npm as far as I can tell, you'll have to wait a bit until @erikras publishes it and then update your redux-form version :slightly_smiling_face:

aislanmaia:
Thanks for explaining it for me, @Zagitta. 
I hope this will be published soon. :smile:

erikras:

erikras:
@m1m6 `FormSection`s are primarily for repeatable, reusable data _shapes_. The best example is an address, where you have the same `street`, `city`, `postalCode` fields that can occur in multiple places. As for validation, that's up to you. Surely a `validateAddress()` function could be reused for multiple subsets of your form data. You could also use the new [field-level validation](http://redux-form.com/6.4.3/examples/fieldLevelValidation/), although that will not be as flexible if you define the validation inline with the fields (e.g. maybe one address section requires the `street`, but another doesn't).

m1m6:
@erikras That's right, but i didn't mean validation by it self here, i meant other redux-form stuff like initialValue also, what i'm trying to do is to separate form's from each others, so that i can use any form at any place just by calling it.

gustavohenke:
Hi @m1m6, if I understood you, you're trying to create reusable form sections, so you could use some `ContactSection` for Users and for Customers?

As the whole idea of React is to compose stuff, you would need to compose selectors, validators and etc to build your forms.

m1m6:
@gustavohenke  Yes, that's what i'm trying to do, can u please provide an example or a road map about how could i do that ?

gustavohenke:
Sure.
Take a look at the following libs:
- https://github.com/jfairbank/revalidate
- https://github.com/reactjs/reselect

They are very helpful to compose stuff and to get your imagination started!

```js
import { createSelector } from 'reselect';

const userSelector = state => state.user;

// Can be reused if customer contact info uses the same keys
const contactSelector = object => ({
  phone: object.phone,
  email: object.email
});

const userContactSelector = createSelector(
  userSelector,
  contactSelector
);

// Same idea for customer business rules
const userInitialValuesSelector = createSelector(
  userSelector,
  userContactSelector,
  ( user, contacts ) => ({
    userPersonalData: { foo: user.foo, bar: user.bar },
    contacts
  })
);

// then, in your mapStateToProps:
const mapStateToProps = state => ({
  initialValues: userInitialValuesSelector( state )
});
```

m1m6:
@gustavohenke  Thanks a lot bro, this is very help full guide :)

jozsi:
I have another use case for @m14t's question regarding passing `reduxForm` props to `FormSection`.

I have a component that is sometimes connected directly via `reduxForm` (scenario 1), and sometimes it's wrapped inside a `FormSection` (scenario 2).
Inside my component, I have a few fields that are inter-connected: if a field gets emptied, it will empty the other fields as well.

### Scenario 1
I just use `this.props.change('fieldToReset', null)`.

### Scenario 2
Things get complicated here. First, I thought I can just pass the `change` method as a prop to my component. But I can't just call `this.props.change('fieldToReset', null)` as above, because it's wrapped inside a `FormSection` that prefixes it.

@erikras - is there any other way than using a prefixer method for the field name that takes into account the `sectionPrefix` found inside the context?

PR #1945, opened 2016-10-11
Fix sync validation

### LINKED ISSUE #1925: [6.1.0] syncValidateFailed set to true after calling initialize
On some of my wizard-style forms, I manually call the `initialize` action creator to initialize the form.

In `6.0.5`, if I initialized the form I would then be able to to submit the form immediately. In `6.1.0`, `syncValidateFailed` is getting set to true and I can't submit until I update a field at which point the sync errors are reevaluated and `syncValidateFailed` goes away.

I do have a field that's failing sync validation, however it isn't registered at the time so this shouldn't be causing the issue.

ortex:
fix #1925


ortex:
@erikras Please, add a label "Hacktoberfest" to #1925

erikras:

PR #1946, opened 2016-10-12
Store number input value as string to fix #1383

### LINKED ISSUE #1383: Can't type 1.01 on number input
(Only tested on 6.0.0-rc.3)

`<Field name="foo" type="number" component="input" />`

I can't type `1.01` in this field, because when I type `1.0` it immediately gets normalized to `1`

JosephEarl:
Removes the `parseFloat` for number and range types so that decimals can be entered. Fixes #1383.

### DIFF
--- a/src/events/__tests__/getValue.spec.js
+++ b/src/events/__tests__/getValue.spec.js
@@ -109,7 +109,7 @@ describe('getValue', () => {
         type: 'number',
         value: '3.1415'
       }
-    }, true)).toBe(3.1415)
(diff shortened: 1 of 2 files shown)


erikras:

skolmer:
@erikras I don't think it is a good idea to release this change as 6.1.1. This is a breaking change and should not be released as a patch version. It will break things for users that are relying on typed number fields. I can understand that this is an issue and needs to be resolved but since this is not yet published to npm please consider reverting this change or at least create a new major version.

PR #1992, opened 2016-10-19
ReduxForm resets it's initialValues if enableReinitialize is true

### LINKED ISSUE #1902: Form still dirty after `initialize(formName, newData, true)`
We're using redux form to edit some pages, and I'm having some trouble re-initializing the form after saving.

The submit callback is 

```
const mapDispatchToProps = (dispatch) => ({
 submit: (page) => dispatch(updatePage(page)).then((page) => {
      dispatch(initialize(`page/${page.slug}`, page, false));
      dispatch(reset(`page/${page.slug}`));
      return page;
    });
});
```

This seems to be working as expected. The current and initial values for the form are equal, however the `dirty` flag is set to 'true' until I reload the page.

Any ideas on why this might be happening?

This seems to be effecting the state as I expected, the form values stay the same, but the

ncphillips:
Andd the store's version of the initialValues is updated. The values also get set to this new initialValue if keepDirtyOnReinitialize is false.

closes #1902

@skolmer

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -437,13 +437,13 @@ const createReduxForm =
             const formState = getIn(getFormState(state) || empty, form) || empty
             const stateInitial = getIn(formState, 'initial')
 
-            const shouldUpdateInitialValues = !deepEqual(initialValues, stateInitial) && enableReinitialize
-            const shouldResetValues = shouldUpdateInitialValues && !keepDirtyOnReinitialize;
+            const shouldUpdateInitialValues = enableReinitialize && !deepEqual(initialValues, stateInitial)
+            const shouldResetValues = shouldUpdateInitialValues && !keepDirtyOnReinitialize
 
             let initial = initialValues || stateInitial || empty
 
             if (shouldUpdateInitialValues) {
-              initial = stateInitial 
+              initial = stateInitial
             }
 
             let values = getIn(formState, 'values') || initial
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -258,13 +258,13 @@ const createReduxForm =
               const isBlurredField = !submitting &&
                 (!asyncBlurFields || ~asyncBlurFields.indexOf(name.replace(/\[[0-9]+\]/g, '[]')))
               if ((isBlurredField || submitting) && shouldAsyncValidate({
-                asyncErrors,
-                initialized,
-                trigger: submitting ? 'submit' : 'blur',
-                blurredField: name,
-                pristine,
-                syncValidationPasses
-              })) {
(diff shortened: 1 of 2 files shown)


ncphillips:
This is going to fail, because of a test that sets an expectation on the number of render calls on the form component. I do no think this is an actual issue, as it seems like a pretty indirect way to test whether or not something's functioning correctly.

src/reduxForm.js
```diff
@@ -433,11 +433,25 @@ const createReduxForm =
...
         const connector = connect(
           (state, props) => {
-            const { form, getFormState, initialValues } = props
+            const { form, getFormState, initialValues, enableReinitialize, keepDirtyOnReinitialize } = props
             const formState = getIn(getFormState(state) || empty, form) || empty
             const stateInitial = getIn(formState, 'initial')
-            const initial = initialValues || stateInitial || empty
-            const values = getIn(formState, 'values') || initial
+
+            const shouldUpdateInitialValues = !deepEqual(initialValues, stateInitial) && enableReinitialize
```
skolmer: @ncphillips Thanks for working on this!
To get a small performance gain I would consider to reorder this condition:
`const shouldUpdateInitialValues = enableReinitialize && !deepEqual(initialValues, stateInitial)`

ncphillips: nice catch!

ncphillips:
I built this and used it in my own project, and it does NOT solve my problem entirely.

src/reduxForm.js
```diff
@@ -433,11 +433,25 @@ const createReduxForm =
...
+            const shouldUpdateInitialValues = enableReinitialize && !deepEqual(initialValues, stateInitial)
+            const shouldResetValues = shouldUpdateInitialValues && !keepDirtyOnReinitialize
+
+            let initial = initialValues || stateInitial || empty
+
+            if (shouldUpdateInitialValues) {
+              initial = stateInitial
+            }
+
+            let values = getIn(formState, 'values') || initial
```
skolmer: Should this be changed to `const values = getIn(formState, 'values')`. If values is null and initial is something else pristine should be false, right?

skolmer:
@ncphillips I have fixed the test and added a PR to your fork. Lets get this change merged even if it does not fix all of your issues.

src/reduxForm.js
skolmer: ok, forget this one. it results in 26 failing tests ;)

skolmer:
@erikras What is your opinion on this approach to get enableReinitialize working as expected?

erikras:
Seems reasonable.

erikras:

PR #2030, opened 2016-10-26
Don't swallow errors in `onSubmit`

### LINKED ISSUE #1859: handleSubmit swallows errors thrown in onSubmit functions
I was having an issue where I was getting SET_SUBMIT_FAILED on every submit, even though my form had no validation errors. The error would come back empty. After digging into `handleSubmit`, it turns out that the `try` here will catch any type of error and will only give you error messages if caught error has `submitError.errors`.

https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L20
```javascript
      try {
        result = submit(values, dispatch, props);
      } catch (submitError) {
        var error = submitError instanceof _SubmissionError2.default ? submitError.errors : undefined;
```

In my case, I had a `type` error in my `onSubmit` function that was very hard to track down. Should we just rethrow if the error is a general runtime error? Something like this:

```javascript
      try {
        result = submit(values, dispatch, props);
      } catch (submitError) {
        var error = submitError instanceof _SubmissionError2.default ? submitError.errors : undefined;
        if (!error) throw submitError;
```

dallonf:
If an error is thrown by the onSubmit function that is not a SubmissionError, it will now be re-thrown instead of silently ignored. It will retain the previous behavior if you pass an onSubmitFailed prop - personally I don't like that too much, but it would be a breaking change according to the test suite.

Should fix #1859.

Would appreciate feedback!

### DIFF
--- a/src/__tests__/handleSubmit.spec.js
+++ b/src/__tests__/handleSubmit.spec.js
@@ -317,9 +317,14 @@ describe('handleSubmit', () => {
     const asyncValidate = createSpy().andReturn(Promise.resolve())
     const props = { dispatch, startSubmit, stopSubmit, touch, setSubmitFailed, setSubmitSucceeded, values }
 
+    const resolveSpy = createSpy()
+    const errorSpy = createSpy()
+
     return handleSubmit(submit, props, true, asyncValidate, [ 'foo', 'baz' ])
-      .then(result => {
-        expect(result).toBe(undefined)
+      .then(resolveSpy, errorSpy)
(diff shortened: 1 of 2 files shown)


erikras:
Looks good.

andreme:
I don't see a way to intercept errors that are not SubmissionErrors? @erikras will you accept a PR that adds the original error as a 3rd argument to onSubmitFail?

erikras:
@andreme Sure.

erikras:

Issue #2016, opened 2016-10-24
[6.1.1] Bug: state.registeredFields not keeping track of number of calls


huan086:
Suppose you have 2 Field referring to the same property

``` jsx
let showSecondTime = true;
<Field name="theProp" component="input" type="text" />
{showSecondTime && <Field name="theProp" component="input" type="text" />}
```

when showSecondTime is set to false subsequently, the 2nd Field get unmounted, and that clears registeredFields.theProp. However, we still need that for the first Field!
# Proposed solution

keep track of the registration count

huan086:
Related: https://github.com/erikras/redux-form/issues/1990

erikras:
Published in [`v6.2.0`](https://github.com/erikras/redux-form/releases/tag/v6.2.0).

huan086:
Still wrong. Hopefully i'll have some time to fix this

sergi:
@huan086 any luck with this? I am still experiencing with v6.2.0

huan086:
@sergi No... Don't think I'll have time to fix it myself until next year. Could you to give it a shot at fixing?

sergi:
@huan086 Unfortunately not until next year...

halt-hammerzeit:
@sergi @huan086 Btw I wrote my alternative to `redux-form` which fixes this issue.
Or maybe someone coming here from google would be interested.
https://github.com/halt-hammerzeit/simpler-redux-form

huan086:
@halt-hammerzeit I've seen your library. The library don't have the features that I need. Would be great if you put your effort in improving redux-form instead. As for the bloat, don't worry. redux-form is written in ES6, so a bundler like rollup will totally remove the features you didn't use

halt-hammerzeit:
@huan086 And what are the features you wanted that aren't present in my library?

huan086:
@halt-hammerzeit Fields, FieldArray, form level validation, submit success/fail callbacks, place for me to hook custom reducer onto form reducer (I wrote a custom plugin that does splice that accepts multiple inserts to overcome the limitation of redux-form splice), possibly many other callbacks that I can't remember that redux-form provides to allow me to modify its behavior

Basically, it's a large form that redux-form v5 is also not suitable...

erikras:
Fix released as [`v6.5.0`](https://github.com/erikras/redux-form/releases/tag/v6.5.0).

Issue #1990, opened 2016-10-19
Bug: UNREGISTER_FIELD called when component Field is mounted in gets unmounted (even with destroyOnUnmount: false)


mschipperheyn:
I have a problem with a wizard like structure that contains of various components that each have a structure like so.

```
@reduxForm({
  form: 'posteditor',
  validate,
  destroyOnUnmount:false,
})
class FuturePostModal extends Component {
```

So, each is referencing 'posteditor' as form in order to centrally store the input data. Some of the field components are part of Modals. So, the Field is part of one component and the actual field editor is opened as part of a Modal, something like this.

```
class PostActionComponents extends React.PureComponent {

  constructor(props) {
    super(props);
  }

  render() {
    return (
      <View>
        { this.getTagsModal() }
      </View>
    );
  }

  getTagsModal(){
    const { tagModalIsVisible, tags } = this.props.state;
    return (
      <TagModal
          modalVisible={tagModalIsVisible}
          onClose={this.onCloseTagModal}
          onCancel={this.onCancelTagModal}
          tags={tags}
          visible={tagModalIsVisible}
          />
     )
  }

  onCloseTagModal(selectedItems = []){
    const { actions } = this.props;
    actions.saveTags(selectedItems);
    actions.toggleTagModal();
  }

  onCancelTagModal(){
    const { actions } = this.props;
    actions.toggleTagModal();
  }
}


@reduxForm({
  form: 'posteditor',
  destroyOnUnmount:false,
})
class TagModal extends React.PureComponent {

  constructor(props) {
    super(props);
    this.handleSubmit = this.handleSubmit.bind(this);
  }

  handleSubmit(data){
    this.props.onClose(data);
  }

  tags(field) {
    return (

      <SearchBoxHolder
        id={field.name}
        localSearch={field.localSearch}
        url={field.url}
        isRequired={field.isRequired}
        placeholder={field.placeholder}
        selectMultiple={field.selectMultiple}
        maxCountSize={field.maxCountSize}
        minQueryLength={field.minQueryLength}
        queryDelay={field.queryDelay}
        initialValues={field.initialValues}
        idLabel={field.idLabel}
        resultsKey={field.resultsKey}
        resultCountKey={field.resultCountKey}
        resultLabel={field.resultLabel}
        onBack={field.onCancel}
        onClose={(id,data) => {
          field.input.onChange(data => data.map(el => el.id).join(','));
          field.onClose(data);
        }}
        />
    )
  }

  render(){

[...]

PR #2040, opened 2016-10-28
Don't unregister fields on unmount when a single form is concurrently mounted

jscinoz:
This relates to #1990  and #2016. This patch suppresses unregister() for a field if the containing form is still mounted elsewhere, and the containing form itself is being unmounted, rather than just the field itself.

This is useful in conjunction with destroyOnUnmount: false when you have small, self-contained sub-forms embedded in larger forms, or different parts of the UI.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -92,7 +92,10 @@ const createReduxForm =
         pure: true,
         ...initialConfig
       }
+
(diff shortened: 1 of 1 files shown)


erikras:

jscinoz:
Great! Thank you :)

PR #2064, opened 2016-11-02
Add original error as 3rd argument to onSubmitFail

andreme:
This PR is much bigger than I expected initially. When I started to look at handleSubmit it was hard for me to understand and had a lot of duplicated code, so I decided to update it. There are still some weird parts, I will highlight them inline.

This is a breaking change in the sense that ReduxForm.submit now always returns a Promise. The docs already said that is the case, but that was not true:
> submit() : Promise
> Submits the form. [You'd never have guessed that, right?] Returns a promise that will be resolved when the form is submitted successfully, or rejected if the submission fails.

I would like to make some more improvements to handleSubmit, but for now I tried for it to be as close to the previous behaviour as before.

I understand if you will not merge this in, I can rework it to simply add the 3rd param.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "is-promise": "^2.1.0",
     "lodash": "^4.12.0",
     "lodash-es": "^4.12.0",
+    "promise": "^7.1.1",
     "shallowequal": "^0.2.2"
   },
   "devDependencies": {
--- a/src/__tests__/handleSubmit.spec.js
+++ b/src/__tests__/handleSubmit.spec.js
@@ -16,18 +16,19 @@ describe('handleSubmit', () => {
     const props = { startSubmit, stopSubmit, touch, setSubmitFailed, setSubmitSucceeded, values }
 
     handleSubmit(submit, props, false, asyncValidate, [ 'foo', 'baz' ])
-
-    expect(submit).toNotHaveBeenCalled()
-    expect(startSubmit).toNotHaveBeenCalled()
-    expect(stopSubmit).toNotHaveBeenCalled()
-    expect(touch)
-      .toHaveBeenCalled()
-      .toHaveBeenCalledWith('foo', 'baz')
-    expect(asyncValidate).toNotHaveBeenCalled()
-    expect(setSubmitSucceeded).toNotHaveBeenCalled()
-    expect(setSubmitFailed)
-      .toHaveBeenCalled()
-      .toHaveBeenCalledWith('foo', 'baz')
+      .then(() => {
+        expect(submit).toNotHaveBeenCalled()
+        expect(startSubmit).toNotHaveBeenCalled()
+        expect(stopSubmit).toNotHaveBeenCalled()
+        expect(touch)
+          .toHaveBeenCalled()
+          .toHaveBeenCalledWith('foo', 'baz')
+        expect(asyncValidate).toNotHaveBeenCalled()
+        expect(setSubmitSucceeded).toNotHaveBeenCalled()
+        expect(setSubmitFailed)
+          .toHaveBeenCalled()
+          .toHaveBeenCalledWith('foo', 'baz')
+      })
   })
 
   it('should stop and return errors if sync validation fails', () => {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "is-promise": "^2.1.0",
     "lodash": "^4.12.0",
     "lodash-es": "^4.12.0",
+    "promise": "^7.1.1",
```
andreme: promise is already used by another module in the project so I thought it would be best to use it here too

src/__tests__/handleSubmit.spec.js
```diff
@@ -152,7 +156,9 @@ describe('handleSubmit', () => {
         expect(stopSubmit).toNotHaveBeenCalled()
         expect(onSubmitFail)
           .toHaveBeenCalled()
-          .toHaveBeenCalledWith(values, dispatch)
+        expect(onSubmitFail.calls[0].arguments[0]).toEqual(values)
+        expect(onSubmitFail.calls[0].arguments[1]).toEqual(dispatch)
+        expect(onSubmitFail.calls[0].arguments[2]).toBeA(SubmissionError)
```
andreme: change for the 3rd argument

src/__tests__/handleSubmit.spec.js
```diff
@@ -385,7 +391,7 @@ describe('handleSubmit', () => {
 
   it('should submit when there are old submit errors and persistentSubmitErrors is enabled', () => {
     const values = { foo: 'bar', baz: 42 }
-    const submit = createSpy().andReturn(69)
+    const submit = createSpy()
```
andreme: return value was never used

src/__tests__/reduxForm.spec.js
```diff
@@ -1814,7 +1821,9 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
           expect(onSubmit).toNotHaveBeenCalled()
           expect(onSubmitFail)
             .toHaveBeenCalled()
-            .toHaveBeenCalledWith(errors, store.dispatch)
+          expect(onSubmitFail.calls[0].arguments[0]).toEqual(errors)
+          expect(onSubmitFail.calls[0].arguments[1]).toEqual(store.dispatch)
+          expect(onSubmitFail.calls[0].arguments[2]).toBeA(SubmissionError)
```
andreme: change for the 3rd argument

src/handleSubmit.js
```diff
@@ -9,88 +10,86 @@ const handleSubmit = (submit, props, valid, asyncValidate, fields) => {
...
-    }
-  } else {
-    setSubmitFailed(...fields)
-    if (onSubmitFail) {
-      onSubmitFail(syncErrors, dispatch)
-    }
-    return syncErrors
-  }
+      if (onSubmitFail) {
+        onSubmitFail(errors, dispatch, submitError)
```
andreme: change for the 3rd argument

src/handleSubmit.js
```diff
@@ -9,88 +10,86 @@ const handleSubmit = (submit, props, valid, asyncValidate, fields) => {
...
-          onSubmitFail(error, dispatch)
-        }
-        if (error || onSubmitFail) {
-          // if you've provided an onSubmitFail callback, don't re-throw the error
-          return error
-        } else {
-          throw submitError
-        }
+
+  let asyncValidationFailed = false
```
andreme: Depending on whether the async validation fails or not the errors have to be returned in a different way (some tests check for that). I think handleSubmit should always return the errors in the same way.

src/handleSubmit.js
```diff
@@ -9,88 +10,86 @@ const handleSubmit = (submit, props, valid, asyncValidate, fields) => {
...
-            setSubmitFailed(...fields)
-            if (onSubmitFail) {
-              onSubmitFail(error, dispatch)
-            }
-            if (error || onSubmitFail) {
-              // if you've provided an onSubmitFail callback, don't re-throw the error
-              return error
-            } else {
-              throw submitError
+          }, asyncErrors => {
```
andreme: asyncValidate can now also throw a SubmissionError

src/handleSubmit.js
```diff
@@ -9,88 +10,86 @@ const handleSubmit = (submit, props, valid, asyncValidate, fields) => {
...
+          return Promise.reject(errors)
+        } else {
+          return errors
+        }
+      } else {
+        if (!onSubmitFail) {
+          return Promise.reject(submitError)
+        }
+      }
+    })
```
andreme: These are the current results:
- If there are errors from asyncValidate: reject the promise with the errors
- If there are errors from validate or SubmissionError: resolve the promise with the errors
- In case of another error: reject the promise with the error, if onSubmitFail was not provided

I think it should be this way:
- If there are errors from asyncValidate, validate or SubmissionError: resolve the promise with the errors
- In case of another error: reject the promise with the error, if onSubmitFail was not provided

The reason for resolving the promise in case of errors is that rejecting would result in unhandled rejection warnings. Also, the errors have been stored in redux by stopSubmit and I would consider that handled.

src/__tests__/handleSubmit.spec.js
```diff
@@ -16,18 +16,19 @@ describe('handleSubmit', () => {
...
-    expect(stopSubmit).toNotHaveBeenCalled()
-    expect(touch)
-      .toHaveBeenCalled()
-      .toHaveBeenCalledWith('foo', 'baz')
-    expect(asyncValidate).toNotHaveBeenCalled()
-    expect(setSubmitSucceeded).toNotHaveBeenCalled()
-    expect(setSubmitFailed)
-      .toHaveBeenCalled()
-      .toHaveBeenCalledWith('foo', 'baz')
+      .then(() => {
```
andreme: most of the changes in the tests are just whitespace, because I had to move the expect's into then()

andreme:
I will create a test to bring the coverage back to 100% if you want me to progress with this PR.

package.json
erikras: What benefit does it provide?

erikras:
This is a pretty big breaking change that will upset many users.

I largely agree with your reasoning, but I think we should set these ideas aside for now until a larger breaking change is called for.

> I understand if you will not merge this in, I can rework it to simply add the 3rd param.

Yes, do that for now. Maybe in another PR to preserve your `handleSubmit` work here.

package.json
andreme: Do you already use another promise library? Or do you mean the native Promises? In that case the users would need the poly fill to make sure it always works?

Issue #2069, opened 2016-11-03
Do we need getFormSubmitErrors ?


rndm2:
I need to get submitErrors, and I think it should be created new selector getFormSubmitErrors like getFormSyncErrors. Isn't it?

bmv437:
If you think it's a useful feature, then I would suggest you come up with an implementation and create a PR 😄

rndm2:
Ok, I will

dpwrussell:
I think this could probably be closed.

gustavohenke:
Yup @dpwrussell!

PR #2072, opened 2016-11-03
Added getFormSubmitErrors selector

rndm2:
See https://github.com/erikras/redux-form/issues/2069


erikras:

mmahalwy:
Doesn't this almost feel like it should be passed as a prop to the form? I find it painful to have to include `react-redux`, set everything up just to get this one field.

PR #2077, opened 2016-11-04
Document arguments passed to onSubmit function

MilllerTime:
Documented the three arguments passed to `onSubmit`. The third `props` argument [was added recently](https://github.com/erikras/redux-form/releases/tag/v6.0.2), but still not documented. I found it useful to know!


erikras:
Thanks!

PR #2083, opened 2016-11-04
Revert "fix checkox behavior"

### LINKED ISSUE #1993: [6.1.1] False boolean values are being filtered out of values
Any false value that gets passed as a payload gets filtered out and never gets set in the form `values` hash.

erikras:
This change caused problems. Fixes #1993.


erikras:

PR #2091, opened 2016-11-07
Values passed to the sync validation function must be defined

### LINKED ISSUE #2090: [6.2.0] Initial data passed to sync validate is undefined
The changes in https://github.com/erikras/redux-form/pull/1992 introduced a subtle change in the data that is passed to the sync validate function.

I have a form that uses ``enableReinitialize`` and sync validation. It also receives ``initialValues`` when connecting the component.

Simplified code example:

```typescript
const validate = (values, props) => { /* Some validation rules */ };

const EditorForm = reduxForm({
  form: 'userForm',
  enableReinitialize: true,
  validate,
  asyncValidate
})(UserEditor);

export default connect<StateProps, DispatchProps, OwnProps>(
  (state: ApplicationState, ownProps: OwnProps) => ({
    initialValues: Object.assign({}, state.authorisation.userDetails, { companyId: ownProps.companyId })
  }),
  {}
)(EditorForm);
```

When the form is mounted initially, it also calls the validate function. Now, with version 6.1.1, the ``values`` argument contained the value provided by ``initialValues``. In 6.2.0 however, this value is now ``undefined``, causing a runtime error in my application.

I've pinpointed the cause to this specific line: [src/reduxForm.js#L475](https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L475)

In my case the variable ``shouldUpdateInitialValues`` is set to ``true`` because I'm using ``enableReinitialize`` and ``initialValues`` differs from ``stateInitial`` (this value is undefined).

So in this case, ``initial`` is updated with ``stateInitial``, therefore becoming ``undefined``. This value is then passed to the ``validate`` function, causing an error.

A fix on my end is trivial, I could just assign a default value in the destructuring, but I'm wondering whether this is the intended behaviour, or that a subtle bug was just introduced? 

I suppose from an API-point-of-view it perhaps seems more sensible to always pass at least an empty object as values to the ``validate`` function, in order to prevent these kind of weird runtime errors. As far as I can tell, this was how it worked in the previous version :)

dmeenhuis:
Closes #2090

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -2441,8 +2441,8 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
         enableReinitialize: true,
         initialValues: { foo: 'bar' },
         validate: (values) => {
-          expect(values).toExist();
-          return {};
+          expect(values).toExist()
+          return {}
         }
       })(Form)
 
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -669,7 +669,7 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
       expect(formRender.calls.length).toBe(1)
 
       expect(inputRender).toHaveBeenCalled()
-      expect(inputRender.calls.length).toBe(1)      
+      expect(inputRender.calls.length).toBe(1)
 
       // Expect that input value has been initialized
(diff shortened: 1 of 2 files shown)


antoinerousseau:
we need this! :)

eliseumds:
Oh yeah, boy. Tks

erikras:
Nice catch, @dmeenhuis! Thanks.

erikras:

PR #2110, opened 2016-11-10
Added @@ infront of action names

liamhubers:
Following the convention of adding @@ in front of action names


DanielRuf:
@liamhubers can you point me to the convention? At least I find nothing about it http://redux.js.org/docs/basics/Actions.html

liamhubers:
I had a look and can't find it written anywhere, but people seem to be doing it anyway:
https://github.com/reactjs/redux/blob/master/src/createStore.js#L11
https://github.com/reactjs/react-router-redux/blob/a304bda820ba57f68961035eff1950e618047b81/src/reducer.js#L5

It's convenient me because I need to filter out all third-party redux actions

DanielRuf:
Ah ok. Yes,s aw it when I printed the triggered actions and then saw the double `@`. Good to know (y)

erikras:
This _could_ be a breaking change if anyone was dumb enough (or just n00b enough) to listen for one of these actions via an `action.type === 'stringvalue'`.

erikras:

PR #2113, opened 2016-11-10
Stop unregistering fields if `destroyOnUnmount` is false & always pass `keepDirtyOnReinitialize`as second argument to initialize action

ncphillips:


bbenezech:
Just tried it, `keepDirtyOnReinitialize` finally works for real. Thanks @ncphillips

erikras:
Great PR, @ncphillips.

Thanks for the confirmation, @bbenezech. That helps me a lot.

ncphillips:
Thank you guys @erikras and @bbenezech. This is a great library, and this was the one serious problem I've had with it so far.

erikras:

PR #2117, opened 2016-11-11
Clarify Intent of Tests Surrounding Several Props

ncphillips:
These props are:

* dirty
* pristine
* valid
* invalid

I also introduced a helper function called `propsAtNthRender` which replaced a particular pattern of function calls throughout the entire file. I believe this refactor greatly improves the readability of the tests it touches.


erikras:
Looks good. Thanks!

PR #2119, opened 2016-11-12
remove errors caused by using {...rest} and add DateTimePicker example

dagda1:
The react-widgets example was erroring when using `{...rest}` like this:
```
const renderMultiselect = ({ input, ...rest }) =>
  <Multiselect {...input}
    onBlur={() => input.onBlur()}
    value={input.value || []} // requires value to be an array
    {...rest}/>
```

With this error:

> warning.js:36 Warning: Unknown prop `meta` on <div> tag. Remove this prop from the element. For details

I removed rest and used explicit parameters.

I am not convinced this is the best way, the other way would be to do something like this:
```
const colors = [ { color: 'Red', value: 'ff0000' },
  { color: 'Green', value: '00ff00' },
  { color: 'Blue', value: '0000ff' } ]

/* eslint-disable no-unused-vars */
const renderDropdownList = ({ input, meta, ...rest }) =>
  <DropdownList {...input} {...rest} />

const renderMultiselect = ({ input, meta, ...rest }) =>{
  const value = input.value

  delete input.value

  return (
    <Multiselect {...input}
                 {...rest}
                 value={value || null}/>
  )
}

const renderSelectList = ({ input, meta, ...rest }) =>
  <SelectList {...input} {...rest} />
```

But if you were seeing these examples for the first time, I don't like the `renderMultiselect` above even though this what I used in my own project.

the eslint-disable comment is also confusing so I am not sold on this either.

Any feedback appreciated.

I also added a `DateTimePicker` example also:

```
const renderDateTimePicker = ({ input: { onChange, value }, showTime }) =>
  <DateTimePicker
    onChange={onChange}
    format="DD MMM YYYY"
    time={showTime}
    value={value && new Date(value) || null}
  />

```

### DIFF
--- a/examples/react-widgets/package.json
+++ b/examples/react-widgets/package.json
@@ -12,6 +12,7 @@
     "html-loader": "^0.4.3",
     "json-loader": "^0.5.4",
     "markdown-loader": "^0.1.7",
+    "moment": "^2.16.0",
     "raw-loader": "^0.5.1",
     "react": "^15.3.1",
     "react-dom": "^15.3.1",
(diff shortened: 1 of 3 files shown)


erikras:

PR #2126, opened 2016-11-16
Export registerField and unregisterField actions

mgh:


erikras:
These were kept private for a reason. Why do you want access to them?

mgh:
I'm dynamically generating forms with some data in them by dispatching a sequence of `initialize` and `change` actions before (potentially) mounting the form component itself (with `keepDirtyOnReinitialize: true`).  This worked fine before v6, but now it is no longer possible since the new behavior of the form is to only preserve registered fields.  It seems like it is now not possible to use many of the action creators without access to `registerField`.

The docs say:

> redux-form exports all of its internal action creators, allowing you complete control to dispatch any action you wish

erikras:

PR #2134, opened 2016-11-17
Add original error to onSubmitFail if possible

andreme:
BC version :)


erikras:

Issue #375, opened 2015-12-03
Nested fields objects are mutating


iNikNik:
Hello there! I found some little bug - nested fileds are mutating when you change something:

```javascript

class MyForm extends PureComponent {
    // ... some stuff
    render() {
        const { fileds: { someField, mainProduct } } = this.props;
        return (
            <div>
                <input {...someField} />
                <MainProductInput fields={mainProduct} />
            </div> 
        );
    }
}

class MainProductInput extends PureComponent {
    // ... some stuff
    render() { 
        // NEVER rerenders because fields objects always the same
        const { fileds: { title, other } } = this.props;
        return (
            <div>
                <input {...title} />
                <input {...other } />
            </div> 
        );
    }
}

reduxForm({
  form: 'mySuperForm',
  fields: [ 'someField', 'mainProduct.title',  'mainProduct.other' ],
})(MyForm)
```

I mean if you pass down whole nested field object - your children component will never rerenders.
I believe we should map `myFields` values and create a new objects for all nested fields here ->
https://github.com/erikras/redux-form/blob/master/src/readFields.js#L25

```javascript
function cloneFields(fields) {
   // mapValues like lodash.mapValues
  return mapValues(fields, field => isNested(field) ? cloneFields(field) : field);
}
const fieldObjects = cloneFields(myFields);
```

**PS guys, I want to tell you really big thanks! You're doing nice job with redux-form)**

iNikNik:
If I'm doing like this:
```javascript
render() {
        const { fileds: { someField, mainProduct } } = this.props;
        return (
            <div>
                <input {...someField} />
                <MainProductInput fields={ {...mainProduct} } />
            </div> 
        );
    }
```
Everything works perfect.

erikras:
Hmm... It is supposed to be creating new instances for any part of the subtree that changes.

Will investigate... :mag:

aanchalgera:
We are facing the exact issue

erikras:
Is this still an issue in v4?

iNikNik:
@erikras I've just checked it - seem issue still here

luisrudge:
@erikras anything I can do to help you figure this one out?

kristian-puccio:
Just found this as well ```
	componentWillReceiveProps(newProps) {
		const isEqual = newProps.fields.customer.email === this.props.fields.customer.email;
		console.log(isEqual, "....isEqual....");
	}
```

I'll try my hand at a pull-request now and see how I go :)

[...]

PR #2146, opened 2016-11-19
Fix arrays of object fields (v5)

asazernik:
Yet another instance of issue #375.

Specifically, any array that contained non-immediate descendants (e.g.
arrayField[].deepObjectField, or arrayField[][], but not plain old
arrayField[]) would remain the same instance if any of its descendant
fields were updated but no items were added to or removed from the
top-level array.

The issue was that we were only making a fresh copy of the array (that's
`changed = true`) in the case where it a raw-value array item was
changed, but not in the general case.

### DIFF
--- a/src/__tests__/readField.spec.js
+++ b/src/__tests__/readField.spec.js
@@ -708,6 +708,62 @@ describe('readField', () => {
     });
   });
 
+  it('should change complex field instances when updating a mixed field', () => {
+    const fields = {};
+    readField({
+      pig: {
+        foo: [
+          {
+            dog: {
+              value: 'hello'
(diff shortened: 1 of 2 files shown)


asazernik:
Now that this is fixed, I've found a related bug where even if the array's identity changes, the identities of the objects it contains does not. I have a broken test in my local checkout, and am working on a fix.

asazernik:
Updated the original commit with a more complete fix.

asazernik:
Heh. Did not realize there was a lint task I could run. Fixed the CI checks.

erikras:

PR #2151, opened 2016-11-21
Persist warnings through the initialize() action.

### LINKED ISSUE #2142: [6.2.0] initialize(currentFormData) clears warnings and errors until the next onChange
Ok, so I've tracked this bug all the way down to the cause, I just don't have a solution. I'd be happy to submit a PR if we can figure out a solution.

A little background:
in `src/reduxForm.js`, `warnIfNeeded()` will always fire `updateSyncWarningsIfNeeded()` on initialRender (`nextProps` === `undefined`), but on subsequent renders, it will only fire the updates if `!deepEqual(values, nextProps.values)`

see the following copy of `warnIfNeeded`

```javascript
warnIfNeeded(nextProps) {
  const { warn, values } = this.props
  if (warn) {
    if (nextProps) {
      // not initial render
      if (!deepEqual(values, nextProps.values)) {
        const { _warning, ...nextSyncWarnings } = warn(nextProps.values, nextProps)
        this.updateSyncWarningsIfNeeded(nextSyncWarnings, _warning)
      }
    } else {
      // initial render
      const { _warning, ...nextSyncWarnings } = warn(values, this.props)
      this.updateSyncWarningsIfNeeded(nextSyncWarnings, _warning)
    }
  }
}
```

this makes sense for most cases, but what if you initialize the form using the current values of the form (in order to get a pristine state)?

```javascript
const warn = (values) => {
  let warnings = {};
  if (!values.message || !values.subject) {
    warnings._warning = "missing required fields";
  }
  return warnings;
}

class MyFormComponent extends Component {
  handleSave = (data) => {
    this.props.saveData(data);
    this.props.initialize(data);
  }

  handleSend = (data) => {
    this.props.sendData(data);
  }

  render() {
    const {pristine, warning, handleSubmit} = this.props;
    return (
      <div>
        <Field name='subject' component='text' />
        <Field name='message' component='text' />
        <button type='button' 
                disabled={pristine} 
                onClick={handleSubmit(this.handleSave)}>
          Save Draft
        </button>
        <button type='button' 
                disabled={!!warning} 
                onClick={handleSubmit(this.handleSend})>
          Send
        </button>
      </div>
    );
  }
}

const MyForm = reduxForm({
  form: 'myForm',
  warn
})(MyFormComponent);

```

Since the action handler for the `initialize` action starts from an [empty state](https://github.com/erikras/redux-form/blob/master/src/reducer.js#L185), warnings get cleared from the state. and since it's not initialRender, and the [values of the form haven't changed](https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L194), it won't re-calculate warnings.

We need a way to the form to recalculate warnings even if the form has been initialized to it's current values in order to achieve a pristine state.

Technically there is a similar issue with sync errors and the `validateIfNeeded()` function, but you can't fire a submit on a form with errors anyway (hence why I'm using warnings).

I'd love to get some discussion going around a solution, and like I said, I'd be happy to submit a PR with the final fix.

bmv437:
fixes #2142

### DIFF
--- a/src/__tests__/reducer.initialize.spec.js
+++ b/src/__tests__/reducer.initialize.spec.js
@@ -262,6 +262,44 @@ const describeInitialize = (reducer, expect, { fromJS }) => () => {
         }
       })
   })
+
+  it('should persist warnings if they exist', () => {
+    const state = reducer(fromJS({
+      foo: {
+        registeredFields: [
+          { name: 'myField', type: 'Field' }
+        ],
+        values: {
+          myField: 'newValue'
+        },
+        initial: {
+          myField: 'initialValue'
+        },
+        warning: 'form wide warning',
+        syncWarnings: {
+          myField: 'field warning'
+        }
+      }
+    }), initialize('foo', { myField: 'newValue' }, true))
+    expect(state)
+      .toEqualMap({
+        foo: {
+          registeredFields: [
+            { name: 'myField', type: 'Field' }
+          ],
+          values: {
+            myField: 'newValue'
+          },
+          initial: {
+            myField: 'newValue'
+          },
+          warning: 'form wide warning',
+          syncWarnings: {
+            myField: 'field warning'
+          }
+        }
+      })
+  })
 }
 
 export default describeInitialize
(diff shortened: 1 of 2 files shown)


erikras:

PR #2154, opened 2016-11-21
a Fields component within FormSection should not need to know the FormSection name

### LINKED ISSUE #2148: FormSection should not add prefixName to its children component.
Greeting,

I was switching to 6.2.0, but I encountered an issue.
When I use FormSection with Fields with custom component, like the following:
`<FormSection name="hkaddress">
  <Fields component={Address} path="hkaddress" names={['province', 'city', 'county', 'address', 'postCode']} />
</FormSection>`

But in Address component, the props object contains {hkaddress: {province, city, county}}, so I have to pass a path to find the Field object.

> getPath() {
        let path = this.props.path;
        let arr = path.split('.');
        let parent = this.props;
        _.each(arr, (field)=>parent=parent[field]);
        console.log(parent);
        this.province = parent.province;
        this.city = parent.city;
        this.county = parent.county;
        this.address = parent.address;
        this.postCode = parent.postCode;
    }

> render() {
        this.getPath();
        //const { province, city, county, address, postCode } = this.props[this.props.path]
        let addressTitle = this.props.addressTitle;;
      return (
        <div className="form-group">
            <label className="control-label col-sm-1">{addressTitle}<span className="required">*</span></label>
            <div className="col-sm-1"><select className="form-control" {...this.province.input} onChange={this.handleProvinceChange.bind(this)} ><option value="">Province</option>{this.provinceOption()}</select></div>
            <div className="col-sm-1"><select className="form-control" {...this.city.input} onChange={this.handleCityChange.bind(this)} ><option value="">City</option>{this.cityOption(this.province)}</select></div>
            <div className="col-sm-1"><select className="form-control" {...this.county.input} ><option value="">County</option>{this.countyOption(this.province, this.city)}</select></div>
            <div className="col-sm-5"><input className="form-control" {...this.address.input} placeholder="地址" type="text" onBlur={this.validateAddress.bind(this)} /><p className="text-danger">{this.state.addressValidate}</p></div>
            <label className="control-label col-sm-1">Postcode<span className="required">*</span></label>
            <div className="col-sm-2"><input className="form-control" {...this.postCode.input} placeholder="Post code" type="text" maxLength="6" onBlur={this.validatePostCode.bind(this)} /><p className="text-danger">{this.state.postCodeValidate}</p></div>
        </div>
      )
    }

I think the Fields/Field should not care about the parent path, it's FormSection's responsibity to attach the values object to its parent object.

Please suggest the right way to use FormSection and Fields.

Thanks and Best Regards.

bmv437:
This fixes #2148,

Currently a `Fields` component within a `FormSection` needs to know what the FormSection prefix is in order to access the fields inside the `Fields` component. This PR makes it so that the form sectionPrefix doesn't carry over into the `Fields` component props. Also included is a test case.

```javascript
<FormSection name='parent'>
  <Fields name={['foo', 'bar']} component={renderFields} />
</FormSection>

...

// current
const renderFields = ({parent: {foo, bar}}) => (
  <div>
    <input {...foo} />
    <input {...bar} /> 
  </div>
)

// proposed
const renderFields = ({foo, bar}) => (
  <div>
    <input {...foo} />
    <input {...bar} /> 
  </div>
)

```

### DIFF
--- a/src/ConnectedFields.js
+++ b/src/ConnectedFields.js
@@ -49,7 +49,7 @@ const createConnectedFields = ({ deepEqual, getIn }) => {
 
     render() {
       const { component, withRef, _fields, _reduxForm, ...rest } = this.props
-      const { asyncValidate, blur, change, focus } = _reduxForm
+      const { asyncValidate, blur, change, focus, sectionPrefix } = _reduxForm
       const { custom, ...props } = Object.keys(_fields).reduce((accumulator, name) => {
         const connectedProps = _fields[ name ]
         const { custom, ...fieldProps } = createFieldProps(getIn,
@@ -64,7 +64,8 @@ const createConnectedFields = ({ deepEqual, getIn }) => {
           asyncValidate
         )
         accumulator.custom = custom
-        return plain.setIn(accumulator, name, fieldProps)
+        const fieldName = sectionPrefix ? name.replace(`${sectionPrefix}.`, '') : name
+        return plain.setIn(accumulator, fieldName, fieldProps)
       }, {})
       if (withRef) {
         props.ref = 'renderedComponent'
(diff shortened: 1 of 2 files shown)


erikras:
Looks good.

erikras:

PR #2170, opened 2016-11-27
Fix typos "aynchronous" -> "asynchronous"

oyeanuj:
Just noticed a little typo, small ways for us to give back :)


erikras:
Thanks!

Issue #2062, opened 2016-11-01
Can we not convert null to empty string?


sbunchridecell:
Can we not do this? https://github.com/erikras/redux-form/blob/b531cbeb72f97b85771845664ac68bf2da8360c7/src/createFieldProps.js#L53

When I pass null, I expect null, not an empty string.

martinhorsky:
I think it i intentionally behaviour.

As mentioned in https://github.com/facebook/react/releases/tag/v15.0.0:

> React 15 warns you on a null input value and offers you to clarify your intention. To fix the warning, you may explicitly pass an empty string to clear a controlled input, or pass undefined to make the input uncontrolled.

If you need null value for your own Field component, I suggest to use `parse` prop:

```javascript
<Field
  name="priority"
  component={NumberInput}
  format={(value) => (value === '' ? null : Number(value))}
  min={0}
  max={10}
/>
```

sbunchridecell:
Sure, thanks.

erikras:
Published in [`v6.2.1`](https://github.com/erikras/redux-form/releases/tag/v6.2.1).

PR #2171, opened 2016-11-27
Allow null to empty string conversion override using format

martinhorsky:
See #2062


erikras:
Ah, okay, so the answer to "Why would you want this?" is "Numbers." Okay.

martinhorsky:
Thank you Erik for merge. 

Yes. Exactly. Or for any other custom component where the `null` is real value.

erikras:

PR #2175, opened 2016-11-28
Rewrite setInPath for 2.5x speed improvement.

huan086:
See https://jsperf.com/plainsetin for benchmark. Makes a huge difference in real world due to much less stress on GC.


erikras:
Nice work, @huan086!

erikras:

PR #2187, opened 2016-11-29
[FieldArray] Pass reference to the `fields` prop in map, forEach, and reduce callbacks

Whoaa512:
It's probably not well know but creating functions inline within a render call is a performance hit since new functions are created on each re-render. This in turns clutters our memory space and makes the GC work unnecessarily.

To avoid this it would be helpful if `fields.map`, `fields.forEach`, and `fields.reduce` passed along a `fields` reference to the callback provided. This allows us to create a callback method that can be passed directly the `fields.map` while still allowing access to `swap`, `remove`, `pop`, etc.

Example:
```javascript
 const renderSubFields = (member, index, fields) => (
     <li key={index}>
       <button
         type="button"
         title="Remove Member"
         onClick={() => fields.remove(index)}/>
       <h4>Member #{index + 1}</h4>
       <Field
         name={`${member}.firstName`}
         type="text"
         component={renderField}
         label="First Name"/>
       <Field
         name={`${member}.lastName`}
         type="text"
         component={renderField}
         label="Last Name"/>
     </li>
 )
 const renderMembers = ({ fields }) => (
   <ul>
     <button type="button" onClick={() => fields.push({})}>Add Member</button>
     {fields.map(renderSubFields)}
   </ul>
 )
 ```

### DIFF
--- a/docs/api/FieldArray.md
+++ b/docs/api/FieldArray.md
@@ -196,3 +196,35 @@ be called three times, with `'foo.bar[0]'`, `'foo.bar[1]'`, and `'foo.bar[2]'`.
 #### `index : Number`
 
 > The index of the item in the array.
+
+#### `fields : Object`
+
+> A reference to the [`fields` prop](#fields-props) to allow for the access to `swap`, `remove`,
+`pop`, etc., without requiring closure scoping.
+```javascript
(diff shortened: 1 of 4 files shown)


erikras:
Nicely done!

erikras:

PR #2218, opened 2016-12-04
Only wrap FormSection in a div if necessary

SimenB:
I don't want extra `div`s in my DOM 😄

### DIFF
--- a/src/FormSection.js
+++ b/src/FormSection.js
@@ -20,7 +20,13 @@ class FormSection extends Component {
   }
 
   render() {
-    return <div>{this.props.children}</div>
+    const { children } = this.props
+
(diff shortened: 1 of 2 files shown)


erikras:
Nice one. 💥 those semicolons and I'll merge it.

SimenB:
@erikras Updated 😄

SimenB:
I'm thinking it might be better to do `React.Children.only` to force consumer to only pass a single element. Can be a dumb `div` if they want, or do something more, like add classes, animate it etc. That's a breaking change though

erikras:

SimenB:
My last comment was solved beautifully with #2329

PR #2219, opened 2016-12-05
Change setIn to work with seamless-immutable. Fixes #2204.

### LINKED ISSUE #2204: Regression: v6.2.1 breaks seamless immutable
Not sure if it's redux-form or seamless, but I opened this issue there: 

https://github.com/rtfeldman/seamless-immutable/issues/175

The issue stems from the setInWithPath performance patch, specifically changing to arr.slice() instead of [...arr]: 

https://github.com/erikras/redux-form/pull/2175/files#diff-ba79b920ecda0b13b45785ad078f6958L14

This will likely break any redux-form installations using seamless-immutable. For now, we've pegged our redux-form to 6.2.0, but any installs that are loose on the patch version will be affected when they npm/yarn up.

huan086:


erikras:
Thanks!

erikras:
Published in [`v6.3.0`](https://github.com/erikras/redux-form/releases/tag/v6.3.0).

Issue #2215, opened 2016-12-03
FieldArray `fields` prop should have `get` or `index` method to get value


dcurletti:
It would be great if the `fields` props of `FieldArray` had an instance method `get` or `index` that would allow its children to get their value.  This is for scenarios where the `FieldArray` items are created with an initialValue, and that initial value needs to be represented in the UI

Example: 
```
<div>
{fields.map((group, i) => (
	<div key={i}>
		<span>Display Name: {fields.get(i).displayName}</span>
		<Field
				name={`${group}.field_type`}
				component="input"/>
		<Field
				name={`${group}.values`}
				component="input"/>
	</div>
))}
</div>
```

I can't use the `fieldSelector` due to the fact that the field's array is dynamically created and using a `FormSection` instead would result in keeping track of the fields in two different stores.

dcurletti:
I'll see if I can tackle this in the coming days (no promises).  I find it to be somewhat essential for highly dynamic, nested fields.

erikras:
Published fix in [`v6.3.1`](https://github.com/erikras/redux-form/releases/tag/v6.3.1).

PR #2233, opened 2016-12-07
Update createFieldArrayProps.js

jeremy-carbonne:
Fix issue #2215 fields.getAll()[index]


AlexGrs:
We need this feature to access all the values in a FieldArray. It solves for us a display of those values in our fields instead of just having access to the name.

erikras:
This implementation is insufficient, as it does not respect ImmutableJS implementations. I'll make a new PR.

PR #2245, opened 2016-12-08
Field-level Validation

erikras:
Allows for things like:

```jsx
const required = value => value == null ? 'Required' : undefined

<Field name="firstName" component="input" validate={required}/>
```


niksajanjic:
Why is **validate** prop passed down to Field's component? Let's see this example:

```
<Field
  name="email"
  component={Input}
  validate={[isRequired, isEmail]}
  normalize={value => `%${value}%`}
  format={value => (value ? value.slice(1, -1) : '')}
/>
```

Now as a props of Input field I get **name**, **input**, **meta** and **validate**. Why si **validate** passed down? As a matter of fact I don't understand why is **name** passed down either as it's part of **input** object anyways. On the other hand **component**, **format** and **normalize** are not passed down which makes sense.

The reason why I'm asking is because people usually do something like this in their Input component:

```
const Input = (input, meta, ...rest) => {
  return (
      <input {...input} {...rest} />
  );
};
```

This alows them to be able to pass down placeholder, type, maxLength, className and any other valid input attribute. React 15 will have a warning if they pass something that doesn't belong to input tag:

```
<Field
  name="email"
  component={Input}
  validate={[isRequired, isEmail]}
  normalize={value => `%${value}%`}
  format={value => (value ? value.slice(1, -1) : '')}
  placeholder="Input placeholder"
  type="text"
  maxLength="30"
  className="myInput"
/>
```

With this new version we need to expect **validate** and make sure we don't assign it to input field. Maybe it doesn't make sense to pass down any of these props that Field component expects to be passed to it? I personally use ...rest spread operator as I don't want to expect every single valid input attribute inside Iput component.

PR #2246, opened 2016-12-08
Allowed submitting when submit errors exist

### LINKED ISSUE #2244: Submit validation doesn't trigger
Here example code https://jsbin.com/guxoqiqevo/edit?js,console,output

Code above should trigger validation every time when "Push me" button is clicked. But actual result is that validation triggers only once. Just open dev tools and look to output. Every time validation triggers `console.log()` should be handled (but it doesn't).

What's wrong with this? I use this code in production, so I would appreciate for your asap response. 

P.S. Yes, I need to call `handleSubmit` function outside of render method (but actually it doesn't matter). It's my an informed choice.

Thx

erikras:
Fixes #2244 and addresses a comment left by @orfjackal in `handleSubmit.js`.

The problem was that it should still try to submit even if previous submission errors exist. Now it will only not submit if sync or async errors exist.

### DIFF
--- a/src/selectors/isValid.js
+++ b/src/selectors/isValid.js
@@ -3,17 +3,25 @@ import createHasError from '../hasError'
 const createIsValid = structure => {
   const { getIn } = structure
   const hasError = createHasError(structure)
-  return (form, getFormState = state => getIn(state, 'form')) =>
+  return (form, getFormState, ignoreSubmitErrors) =>
     state => {
       const formState = getFormState(state)
       const syncError = getIn(formState, `${form}.syncError`)
       if (syncError) {
         return false
       }
+      if (!ignoreSubmitErrors) {
+        const error = getIn(formState, `${form}.error`)
+        if (error) {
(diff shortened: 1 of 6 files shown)


erikras:

src/selectors/isValid.js
```diff
@@ -3,17 +3,25 @@ import createHasError from '../hasError'
 const createIsValid = structure => {
   const { getIn } = structure
   const hasError = createHasError(structure)
-  return (form, getFormState = state => getIn(state, 'form')) =>
+  return (form, getFormState, ignoreSubmitErrors) =>
```
diegohaz: Hey, @erikras. I think this introduced a breaking change. Now, when calling `isValid('form')` like I did before (because `getFormState` was optional) it throws `TypeError: getFormState is not a function`

diegohaz: Oh, I just realized that there's already an open issue #2255

PR #2247, opened 2016-12-08
Implemented fields.get()

### LINKED ISSUE #2215: FieldArray `fields` prop should have `get` or `index` method to get value

erikras:
Fixes #2215. Closes #2233.


dcurletti:
@erikras this is **HUGE**.  Thank you so much!

erikras:

PR #2270, opened 2016-12-12
restore default value for getFormState removed in d84aa7

adam187:
Hi,

This change https://github.com/erikras/redux-form/commit/d84aa7f055f54ed7a7e0c1eabc7b5b1ec769d331 introduced regression in using `isValid` as selection http://redux-form.com/6.3.1/docs/api/Selectors.md/

Trying to use:
`valid: isValid('myForm')(state)` 
will cause `TypeError: getFormState is not a function`


erikras:
Published in [`v6.4.1`](https://github.com/erikras/redux-form/releases/tag/v6.4.1).

PR #2271, opened 2016-12-12
Allow style forwarding for FormSection element

chris-verclytte:
It is particularly useful when using flex properties.


erikras:
Fix published in [`v6.4.0`](https://github.com/erikras/redux-form/releases/tag/v6.4.0).

PR #2274, opened 2016-12-13
Add FormSection styling via className

martynovs:


cdaringe:
very similar to #2271. perhaps a more "hands-off-make-everybody-happy" is a props arg that does a `{...props}` on the `<div ... />`, so you can use `{ id: ..., className: ..., style: ...}`.

one-off'ing property support probably isn't a super scalable concise approach?  just thinkin out loud :)

erikras:
It seems like what really needs to happen here is that it should pass along all props that are not part of the API.

erikras:
Fix published in [`v6.4.0`](https://github.com/erikras/redux-form/releases/tag/v6.4.0).

PR #2289, opened 2016-12-14
Add selectors for submitting, submitSucceeded and submitFailed

### LINKED ISSUE #2173: [Feature Request] Include 'submitting' and 'submitSucceeded' in the selectors
To implement a solution discussed in #2169 (@erikras's 'Many little forms' solution) cleanly, I need to manage logic in the parent container based on whether each of the little forms (auto-saving each field) are submitting or have submitted, etc. 

Looking through the documentation in the [API](http://redux-form.com/6.2.0/docs/api/Selectors.md/), it seems that while one can check for errors or state of a field, there is no selector giving access to see if form is `submitting`, has `submitSucceeded` or `submitFailed`. 

In simple cases, one can access the state directly through custom selectors. But for sake of consistency as well as for more complex cases, it might be worthwhile having the following selectors:

1. `isSubmitting`: Corresponds to the `submitting` prop.
2. `hasSubmitSucceeded`: Corresponds to the `submitSucceeded` prop.
3. `hasSubmitFailed`: Corresponds to the `submitFailed` prop.

Not sure if there are more complex reasons for not having those selectors, but I would love to understand more!

dpwrussell:
Add selectors, tests and documentation for `submitting`, `submitSucceeded` and `submitFailed`.

Resolves #2173


erikras:
Thanks!

erikras:

tstirrat15:
Were these ever added to exports? I'm getting import errors, and looking at `index.js`, I don't think they were.

PR #2296, opened 2016-12-15
First call of validate always receives undefined/empty object when enableReinitialize:true

### LINKED ISSUE #1901: enableReinitialize: true breaks validation
Given a simple redux-form config like:

```
const form = reduxForm({
  form: 'ProfilesForm',
  enableReinitialize: true,
  validate
})(ProfilesForm);
```

with a validate function like:

```
const validate = values => {
  const errors = {};

  if (!values.name) {
    errors.name = 'Profile name is required.'
  }
  return errors;
};
```

the `validate` function runs twice on load and leads to an errors object:

```
{ name: 'Profile name is required.' }
```

But no errors show. What I would like to do is show an error when that field is blurred or the form is submitted. Removing `enableReinitialize: true` accomplishes this, but I need `enabledReinitialize` because the form is being populated by server data. 

Is this expected?

k-italy:
Resolves #1901 

Starting from changes in #1992, the first run of validate function receives undefined (in v6.2.0) or empty object(since v6.2.1) if ```enableReinitialize``` option is true.

This caused "Cannot read property 'xxx' of undefined" kinda errors in implementation based on v6.1.1 or before.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -525,8 +525,9 @@ const createReduxForm =
             const { form, getFormState, initialValues, enableReinitialize, keepDirtyOnReinitialize } = props
             const formState = getIn(getFormState(state) || empty, form) || empty
             const stateInitial = getIn(formState, 'initial')
+            const initialized = !!stateInitial;
 
-            const shouldUpdateInitialValues = enableReinitialize && !deepEqual(initialValues, stateInitial)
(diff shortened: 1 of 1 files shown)


k-italy:
Oops, I fixed the lint error.

k-italy:
I've just found #2090, which explains the empty starting at v6.2.1. 
But I still think original initialValues should be the one that's passed to validate function initially.

erikras:

Issue #2231, opened 2016-12-06
Wizard Form Live example in the documentation is not working properly


KamilHavlicek:
Live example of http://redux-form.com/6.2.1/examples/wizard/ is most likely not working properly.

![reduxform7](https://cloud.githubusercontent.com/assets/6735567/20948547/fe7f339a-bc14-11e6-9fc5-7b18c8d2f51d.gif)

To reproduce the bug, please follow these "test steps":
1) Enter John into the First Name field
2) Enter Doe into the Last Name field
3) Press Next button (to get to the second step)
4) Click on Previous button (to get to the first step)
5) Click on Next button -> and nothing is happening. Expected behaviour: It should go to the Step 2, but it stays on the first form step.

Interestingly, example http://redux-form.com/6.2.0/examples/wizard/ works, but http://redux-form.com/6.2.1/examples/wizard/ does not.

hamczu:
I have just encountered the same issue - I can confirm it's not related to example code, but it just stopped working. It looks like there's a regression.

kristofferh:
Fields no longer get unregistered when `destroyOnUnmount=false` (but they do get registered) so when you advance to the second screen and then go back it'll try to validate against fields that are no longer visible. I'm not exactly sure what use case #2113 was solving (there's no description in the pull request), but it definitely breaks the multi-form/wizard example.

erikras:
Fix published in [`v6.4.0`](https://github.com/erikras/redux-form/releases/tag/v6.4.0).

PR #2310, opened 2016-12-18
Force unregister fields

kristofferh:
The [wizard form](http://redux-form.com/6.3.1/examples/wizard/) is currently broken because of a change in 6.2.1 (specific PR here: https://github.com/erikras/redux-form/pull/2113). Bug report, with repro-steps: https://github.com/erikras/redux-form/issues/2231. 

When `destroyOnUnmount` is set to `false` it will no-longer unregister fields as components unmount, but it will keep the store's data, and so it will try to validate against fields that have, potentially, been unmounted and are not visible to the user. This PR introduces a new `prop` that will let you force the unregistration of fields. I was hesitant to add another prop, but I can't see a cleaner way of doing it, without re-introducing the bug that #2113 was trying to solve.

I updated the docs and the wizard form example as well.

### DIFF
--- a/docs/api/ReduxForm.md
+++ b/docs/api/ReduxForm.md
@@ -58,6 +58,10 @@ component is unmounted. Defaults to `true`.
 Defaults to `false`.  If the `keepDirtyOnReinitialize` option is also set, the form
 will retain the value of dirty fields when reinitializing.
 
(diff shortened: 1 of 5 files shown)


erikras:
Nice one! Thanks!

erikras:

PR #2319, opened 2016-12-21
Wrong validation of calculating max length

burakdirin:
Wrong validation of calculating max length on example


erikras:
No, the code is correct.

burakdirin:
this code is correct ? :) 

`const minValue = min => value =>value && value < min ? 'Must be at least' + min : undefined`

check out please -> https://jsfiddle.net/0pgb1prk/

gustavohenke:
@burakdirin, the deal with this example is to validate that a number is less than `min`. It's not related to string length.

burakdirin:
Now I understand much more clearly, sorry :)

Issue #2323, opened 2016-12-21
Exception thrown if server returned submit error for fieldArray


Anarios:
If I have a form with field array and if I set error to the field array an exception will be thrown every time I change any field in the array. I've set up a test that fails in this situation.

erikras:
Nice reporting and failing test PR. 👍 Will investigate. 🔍

Anarios:
It's tied to how deleteInWithCleanUp works. Both lodash verison for immutable and self-written for plain fail when they try to delete a path for array when there is actually a string.

PR #2324, opened 2016-12-21
Exception thrown if server returned submit error for fieldArray

Anarios:
If I have a form with field array and if I set error to the field array an exception will be thrown every time I change any field in the array.  This is a test that fails in such a situation. [issue](https://github.com/erikras/redux-form/issues/2323)

### DIFF
--- a/src/__tests__/reducer.change.spec.js
+++ b/src/__tests__/reducer.change.spec.js
@@ -212,6 +212,32 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
         }
       })
   })
+  
+  it('should set value on change and remove field-level submit and async errors', () => {
+    const state = reducer(fromJS({
+      foo: {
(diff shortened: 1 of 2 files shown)


erikras:
This wasn't really a bug, I don't think. I think the problem was that you were not setting the error properly. Look at the [Field Arrays Example](http://redux-form.com/6.4.2/examples/fieldArrays/) and search the page for `_error`. See how I changed your test: https://github.com/erikras/redux-form/commit/98ddff24fdb2b4695455777db894120b6ee0c5b6

Anarios:
@erikras, yes I know that this wasn't a proper way to set error message, I just thought that it might be wrong that we receive an exception and a non-working form after it.

It was happening this way because of server-side validation which sets errors to fiedls and to array-fields in a same way. (which seems logical).

PR #2326, opened 2016-12-22
Created Form component

### LINKED ISSUE #2238: Remote submit breaks when submit function passed to handleSubmit
I've been messing around with [remote submit](http://redux-form.com/6.2.0/examples/remoteSubmit/) lately and it works **beautifully** whenever I pass my submit function to `reduxForm()` as a config value. However, if I instead supply it to my `handleSubmit()` function on the form, my app breaks with the following error:
```
You must either pass handleSubmit() an onSubmit function or pass onSubmit as a prop
```
Note that I am using `redux-form/immutable` and that I'm using stuff like `bindActionCreators` so maybe there is some weird combination there or maybe my implementation is simply wrong, but this seems like a bug to me.

I'm using:
redux-form v6.2.1
redux v3.6.0
react v15.3.1

Here's a simplified version of my code:

**Wrapper.js**
```
import { bindActionCreators } from 'redux';
import { connect } from 'react-redux';
import { submit } from 'redux-form/immutable';

const mapDispatchToProps = function (dispatch) {
  return bindActionCreators({
    submit: submit
  }, dispatch);
};

class Wrapper extends Component {
  render() {
    return (
      <MyForm />
      <SubmitButton submit={submit} />
    );
}

export default connect(mapStateToProps, mapDispatchToProps)(Wrapper);
```

**MyForm.js**
```
import validate from './validate';

class MyForm extends Component {
  mySubmitFunction = (values) => {
    console.log('submitted!');
  }

  render() {
    const { handleSubmit } = this.props;

    return (
      <form onSubmit ={handleSubmit(this.mySubmitFunction)}>
        <Field ...fieldProps />
      </form>
    );
  }
}

export default reduxForm({ form: 'myForm', validate })(MyForm);
```

**SubmitButton.js**
```
export default class SubmitButton extends Component {
  submit = () => {
    this.props.submit('myForm');
  }

  render() {
    return <button onClick={this.submit}>Submit</button>
  }
}
```

Shouldn't this code work? I'm making my submit function local to the `MyForm` class here, but it still breaks if it is an external function. Once again, this works perfectly if I pass an external submit function to `reduxForm()` instead. Is this a bug?

erikras:
Fixes #2238. Based on https://github.com/erikras/redux-form/issues/2238#issuecomment-268641999 from @davidkpiano.

### DIFF
--- a/docs/api/Form.md
+++ b/docs/api/Form.md
@@ -1,0 +1,35 @@
+# `Form`
+
+The `Form` component is a simple wrapper for the React `<form>` component that allows
+the surrounding `redux-form`-decorated component to trigger its `onSubmit` function.
+
+It is only useful if you are:
+
+- performing your submission from inside your form component by passing
+  `onSubmit={this.props.handleSubmit(this.mySubmitFunction)}` to your `<form>`
+  component
+- **AND EITHER:**
+  - initiating your submission via the [`submit()` Instance API](http://redux-form.com/6.3.1/docs/api/ReduxForm.md/#-submit-promise-) (i.e. calling it directly on a reference to your decorated form component)
+  - initiating your submission by [dispatching the `submit(form)` action](http://redux-form.com/6.3.1/examples/remoteSubmit/)
+
+If you are passing in your `onSubmit` function as a config parameter or prop, this component will do nothing for you.
+
+## Importing
+
+```javascript
+var Form = require('redux-form').Form;  // ES5
+```
+```javascript
+import { Form } from 'redux-form';  // ES6
+```
+## Props you can pass to `Form`
+
+Any that you can pass to `<form>`, but only one is required.
+
+### `onSubmit : Function` [required]
+
+> The function to call when form submission is triggered.
+
+## Usage
+
+All you do is replace your `<form>` with `<Form>`.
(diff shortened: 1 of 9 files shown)


davidkpiano:
Awesome! If you take this approach further in the future, you can potentially completely eliminate the need for decorators by just having everything be a prop of this `<Form>` component. Makes it much more composable in existing React applications.

erikras:
> If you take this approach further in the future, you can potentially completely eliminate the need for decorators by just having everything be a prop of this <Form> component.

I don't think so. You have to be able inject props into _something_, which requires a HOC.

davidkpiano:
Ah, gotcha - that's because the API is based on injected props for:

- Getting form-specific state
- Doing form actions such as `reset()` or `submit()`

Just curious, why not adopt this approach:
- Getting form-specific state - **`mapStateToProps`**
- Doing form actions such as `reset()` or `submit()` - **dispatch actions**

instead of having it ad-hoc?

erikras:
That is certainly a potential direction to move in. I used to want libs that did everything for me, but @gaearon's talk at React Europe 2016 about the value of what a lib does _not_ do for you was very enlightening.

erikras:

PR #2329, opened 2016-12-23
Improved FormSection wrapper component

erikras:
When `FormSection` needs to wrap its children in a component, you may now:

* Specify the type of component, e.g. `div`, `section`, etc.
* Pass through props, e.g. `className`, `style`, to said component

Closes #2271 and #2274.


erikras:

Issue #1955, opened 2016-10-13
Example for React native - Immutable - Redux


axelardu:
Hi,

Is it possible to get an example for react-native with rexux & immutable with the latest (6.1.0) tag?
Did not find an example anywhere.

Thanks!

bitdivine:
I just went on a hunt for examples of redux-form as there are none in the examples section http://redux-form.com/6.1.1/examples/  This is what I found:

* Without immutable, this was enough to get me up and running: https://gist.github.com/namelos/34c042c7a6f88808088b17de258a34c4
* With immutable, this SO question very nearly has a working example - see the comments: https://stackoverflow.com/questions/39331420/cant-get-redux-form-6-0-2-working-with-react-native-0-32-0

HTH.

bitdivine:
By the way, I found http://beerpay.io yesterday.  Erik, you should sign up.  I'd buy you a beer.  It's not quite as cool as "Send money every time I comment :beer: on Github" but I guess it's close enough! :-)

mschipperheyn:
I use redux-form in react-native. Do you have any specific issues?

bitdivine:
Yes.  A lack of examples.  As a complete react novice this was a bit of a problem.  But that is now solved and I have my forms all wired up! :-)

JeroenNelen:
@mschipperheyn  and @bitdivine I'm currently bumping into these errors after submitting a form:

I switched to use the formReducer from redux-form/immutable, but when I then submit a value I get the following errors (I added the .toJS() to the values object that I receive in the handleSubmit function):
```
ExceptionsManager.js:62 One of the sources for assign has an enumerable key on the prototype chain. This is an edge case that we do not support. This error is a performance optimization and not spec compliant.handleException 
@ ExceptionsManager.js:62handleError 
@ InitializeJavaScriptAppEngine.js:120reportFatalError 
@ error-guard.js:30guard 
@ MessageQueue.js:47callFunctionReturnFlushedQueue 
@ MessageQueue.js:104onmessage 
@ debuggerWorker.js:44
ExceptionsManager.js:82 
[...]

Issue #1877, opened 2016-09-28
[v6] immutable FormReducer causes fatal exception: "One of the sources for assign has an enumerable key on the prototype chain. This is an edge case that we do not support. This error is a performance optimization and not spec compliant."


jpaas:
If I use the FormReducer from redux-form/immutable, the createFieldProps causes an exception when trying to Object.assign the meta props. Looks like its expecting state to be a js object instead of an immutable one.

axelardu:
Hey there,
Having the same problem with redux-form v6.1.0 .
Did you manage to get it fixed?

jpaas:
No I ended up having to fall back to using the non-immutable reducer

cayter:
Facing the same issue. Is this a bug or it's because of we're not using it correctly?

Prudwell:
Anyone found anything new on this? Same problem for me.

erikras:
Published fix in [`v6.4.2`](https://github.com/erikras/redux-form/releases/tag/v6.4.2).

PR #2336, opened 2016-12-26
Ensure state is merged into meta props as plain object (#1877, #1955)

esbenp:
cb3abb5b41976458fc0c54246df2eeea1f538d34 added functionality that merged the state into the meta object for fields. However, when using Immutable JS structures the state will be an Immutable.Iterable. When developing in React Native the `Object.assign` call here https://github.com/erikras/redux-form/commit/cb3abb5b41976458fc0c54246df2eeea1f538d34#diff-7eb826f425b082db29894569d83899f8R68 causes an exception as described in #1877 and #1955.

```
One of the sources for assign has an enumerable key on the prototype chain. This is an edge case that we do not support. This error is a performance optimization and not spec compliant.
```

I am assuming this is because the Object.assign polyfill is different in the React Native environment (any confirmation on this?). I tested the webpack polyfill in the Immutable example and it pretty much just transfers properties from one object to another without any checks. I am assuming this is why this error has not had any more attention since it does not happen in browser environments.

EDIT: Found the polyfill https://github.com/facebook/react-native/blob/master/packager/react-packager/src/Resolver/polyfills/polyfills.js#L56

As for the fix I added a `toJS` method to the structure types. This also means I had to expand the first argument in `createFieldProps` and `createFieldsProps` to be an object instead of just `getIn`. I am not familiar with the code base so any pointers and directions for style is much appreciated. It should be noted that I have not tested this solution in React Native environment (only a dirty hack that accomplishes the same). And I have not tested all the examples either but tests are passing.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -9,7 +9,7 @@ const propsToNotUpdateFor = [
   '_reduxForm'
 ]
 
-const createConnectedField = ({ deepEqual, getIn }) => {
+const createConnectedField = ({ deepEqual, getIn, toJS }) => {
 
   const getSyncError = (syncErrors, name) => {
     const error = plain.getIn(syncErrors, name)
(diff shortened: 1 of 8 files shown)


erikras:
Looks good.

PR #2346, opened 2016-12-27
Pass submitFailed prop down to Field components

### LINKED ISSUE #567: Pass submitFailed option down to fields
It would be nice if the `submitFailed` prop was passed down to the input fields so that we can have a bit more control over when validation messages are displayed. Right now my logic is `error && (touched || dirty)`, but ideally it would be `error && (submitFailed || dirty)`. This way, validation messages don't pop up just because a user tabs through a field or accidentally clicks outside the field, causing it to blur.

Anyway, thanks for the great library! I'd be happy to submit a pull request for this, but let me know what you think.

eligolding:
fixes #567


eligolding:
Should be trivial to add `submitSucceeded`, but I'm curious to know what the use case would be for that prop at the `Field` level. For `submitFailed` the obvious use case would be (as discussed in #567) deciding whether or not to show a validation error. On `submitSucceeded`, however, usually you would just swap out the form or transition away. Thoughts?

eligolding:
Thank you!

erikras:

PR #2161, opened 2016-11-23
Bind handlers to components

oreqizer:
Makes functions preserve their identity, thus not causing
unnecessary re-renders.

PR #2357, opened 2016-12-29
Pass in name to `createFieldProps`

SimenB:
It disappeared in https://github.com/erikras/redux-form/pull/2161/files#r94125255

It's absence breaks my code


erikras:
That was never in the docs. It was a bug that that was leaking through. It _is_ there as `input.name`.

SimenB:
Fair enough. Is my usage of it rare, or does it make sense to add a note to the changelog?

PR #2360, opened 2016-12-29
Add yarn lockfile

mistadikay:
I suggest to add a `yarn` lockfile to the project. It's going to be ignored when `redux-form` would be installed as a dependency, but it will make contributing experience much better. For details and reference please see: https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all


gustavohenke:
Something went wrong with the coverage here 😕

mistadikay:
Weird, basically tests fail when `NODE_ENV` set to `test` (which is happening in `test:cov` script). Investigating

mistadikay:
Hm, the same issue with coverage was in this commit: https://travis-ci.org/erikras/redux-form/jobs/187009142 You can notice it's failing if uncover `npm run test:cov`
screenshot 2016-12-29 17 14 31

I wonder if it's one of the updated dependencies messing everything up...

erikras:
I had a yarn lock file for a while, but yarn kept acting up and breaking the build in strange ways, either with linting or test coverage, so I removed it. I love the _idea_ of yarn, but it doesn't seem quite mature enough yet.

mistadikay:
Hm, you're right — if I use `npm install` instead of `yarn` — `npm run test:cov` does not throw anything. That's really odd. Do you mind if I investigate it a bit? If we'll be certain that nothing can be done here, this PR can be closed without merging. 😕

erikras:
Yes, by all means, please investigate, and if you solve it, re-submit. 👍

PR #2361, opened 2016-12-29
Cleaned up some prop leakage

### LINKED ISSUE #2356: parse is embedded in Field's custom prop
I think there is a regression with field's custom prop. In version 6.4.1, the "parse" property is passed to the wrapped component in custom prop. This causes error when spreading custom prop in, for example, a basic input. 
I just verified with 6.2.1 and parse was not passed into custom.

erikras:
Fixes #2356.


erikras:

PR #2362, opened 2016-12-29
Fixed bug with FieldArray fields.get()

### LINKED ISSUE #2355: FieldArray fields.get does not reflect redux state
I am having trouble with fields API provided by FieldArray.
When I add an element to FieldArray (with fields.push({})) and then set a property (say foo: "bar") in the new element created, redux state is updated but fields.get(1).foo returns undefined.
If I add another element again, fields.get(1) is defined with the property foo: "bar" set but fields.get(2) is undefined.
The problem is the same with getAll API.

erikras:
Fixes #2355.


erikras:

PR #2364, opened 2016-12-29
Allowed for an array of validators to FieldArray

### LINKED ISSUE #2343: FieldLevelValidation - Using multiple validators for Field Arrays
The validate property on FieldArray does not work with an array of validators.  When used, an error like `FieldArray.js:41 Uncaught TypeError: fn.apply is not a function` is thrown.  This is not the case with Field.  

Please allow both Field and FieldArray to work with a single function or an array of functions.

erikras:
Fixes #2343.


erikras:

PR #2371, opened 2017-01-02
Abstract away the need to call each action creator with the form argument

dagda1:
I've created a brief pull request to outline what I mean.  This is not a full pull request but merely for illustration purposes.

I think it is an unneeded burden of having to remember to add the form argument to every single action in `src/actions.js`, e.g.
```
export const destroy = (form) =>
  ({ type: DESTROY, meta: { form } })
```

This not only relies on the developer remembering to add the form field to every action but also tightly couples the implementation of how forms are identified to the way actions are called.

I think it would be nice if `redux-form` took care of creating names for forms  or at least it could create a form name if none existed.  For example, I have to supply a form name every time I create a redux-form.  What got me thinking was this scenario:
```
    if(this.state.isEditing) {
      const formName = `editForm${id}`;

      return (
        <ResultForm
            {...this.props}
            form={formName}
            initialValues={initialValues}
        />
      );
    }
```

I had come up with my own scheme to uniquely identify the form.

I blogged about a way to do this at the framework level [here](http://www.thesoftwaresimpleton.com/blog/2016/12/20/instance-reducers/).

The first step of this would be to remove the need to add the form argument to every single action creator.

In this PR, I've only done this for `formActions` in `src/reduxForm.js`.

I'm looking for feedback if you think this is a good idea.

If you think it is something worth exploring then I will create a proper pull request, if not, it was a fun thing for me to look at :).

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -52,8 +52,8 @@ export const arraySwap = (form, field, indexA, indexB) => {
 export const arrayUnshift = (form, field, value) =>
   ({ type: ARRAY_UNSHIFT, meta: { form, field }, payload: value })
 
-export const autofill = (form, field, value) =>
-  ({ type: AUTOFILL, meta: { form, field }, payload: value })
+export const autofill = (field, value) =>
+  ({ type: AUTOFILL, meta: { field }, payload: value })
 
 export const blur = (form, field, value, touch) =>
   ({ type: BLUR, meta: { form, field, touch }, payload: value })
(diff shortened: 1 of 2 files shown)


erikras:
Thank you. This is very clear. As is often the case, code > prose.

The problem with this approach is that it doesn't allow sharing of the action creators to any module that wants to import them and dispatch a "initialize the login form" action. As a library, it is better to export action creators that require a `form` parameter. Internally, `redux-form` pre-binds the `form` parameter so that subsequent calls to it need not worry about the form name, but the library itself needs to export the whole function, I think.

I _do_ think that your approach might be useful in an app itself, but I don't think it's appropriate for a library.

dagda1:
ok, thanks for the feedback. i've had a look at the issues and i'm not sure which ones are actual issues and not questions.  do you know of any definite issues i could look at?

erikras:
> do you know of any definite issues i could look at?

I'm not sure what you're referring to. What are you looking for in the issues?

PR #2385, opened 2017-01-04
Event Callbacks

### LINKED ISSUE #2334: onBlur functionality in latest version of redux-form not working
After updating to the latest version of redux-form (v 6.4.1), the onBlur is not working as expected. For example: 

I have the following two Dropdown components where the 2nd dropdown is values are dynamically allocated based on the first dropdown values.

```
constructor(props) {
  super(props);
  this.setDropdownValue = this.setDropdownValue.bind(this);
}

setDropdownValue(option) {
    const routingOptions = this.props[entitiesRoutingMap[option]] || [
      {
        id: null,
        name: 'None',
        disabled: true,
      },
    ];
    this.setState({ identityType: option, routingOptions });
  }

<Field component={Dropdown} valueKey="value" popup options={entities} className="row-popup" name="identityType" labelKey="label" label="Routing Type" description="Select the routing type you need" onBlur={this.setDropdownValue} />
        
<Field component={Dropdown} labelKey="name" valueKey="id" name="identity" popup searchable options={this.state.routingOptions} label="Routing Entry" className="row-popup" description="Some description" />
```

This code works perfectly fine for me with redux-form v6.3.2 . But in the latest version of the redux-form for some reason, this fails.

And when I install the latest version of redux-form, it asks to downgrade the react version and the redux version as well.

Not sure if anyone else is facing the same issue or not.

erikras:
Introduces new `onBlur`, `onChange`, `onFocus`, `onDragStart`, and `onDrop` callback props that can be given to `Field`, which also provide `event.preventDefault()` functionality to cancel events triggering Redux actions.

Fixes #2334.

### DIFF
--- a/docs/api/Field.md
+++ b/docs/api/Field.md
@@ -72,13 +72,41 @@ would pass `value => value.toUpperCase()`. The parameters to your normalize func
 > All the values in the entire form before the current change. This will be an Immutable `Map` if
 > you are using Immutable JS.
 
+#### `onBlur : (event, newValue, previousValue) => void` [optional]
+
+A callback that will be called whenever an `onBlur` event is fired from the underlying input.
+If you call `event.preventDefault()`, the `BLUR` action will _NOT_ be dispatched, and the value
+and focus state will not be updated in the Redux store.
+
+#### `onChange : (event, newValue, previousValue) => void` [optional]
+
+A callback that will be called whenever an `onChange` event is fired from the underlying input.
+If you call `event.preventDefault()`, the `CHANGE` action will _NOT_ be dispatched, and the value
+will not be updated in the Redux store.
+
+#### `onDragStart : (event) => void` [optional]
+
+A callback that will be called whenever an `onDragStart` event is fired from the underlying input.
+
+#### `onDrop : (event, newValue, previousValue) => void` [optional]
+
+A callback that will be called whenever an `onDrop` event is fired from the underlying input.
+If you call `event.preventDefault()`, the `CHANGE` action will _NOT_ be dispatched, and the value
+will not be updated in the Redux store.
+
+#### `onFocus : (event) => void` [optional]
+
+A callback that will be called whenever an `onFocus` event is fired from the underlying input.
+If you call `event.preventDefault()`, the `FOCUS` action will _NOT_ be dispatched, and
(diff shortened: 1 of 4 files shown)


kolpav:
Been waiting for that, thanks! redux verstion of `preventDefault` is nice touch.

erikras:

tecnolocoEdan:
In 6.3.x this worked like a charm. Then in 6.4.x it didn't not work at all. Now in 6.5.0 you may think this is working but what if I want all event object in my callback. In the next example I was using the event object to access de target and then the value, with this implementation the event object is overwritten to just pass preventDefault method.

Callback:

`onColorTypeChange = (e)=>{
        const {change} = this.props;
        if(e.target.value == 'color'){
            change('defaultTo.value', '#fff')
        }else{
            change('defaultTo.value', null)
        }
    };`

Component:

`const RadioGroupForm = ({
    input,
    type,
    label,
    placeholder,
    required,
    disabled,
    dataOptions,
    onChange,
    meta:{touched, error, dirty}
})=>{

    const onChangeHandler = (e)=>{
        onChange && onChange(e);
        input.onChange(e.target.value);
    };


    return(
        <div className={'form-group ' +((disabled)? 'form-group-disabled': '')}>
            {label && <label className="form-group-label">{label} {(required)?<span>*</span>:''}</label>}
            <div className="form-radio-group">
                {
                    dataOptions.map((option, index)=>
                        <div className="form-radio-item" key={`radio-${input.name}-${index}`}>
                            <input {...input} className="form-control" id={`radio-${input.name}-${index}`} type="radio" value={option.value} onChange={onChangeHandler} checked={input.value==option.value} placeholder={placeholder} disabled={disabled}/>
                            <label htmlFor={`radio-${input.name}-${index}`}><span className="radio-circle"></span>{option.label}</label>
                        </div>
                    )
                }

            </div>
            {touched && error && !disabled && <div className="form-group-error-container">{error}</div>}
        </div>
    )
};

export default RadioGroupForm;`

gustavohenke:
> with this implementation the event object is overwritten to just pass preventDefault method.

Not really. Check here: https://github.com/erikras/redux-form/pull/2385/files#diff-e22d664cbd78ed0741febe44dcadbd90R66

PR #2387, opened 2017-01-04
Remove `values` from instance API docs

jimniels:
Been using v6 for a little while and have noticed the fact that you can't do `this.props.values` anymore. The reasoning for this appears to be stated in point 1 [here](https://github.com/erikras/redux-form/issues/726):

> When writing the initial API, I provided things that seemed convenient for small forms, like having this.props.values be the current state of the values of your form. This required having the wrapping form component to be connected to the Redux store, _and to re-render on every value change_.

This makes sense. In order to get current values now, it appears one should use a selector.

With that being the case,  it seems like this part of the docs is incorrect? Shouldn't `values` be removed here? Or am I missing something? The `values` key doesn't appear to be part of my props object.

screen shot 2017-01-04 at 1 09 39 pm

### DIFF
--- a/docs/api/ReduxForm.md
+++ b/docs/api/ReduxForm.md
@@ -293,10 +293,6 @@ when the form is submitted successfully, or rejected if the submission fails.
 
 `true` when the form is valid (has no validation errors), `false` otherwise.
 
-#### `values : Object`
-
-The current values of all the fields in the form.
-
 #### `wrappedInstance : ReactElement`
 
 A reference to the instance of the component you decorated with `reduxForm()`. Mainly useful for


erikras:
Hmm.. `values` is a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get). I'm not sure why it isn't showing up in dev tools, but it's there in the tests.

Can you actually try to execute `form.values` in the js console?

erikras:
Also, you might be misunderstanding what "instance API" is . That's not `this.props.values`; that's `this.refs.myForm.values` from the decorated form's containing component.

jimniels:
Well you're right, I don't fully understand what "instance API" is :) but your comments make it somewhat clearer. I added a ref to my decorated form's component, i.e.

```
componentDidMount() {
  console.log(this.refs.EndorseSectionIForm);
}
// then later in render()
<div>
  <EndorseSectionIForm ref={'EndorseSectionIForm'} />
</div>
```
But that object doesn't seem to have what I'm looking for. 

screen shot 2017-01-05 at 4 55 55 pm

But if I look in `this.refs.EndorseSectionIForm.getState()` that has my `values` key with my form's values.

screen shot 2017-01-05 at 4 59 57 pm

To be honest, a lot of this is over my head, so I'm fine closing the PR since my proposed doc changes are invalid.

erikras:
Instance API refers to when you are outside of your component, and you have a reference to **an instance** of your component.

Thanks for your willingness to suggest changes. :+1:

PR #2393, opened 2017-01-05
[BUG] callback should take item as argument

yenshih:
A template string without eval seems ridiculous.
I guess this might be a bug.


gustavohenke:
`FieldArray` is supposed to work exactly as that, and the tests are used to confirm this behavior.

PR #2395, opened 2017-01-05
Set proper <title> tag to examples

### LINKED ISSUE #1891: Kindly set proper <title> tag for examples
The `<title>` tag for examples is currently set to simple "Redux Form". See this, for example,

https://github.com/erikras/redux-form/blob/master/examples/asyncValidation/index.html#L6

There is no way to discern from tab title which example is open in which tab.

screen shot 2016-10-02 at 15 46 12

gustavohenke:
Closes #1891


erikras:
Wow! If this is the level of detail we are fixing, the product must be pretty good. 😆 👍

gustavohenke:
Ah, that was a easy one I found while digging the issues :)

Issue #2394, opened 2017-01-05
Access to props from onSubmitSuccess and onSubmitFail


rmevans9:
# Problem
When handling a submit success/failure I need to call to a callback that is passed on the props of a component wrapped as a redux form. I also need this for onSubmit but worked around that by passing the onSubmit method directly to the handleSubmit method. This work around is not available for onSubmitSuccess and onSubmitFail (as far as I can tell).

# Solution
Props are available within the handleSubmit method that calls onSubmitSuccess and onSubmitFail so we could just pass props to the onSubmitSuccess and onSubmitFail callbacks in there.

### onSubmitSuccess Changes
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L38
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L58

### onSubmitFail Changes
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L22
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L46
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L76
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L86

### Docs changes
https://github.com/erikras/redux-form/blob/master/docs/api/ReduxForm.md#onsubmitfail--function-optional
https://github.com/erikras/redux-form/blob/master/docs/api/ReduxForm.md#onsubmitsuccess--function-optional

# Possible issues with solution
Although props is passed into the handleSuccess method there are changes that occur that could affect props. I don't think passing them directly will reflect these changes. I would have to test this to prove it one way or the other but it is a possible risk

# Questions
Does a change like this make sense? This could be an edge case for me but on the other hand I could see having access to props in those callbacks being useful for many things.

If it makes sense I can put up a PR to introduce this change.

gustavohenke:
Instead of showing where the code should be changed, why don't you change it yourself?
PRs are always welcome!

rmevans9:
@gustavohenke I intend to. The reason for the ticket prior to submitting the PR was to gauge interest from the project owners and to see if there was a specific reason this wasn't done.

erikras:
Fix merged. Will comment here again when released.

erikras:
Fix released as [`v6.5.0`](https://github.com/erikras/redux-form/releases/tag/v6.5.0).

PR #2396, opened 2017-01-06
Add Props to onSubmitSuccess and onSubmitFail

rmevans9:
Completes #2394 

In some situations it would be helpful to have access to the props on the component during the onSubmitSuccess and onSubmitFail callbacks. This PR passes them in on those callbacks.


erikras:

PR #2400, opened 2017-01-06
Provided form name to Field as meta prop

### LINKED ISSUE #2386: Pass "form" into props for each field
I have a need to pass the name of a form down into each `Field` / `FieldArray`. This allows me to do things like fire asynchronous actions from the field that will update the correct form / field name in the redux store via redux actions after the asynchronous action is completed.

Currently each `Field` receives it's name but in certain situations it's important to also know the form name.

erikras:
Fixes #2386.


erikras:

PR #2401, opened 2017-01-06
Avoided unnecessary deepEqual() for pristine

### LINKED ISSUE #2380: Unnecessary pristine check
On lines 551 through 555 of reduxForm.js an unnecessary and inefficient call to `deepEqual` is made.
```
if (shouldResetValues) {
    values = initial
}

const pristine = deepEqual(initial, values)
```

if `shouldResetValues` is true, then we know that `values` and `initial` are the same, therefore we can short circuit and set pristine to true without having to make an expensive `deepEqual` call on those two objects.

erikras:
FIxes #2380.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -552,7 +552,7 @@ const createReduxForm =
               values = initial
             }
 
-            const pristine = deepEqual(initial, values)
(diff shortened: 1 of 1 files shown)


erikras:

PR #2402, opened 2017-01-06
Multi-destroy and getFormNames selector

### LINKED ISSUE #2397: All forms selectors and actions
in follow up on https://github.com/erikras/redux-form/issues/1868. 

I made a page that builds dynamic "wizard" forms.
when click a 'tab/button' it will render a form like this:

```javascript
render() {
  const formName = `formSpot-${buttonIndex}`

  return(
    <formWizzard formName={formName} initialValues={initialValues} />
  )
}
```
which works perfect and have these benefits:
- it won't destroy so when go back to a 'tab/button' it will remember the state you where in
- it will only render the forms i requested to make (so if i have 4 buttons and will click on button 1 and 3 it will only register the forms 1 and 3)

as shown in the reduxDevtools:
screen shot 2017-01-06 at 10 51 29

But now i want to destroy these forms when the container unmounts: 

```javascript
//Make a list (not state cause i don't need to re-render my component.
constructor(props) {
  super(props);
  //using immutable and set up an array of the formNames
  this.formsNameArray = fromJS([]);
}
//a Function to push the new formname in the array
_updateFormsNameList(formname) {
  if (!this.formsNameArray.includes(formname)) {
    this.formsNameArray = this.formsNameArray.push(formname);
  }
}
// import { destroy } from 'redux-form/immutable'; and then connect it through mapDispatchToProps.
componentWillUnmount() {
  this.formsNameArray.forEach((formName) => this.props.dispatchDestroyForm(formName));
}
```

So 3 questions. 
1) Is this the way to go?
2) what do you think about a selector to get all the form names. 
3) Should we have actions like destroy(array:formnames) instead of destroy(string:formname)

Thanks in advance. 

extra:
Oh, and i have the same for these selectors. I would especially know which form is Dirty.
    syncErrors: getFormSyncErrors('myForm')(state),
    asyncErrors: getFormAsyncErrors('myForm')(state),
    submitErrors: getFormSubmitErrors('myForm')(state),
    dirty: isDirty('myForm')(state),
    pristine: isPristine('myForm')(state),
    valid: isValid('myForm')(state),
    invalid: isInvalid('myForm')(state),
    submitting: isSubmitting('myForm')(state),
    submitSucceeded: hasSubmitSucceeded('myForm')(state),
    submitFailed: hasSubmitFailed('myForm')(state)

erikras:
* Allowed multiple form names to be passed to `destroy()` action creator.
* Created `getFormNames(state)` selector

Fixes #2397.

Allows for fun stuff like:
```jsx
import { getFormNames, isDirty } from 'redux-form'

const dirtyForms = getFormNames()(state).filter(form => isDirty(form)(state))
```

### DIFF
--- a/docs/api/Selectors.md
+++ b/docs/api/Selectors.md
@@ -15,6 +15,7 @@ import {
   getFormSyncErrors,
   getFormAsyncErrors,
   getFormSubmitErrors,
+  getFormNames,
   isDirty,
   isPristine,
   isValid,
@@ -30,6 +31,7 @@ MyComponent = connect(
     syncErrors: getFormSyncErrors('myForm')(state),
     asyncErrors: getFormAsyncErrors('myForm')(state),
     submitErrors: getFormSubmitErrors('myForm')(state),
+    names: getFormNames('myForm')(state),
     dirty: isDirty('myForm')(state),
     pristine: isPristine('myForm')(state),
     valid: isValid('myForm')(state),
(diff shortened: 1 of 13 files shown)


erikras:

pkuzhel:
Attempting to import getFormNames from redux-form and getting undefined on v6.5.0. Any idea why this might happen?

PR #2403, opened 2017-01-06
Removed array-findindex-polyfill

### LINKED ISSUE #2337: Remove array-find-index-polyfill?
I've upgraded redux-form from v3 to v6, a weird thing happened.

I realized it is redux-form that includes the polyfill.

It is fine in most cases, but it did cause an unexpected bug due to some bad code in our project:

```js
for (var key in arr) { } // wrong
for (var item in arr) { } // right
```

Just a suggestion, let's remove the (aggressive) polyfill? 😆  ...since it is only used in one place.

image

erikras:
Silly to use this for just one loop.

Fixes #2337.


erikras:

PR #2409, opened 2017-01-09
Restore old behavior of passing down name prop for Field, to be in li…

huan086:
…ne with documentation at http://redux-form.com/6.4.3/docs/api/Field.md/#-name-string-

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -172,6 +172,7 @@ const createConnectedField = ({ deepEqual, getIn, toJS }) => {
         {
           ...rest,
           form: _reduxForm.form,
+          name,
           onBlur: this.handleBlur,
           onChange: this.handleChange,
           onDrop: this.handleDrop,
--- a/src/__tests__/Field.spec.js
+++ b/src/__tests__/Field.spec.js
@@ -1527,7 +1527,7 @@ const describeField = (name, structure, combineReducers, expect) => {
 
       expect(renderSpy).toHaveBeenCalled()
       const props = renderSpy.calls[ 0 ].arguments[ 0 ]
-      Object.keys(apiProps).forEach(key => expect(props[ key ]).toNotExist())
+      Object.keys(apiProps).filter(key => key !== 'name').forEach(key => expect(props[ key ]).toNotExist())
     })
 
     it('should only rerender field that has changed', () => {


huan086:
@oreqizer I'm adding the Field name prop back, in relation to my comment at https://github.com/erikras/redux-form/commit/92399dd59be736b381f79ec66e25920c74363658

gustavohenke:
What you linked here is the instance API, not the props. They are different.

Small example to get you started in the difference between them:

```js
class MyComponent extends Component {
  // This getter is part of the API of MyComponent
  get result () {
    return this.props.a + this.props.b;
  }

  render () {
    return <span>{ this.result }</span>;
  }
}

// 'a' and 'b' are props
<MyComponent a={ 5 } b={ 3 } />
```

See? The instance API are methods and properties/getters available in the definition of your component. 
Props are those attributes you pass to the component when it's used in JSX code.

huan086:
The instance api is BROKEN by the pull request I mentioned. In MyComponent, it used to get this.props.name. After that commit, it does not receive name any more, which is unintended as the original behaviour is what's documented.

Please merge

gustavohenke:
`name` was never documented as a prop provided for your custom field component.
What you're probably looking for is `input.name`.

huan086:
From the docs:


Instance API

The following properties and methods are available on an instance of a Field component.

dirty : boolean


true if the current value is different from the initialized value, false otherwise.

name : String


The name prop that you passed in.

gustavohenke:
It's under "Instance API" section. You're confusing stuff, it's not a prop.

huan086:
I'm saying the instance api is broken...

MyComponent is not receiving name, contrary to what is documented. The offending pull request removed the name while refactoring the code to reuse event habdlers

gustavohenke:
Ah, okay.
But anyway I could not reproduce



```
λ npm ls redux-form
my-project@0.2.2
+-- redux-form@6.4.3
```

Do you have some code that reproduces your issue?

gustavohenke:
Hmm, as you edited `ConnectedField` in this PR, I must say that then you're right, `name` is not available in the instance API.
But `ConnectedField` is a component that is not documented anyway, so you should not be using it.

huan086:
Ok, let me go and learn how to write the unit test to make sure props don't disappear again. My PR a few months back broke `name` for FieldArray while refactoring to reuse the ConnectedX components...

huan086:
I've read through the unit tests and understand what's instance API now. So I had confused props with instance API. However

1. `names` is passed down as props for `Fields`
2. `name` is passed down as props for `FieldArray`. I had accidentally removed the `name` prop in a PR, and it was restored quickly.
3. `name` **used** to be passed down as props for `Field`. The PR removed it while refectoring. Looks like an accidental removal to me.

The change had greatly impacted me as I had been using

```jsx
const { name, input } = props;
return (
    <div>
        <label htmlFor={name}>label</label>
        <input {...input} id={name} />
    </div>
);
```

All my `htmlFor` is broken. All my UI tests are broken due to to missing id.

# Choice 1

Document the behavior that `names` is provided as prop for Fields, and `name` is provided as prop for FieldArray **and** Field

# Choice 2

Document the behavior that `names` is provided as prop for Fields, and `name` is provided as prop for FieldArray but not for Field because reasons

# Choice 3

Remove `names` and `name` from props for Fields and FieldArray because they are undocumented, breaking everyone's code and be hated.


# My choice

Shall be resubmitting my PR taking choice 1

huan086:
@gustavohenke, I'm facing problems.

I was expecting that `name` and `names` are passed down from Field and Fields to the component unchanged, just like custom props.

However, inside FormSection, `name` and `names` props get prefixed too (not just input.name). This makes `name` and `names` useless, since they can't be used to indicate the names of the keys to retrieve meta and input in the props.

Use case is that I have a form with several groups of checkboxes/textboxes that needs to be checked/filled. When all checkboxes/textboxes in a group are checked/filled, I display a message. Currently, I'm depending on the `names` prop to know the keys where I can retrieve the input prop. Thus `names` must be the original names I've passed in.

# Question

Should I change the current behavior of `name` and `names` inside of FormSection so that they retain the original value?

# Related documentation problem

The instance API states

`name`: The name prop that you passed in.

However, it should have been

The name prop that you passed in, prefixed accordingly when inside a FormSection

huan086:
@m14t need your opinion on this, as you too has been depending on the undocumented name prop being passes to your component.

1. What do you use the name prop for? How did my accidental removal of the prop affect you?
2. Are you expecting the name prop to be what you passed in? Or do you expect that it gets prefixed when inside FormSection

m14t:
@huan086 Thanks for bringing me into the conversation. It's been a bit since I've had to deal with this issue.  Do you have a link to my prior post to help job my memory?

huan086:
@m14t https://github.com/erikras/redux-form/pull/1816 is the pull request you made to fix my accidental removal of FieldArray name.

Now, it looks like https://github.com/erikras/redux-form/commit/92399dd59be736b381f79ec66e25920c74363658 accidentally removed name from Field

None of these accidents get caught by unit tests because the `name` prop is undocumented and never meant to be passed to the component.

However, it looks like you and I had depended on the `name` and `names` to do certain things, thus I'm proposing to properly document the behavior.

While writing the unit tests to formalize the behavior, I realize that the current behavior is that `name` and `names` get prefixed by FormSection. I was expecting that those props be passed through untouched, i.e. treated just like additional custom props.

I am not sure to formalize the behavior as passing the `name` and `names` prefixed (current behavior), or passing the `name` and `names` through, unprefixed (changing the behavior).

Thus would like to know people had been using the `name` and `names` undocumented props that get passed to the component

m14t:
Thanks!

So our use-case is that we have a large form that represents a "Page" on our site.  Each page can have many components of various types (such as Article, Related Posts, and Comments), in any order, etc.  Each of those components have various fields that you need to fill in and are different from other components.

This makes the structure of our form vary quite a bit from page to page (and in-fact can change after the user loads the page, when they add additional components to a page).

Our data structure is essentially:
```javascript
page: {
  id: 'id-1',
  slug: '/some-slug'
  title: 'My SEO title',
  components: [
    {
      type: 'Article',
      showIf: 'loggedOut',
      theme: {
        name: 'theme-1',
      }
    }
  ]
}
```

We then essentially have `page.components.map()` that renders a `<FieldArray>` for each component with the correct sub-inputs based on the `type`.

Those sub components will essentially be passed `page.components[0]` as the `name`, and so we are able to render the sub-inputs like:

```jsx
  <Field
    component={ThemeInput}
    name={`${props.name}.theme`}
  />
```

And `ThemeInput` is able to render its input like:
```jsx
  <SelectField
    name={`${this.props.name}.name`}
    onChange={this.onChange}
    value={this.getValue()}
  >
    {/* ... */}
  </SelectField>
```

---

Does that help?

huan086:
It looks like you don't need `FieldArray`, since you are calling `page.components.map()` yourself. Thus, you could do

```jsx
page.components.map((component, index) => <Field name=`page.components[${index}]` component={getComponent(component.type)} />)
```

Then inside SelectField, you will use
```jsx
<SelectField name={`${this.props.input.name}.name`}
```

What's the point of using FieldArray?

_BTW, if you update to the latest redux-form, your code will break too, since you are reading from `props.name` in ThemeInput_

m14t:
The `page.components.map()` loops over each of the components.  The `<FieldArray>` is then used to render the multiple inputs that that type has... so it's definitely needed _right now_, but I think you are right that we could update this to make those "multiple inputs" into a single custom field type, and get back in line with the documented behaviors.  (Unfortunately this complex form was literally our first usage of this library so it seems that we were using it slightly incorrectly.)

Thanks for your suggestion, and I hope this was useful to you.  I'll try to purge the usage of `props.name` from our codebase in the next few days and let you know if we have an issues.

huan086:
Given that passing down `name` seems to encourage incorrect code which should have been reading from input.name or fields.map instead, I'm now leaning towards

# Choice 4

Remove names and name from props for Field, Fields and FieldArray because they are undocumented, breaking everyone's code and be hated.

Add the props `inputs` and `metas` to Fields and deprecate props keyed by name. This would allow Fields component to handle the inputs dynamically, getting the name from `props.inputs[i].name` instead of `props.names`

gustavohenke:
Hi @huan086, I was away for some days.

First, I agree with your Choice 4, it seems feasible to me.

Second, I must say that I'll simply ignore any hate messages about removing undocumented features.
I'll quote my friend @FagnerMartinsBrack in his awesome article [Library Internals Do Not Exist](https://medium.com/@fagnerbrack/library-internals-do-not-exist-1ff50a30e3b0#.c3td29q1b):

> Just because it "behaves" correctly, it doesn't mean that a library is being used the way it should. It is the documentation, A.K.A. the **public API**, that serves as a reference to verify if somebody is using the API correctly. It is the contract between the author and consumer, the single source of truth.  
> Nothing else.

PR #2415, opened 2017-01-09
Fix prop name

chiefGui:
On the `ContactPage` component, the prop that is being passed is `onSubmit` instead of `handleSubmit`.

This issue may lead to confusion to beginners.


chiefGui:
Nevermind.

I'll close this a potential mistake lead by the transformation of the `onSubmit` prop into `handleSubmit`.

PR #2423, opened 2017-01-10
Choose getIn based off Iterable.isIterable()

dxinteractive:
Currently `structure.getIn` breaks when using Immutable `Record`s or other keyed Immutable objects that aren't `Map`s or `List`s, because it tries to use the `plainGetIn` on those.

I've tested this fix in the context of an app I'm working on and these changes work in my case, but haven't verified with any actual tests. By definition all `Iterable`s have `getIn` so I think it's a low risk change, at least one that's guaranteed to not break any existing desired behaviours with `Map`s and `List`s.

The symptoms are that `ConnectedField`s are unable to find the field's values in state and `value` is subsequently passed down as an empty string.

### DIFF
--- a/src/structure/immutable/index.js
+++ b/src/structure/immutable/index.js
@@ -12,7 +12,7 @@ const structure = {
   empty: Map(),
   emptyList,
   getIn: (state, field) =>
-    Map.isMap(state) || List.isList(state) ? state.getIn(toPath(field)) : plainGetIn(state, field),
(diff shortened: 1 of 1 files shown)


dxinteractive:
Ooh, hold off on the pull. This is incomplete, as the equivalent `setIn` also can't cope with anything other than `map` and `List` items. There's more logic in there that I would have expected, I'll make some tentative changes and resubmit this PR for review.

erikras:
Good one. 👍

erikras:

Issue #1994, opened 2016-10-19
Set new pristine state after successful submit


oakley808:
Is it possible to reset the pristine status of a form after it's successfully submitted?
Something like

```
const MyForm = reduxForm({ 
  form: 'myForm',  
  onSubmitSuccess: (result, dispatch) => {
      // get the new values and initialize with them?
      // ...but we no access to state here  :-(
      const newInitialValues = getFormValues('myForm')(state);

      dispatch(initialize('myForm', newInitialValues));
    });
  }
})(MyForm);
```

Is there another way to reset either the pristine or dirty state?

jedwards1211:
@erikras the problem with dispatching `initialize` is that it also resets the `submitSucceeded` prop, so the banner I'm showing to indicate success disappears.  I don't think there's an action that can do what I want.

erikras:
Fix released as [`v6.5.0`](https://github.com/erikras/redux-form/releases/tag/v6.5.0).

aplanchamp:
Hello,

I also have the need to reset the prisine after a successfull submission form. I tried this method, here is my code :

`export default connect(mapStateToProps, mapDispatchToProps)(
  reduxForm<ManageQuotingCriteriaFormData, ManageQuotingCriteriaFormProps>({
    form: Forms.QuotingCriteriaForm,
    enableReinitialize: true,
    onSubmitSuccess: (_result: any, dispatch: DispatchRedux<any>, props: any) => {
      dispatch(initialize(Forms.QuotingCriteriaForm, props.values));
    }
  })(ManageQuotingCriteriaForm)
);
`

This actually works since it initializes my form and resets pristine. But it also breaks pristine (or dirty by the way) because then is it always equal to true even if a keep updating my form. It just does not do its job anymore. Any idea where the problem is ? I tried removing `enableReinitialize: true` the problem is the same. If I use`reset` i don't have a problem, but reset is not what I want to achieve...

PR #2426, opened 2017-01-10
feat(initialize): add keepSubmitSucceeded option

jedwards1211:
My proposed solution to #1994 
Also, I made it possible to pass `keepDirty` in the same options argument as `keepSubmitSucceeded` is passed in, for the sake of cleanliness, but I preserved backwards compatibility.


erikras:

jedwards1211:
@erikras thank you!

Issue #2430, opened 2017-01-10
Suggestion: Provide validation functions with access to form props


timhwang21:
Currently, validation rules are provided (field value, object of all form values) as arguments. It would be nice if they were also given access to the props passed to the form. 

My use case is that I have several external props that are used in my validation functions. Normally, I can just write my validation functions inside my form and access `this.props.myStuff`, but there is an edge case where the props passed change. Currently `validateIfNeeded()` seems to happen before my component renders (and my rules recalculated based on the new props), so I'm getting incorrect results from my rules. (Worth noting: I'm using field-level rules).

I could solve this by writing a custom `shouldValidate()` which checks if the props in question change, and if so validate; however, my form is quite large, and I would be needlessly running validation if everything is based in `shouldValidate()`. It would be much easier to handle if rules had direct access to props.

Alternatively, an action dispatcher to run validation and to run warnings would also solve my issue.

gustavohenke:
If I'm not mistaken, this was fixed by #2431!

timhwang21:
You are not, that's my PR :)

erikras:
Fix released as [`v6.5.0`](https://github.com/erikras/redux-form/releases/tag/v6.5.0).

PR #2431, opened 2017-01-10
Provide validation functions with access to form props

timhwang21:
Currently, the API for `validate` passed to `config` allows props to be
accessible by validation rules. The same should be done for field-level
validations. This PR provides props passed to the form as the second argument
of the function returned by `generateValidator()`, making the APIs between the
two forms of validation consistent.

Refers to #2430

### DIFF
--- a/docs/api/Field.md
+++ b/docs/api/Field.md
@@ -121,13 +121,13 @@ localized date formats into `Date`s.
 `parse` is called with the field `value` and `name` as arguments and should return the new
 parsed value to be stored in the Redux store.
 
-#### `validate : (value, allValues) => error` [optional]
+#### `validate : (value, allValues, props) => error` [optional]
 
-Allows you to to provide a field-level validation rule. The function will be given the current value of the field and all the other form values. If the field is valid, it should return `undefined`, if the field is invalid, it should return an error (usually, but not necessarily, a `String`).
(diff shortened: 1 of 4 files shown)


timhwang21:
Note: while I thought this would be a new feature, it's actually an API mismatch between the form level validate prop and the field level validation :)

erikras:
Um, I don't see the part where you are actually passing the `this.props`. What am I missing?

Can you please add a test in `reduxForm.spec.js` that shows the props actually going all the way through?

timhwang21:
The props are already being passed in `reduxForm [`validateIfNeeded()`](https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L182). 

`fieldLevelValidate` is the function returned by `generateValidator()`, and is passed `propsToValidate.values` and `propsToValidate` on line 185. However, the function currently being return by `generateValidator()` [only takes one argument: `values`](https://github.com/erikras/redux-form/blob/master/src/generateValidator.js#L15). 

The only thing my patch does is allow `generateValidator()`'s function to also pass on the props that are already being passed in as an argument.

erikras:
Okay. I see. Nice work.

timhwang21:
Thanks! I was in the midst of writing tests and noticed that there aren't tests for `warnIfNeeded`. I can make a separate PR for it, though I'm having trouble figuring out how to mock `fieldLevelValidate()`, which needs access to `this.fieldValidators`.

erikras:

PR #2445, opened 2017-01-11
Fixed #2444 unexpected value updates when radio lose focus

### LINKED ISSUE #2444: blur on a radio cause redux-form to set incorrect value
Steps to reproduce: take the `Simple` example

```
1. set a default focused element
<label><Field ref={this.setFocus} name="sex" component="input" type="radio" value="female"/> Female</label>

setFocus(el) {
  const found = findDOMNode(el)
  if(found) {
    found.focus()
  }
}

2. click anywhere else to make the radio to lose focus
3. the value of field changed unexpected
```

See the screenshot:

redux-form

The reason: `blur event` of a radio has the value property of the element. `<Field value="female' type="radio">` which will be used to update the form value state when blur.

I did a quick fix to it because this small issue bugs me a lot. Will make a pull request later.

ruanyl:


erikras:
Seems reasonable. Any idea why `codecov` is complaining?

Have you run `npm run test:cov`?

ruanyl:
@erikras added one more test case, `codecov` should be happy now ;)

ruanyl:
@erikras Any schedule for next release?

erikras:

PR #2450, opened 2017-01-12
Add more tests for validateIfNeeded & warnIfNeeded

timhwang21:
References #2431.

* Added `warnIfNeeded()` tests based on `validateIfNeeded()`
* Added a test to check if values and props are correctly passed to `validate` & `warn`

I'm not sure my method for checking props is the most elegant (spying on the props passed to `shouldValidate` and comparing with the argument of `validate` with `Object.keys()`) but I couldn't find a state where the props had the exact same values as well.

Let me know if I missed anything, thanks!


erikras:
Great! 👍

PR #2452, opened 2017-01-12
fix: when the Field type is "file", value can not be updated

sjkyspa:


sjkyspa:
any update about merge pull request and next release?

erikras:
Can you describe the problem that this is solving?

How does this `tmp` library help in the testing? Does it create the same kind of file that is present in the browser with a `<input type="file"/>` input?

sjkyspa:
@erkras

see the following example

```
<Field name="image" type="file" component={FileInput}></Field>
```

```
const FileInput = ({input: {name, value, onChange}, meta: {touched, error}, ...others}) => {
    let files = value;
    return (
        <div>
            {
                files && files[0].name && files[0].type && (
                    <Dropzone
                        name={name}
                        onDrop={(filesToUpload) => {
                            onChange(filesToUpload);
                        }}
                        multiple={false}
                    >
                        <img src={files[0].preview}/>
                    </Dropzone>
                )
            }
        </div>
    )
}
```

the input value is always undefined.. because of the code follow in the createFieldProps 

```
  if (type === 'file') {
    return {
      ...props,
      value: undefined
    }
  }
```


the tmp is use to create tmp file to ensure the correct handling of the file.

sjkyspa:
@erikras any updates? I really need the file input field.

erikras:
Seems like `value || undefined` is pretty much the same as just `value`, apart from cases when `value` is `null`, `''`, `0`, `'0'`, or `false`, which don't seem like they would ever happen. Couldn't it just be `value`?

erikras:

PR #2460, opened 2017-01-15
Add initial value selector

griffinmichl:
This seemed like an important missing selector. While a user of redux-form will usually have access to the initial values they passed in one way or another, exposing a selector for them can make it easier to hook into redux-form with epic, thunks, and sagas (invoking the selector on getState()) and keep business logic out of the component.

I'd be happy to describe a specific use case if desired.


erikras:

PR #2465, opened 2017-01-16
Allow FieldArray to be impure

just-boris:
I am trying to implement some interface that is similar to http://redux-form.com/6.4.3/examples/selectingFormValues/ but fields should be wrapped into `FieldArray`.

Currently, this is not possible, because `shouldComponentUpdate` prevents all further updates, even they are possible. See added test, it didn't passed before, but currently it passes.

### DIFF
--- a/src/__tests__/FieldArray.spec.js
+++ b/src/__tests__/FieldArray.spec.js
@@ -401,7 +401,7 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
     it('should get sync errors from outer reduxForm component', () => {
       const props = testProps({
         values: {
-          foo: 'bar'
+          foo: [ 'bar' ]
         }
       }, {
         validate: () => ({ foo: { _error: 'foo error' } })
@@ -412,7 +412,7 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
     it('should get sync warnings from outer reduxForm component', () => {
       const props = testProps({
         values: {
-          foo: 'bar'
+          foo: [ 'bar' ]
         }
       }, {
         warn: () => ({ foo: { _warning: 'foo warning' } })
@@ -423,7 +423,7 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
     it('should get async errors from Redux state', () => {
       const props = testProps({
         values: {
-          foo: 'bar'
+          foo: [ 'bar' ]
         },
         asyncErrors: {
           foo: {
@@ -437,7 +437,7 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
     it('should get submit errors from Redux state', () => {
       const props = testProps({
         values: {
-          foo: 'bar'
+          foo: [ 'bar' ]
         },
         submitErrors: {
           foo: {
@@ -451,7 +451,7 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
     it('should get submitFailed from Redux state', () => {
       const props = testProps({
         values: {
-          foo: 'bar'
+          foo: [ 'bar' ]
         },
         submitFailed: true
       })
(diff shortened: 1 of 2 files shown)


huan086:
Doesn't this change make the what's shown in the UI different from what's stored in the redux store?

just-boris:
@huan086 it will do exactly the opposite. I have provided a use-case where form state is out of sync with the store. My fixed resolves this problem.

src/__tests__/FieldArray.spec.js
```diff
@@ -1182,6 +1186,58 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
...
+        pure: false,
+        form: 'testForm'
+      })(Form)
+      const dom = TestUtils.renderIntoDocument(
+        <Provider store={store}>
+          <TestForm/>
+        </Provider>
+      )
+
+      const checkbox = TestUtils.scryRenderedDOMComponentsWithTag(dom, 'input')
```
erikras: If you are only showing the field if `hasCollar` is true (on [1207](https://github.com/erikras/redux-form/pull/2465/files#diff-c51b679ed2957be9891b8e3164a067b1R1207)), how will this element ever be rendered?

just-boris: Now, I show the `span` element, when checkbox is checked.

The checkbox field itself at the line 1208 is always visible.

huan086:
@just-boris Could you write tests for Field and Fields as well? I suspect they should not shouldComponentUpdate as well

erikras:
This concerns me a bit, since a great deal of effort went into minimizing unneeded renders. But the tests are checking for excessive renders, so if they are passing, I guess it's okay.

I would like to second @huan086's request for additional tests.

just-boris:
@huan086, did you mean that you also want me to modify the `Field` component? I believe, that it could be done as a separate Pull-request.

@erikras not sure, what should I add to Field test suite, as I did not change anything there.

huan086:
FieldArray, Field and Fields all have the same shouldComponentUpdate. I'm having a confusion now as to how anyone of them would ever get updated since they always receive the same props, thus would stop further rendering. (the actual state gets sneak through via context, which don't trigger an update I think) 

I'm thinking of a test which have 2 text input fields mirror each other. See if it passes. Suspect it fails like your FieldArray test.

If it passes, could someone enlighten me on how ConnectedField and ConnectedFields gets updated without the Field and Fields updating?

just-boris:
I have created an example with two fields with the same name. Everything works fine, though.
http://www.webpackbin.com/E1tZnnHvG

FieldArray is more special than other components because it gives you an access to field values via `fields.get(index)`, which doesn't work correctly now. `Field` and `Fields` do not have this object, so they don't have the problem.

huan086:
Looks like `fields.get` isn't a good idea, since react-redux won't know that you are dependent on certain values. So far, what I've needed is to look at specific properties of each element in the array (most commonly the id). Maybe the API should be changed to the following?

`<FieldArray mapProps={['id', 'other']} component={TheArray} />`

and on the receiving side

`const TheArray = ({ fields, values }) => { /* values contains [ { id: 'the id', other: 'the other property' }, ... ] */ }`

mqklin:
@just-boris does this PR fix [this issue](https://github.com/erikras/redux-form/issues/2577)?

just-boris:
@mqklin Yes, seems related.

erikras:
@just-boris Can you look into these test errors? Something must have changed in `master` since you originally submitted this. I went to test it to merge and it won't build. 😢 

Apologies for letting the dust collect on this. I'd like to merge it asap once I can confirm that it works fine.

overdrivenpotato:
Any word on this PR? This is an urgent issue within our organization

gDelgado14:
I wrote a test case here using @just-boris's webpackbin setup

https://www.webpackbin.com/bins/-KgoQMIyvxdJb7Pvh0eu

just-boris:
Fixed the build.
It failed, because some of recently added tests used wrong value. Actually it never worked, but just was not visible, because redering have been cancelled in `shouldComponentUpdate`

erikras:

mqklin:
I've just updated from 6.5.0 to 6.6.2, and I think this PR broke meta.error passing. Now my fields get meta.error as 'undefined' (in v6.5.0 it was OK).
Code (just for an example):

```
<form>
  <FieldArray name="some-name component={SomeComponent}/>
</form>
...
@reduxForm({validate})
class SomeComponent extends Component {
  render() {
    return (
      <Field component={Input} />
    );
  }
}
...
const Input = (props) => {
  console.log(props.meta.error); // undefined here, but in 6.5.0 was error text
};
```
@just-boris do you have this problem as well?

PR #2467, opened 2017-01-17
Remove undocumented name, validate and warn from FieldArray props.

huan086:
Document FieldArray fields.name prop.


huan086:
Should I rebase this?

Issue #2223, opened 2016-12-06
Field is unregistered when re-arranged in UI


singhba:
```jsx
StockAppState = {
	Primary: [{name: 'ASX'}],
	Secondary: [{name: 'NASDAQ'}]
}

class StockApp extends React.Component {
	render() {
		<div>
			<Component1 Stock={this.props.StockAppState.Primary}
			<Component2 Stock={this.props.StockAppState.Secondary}
		</div>
	}
}
```
When I swap state Primary=[{name: 'NASDAQ'}] & Secondary=[{name: 'ASX'}], it unregisters NASDAQ field.

erikras:
It's unclear to me which component is your form, and where your fields are. If you unmount a field, it will be unregistered. If you remount it again, it will be re-registered.

huan086:
Related: https://github.com/erikras/redux-form/issues/2016

nktssh:
possibly related #1705

erikras:
Fix released as [`v6.5.0`](https://github.com/erikras/redux-form/releases/tag/v6.5.0).

Issue #1705, opened 2016-09-05
Is this library prone to React optimization bug?


halt-hammerzeit:
At first it seems that it's not and that it handles React reusing elements correctly:

``` js
    componentWillReceiveProps(nextProps) {
      if (this.props.name !== nextProps.name) {
        ...
        // unregister old name
        this.context._reduxForm.unregister(this.props.name)
        // register new name
        this.context._reduxForm.register(nextProps.name)
      }
    }

    componentWillUnmount() {
      this.context._reduxForm.unregister(this.name)
    }
```

But what if you had this form:
- `name`
- `email`
- `password`

And then it became this form:
- `email`
- `password`

What happens is:
- `name` field is transformed into `email` field: `name` is unregistered, `email` is registered
- `email` field is transformed into `password` field: `email` is unregistered, `password` is registered
- `password` field is unmounted: `password` is unregistered

And there you have it: no fields are registered though it's clearly nonsense.

huan086:
Related: https://github.com/erikras/redux-form/issues/2016

erikras:
Fix released as [`v6.5.0`](https://github.com/erikras/redux-form/releases/tag/v6.5.0).

PR #2470, opened 2017-01-17
Keep track of registration count in registeredFields

### LINKED ISSUE #1990: Bug: UNREGISTER_FIELD called when component Field is mounted in gets unmounted (even with destroyOnUnmount: false)

### LINKED ISSUE #2016: [6.1.1] Bug: state.registeredFields not keeping track of number of calls

huan086:
For backward compatibility, when registration count is one, the registeredFields will not have the count property. Undo #2040. Fixes #2016, #2223, #1705. Partially fixes #1990.

### DIFF
--- a/src/__tests__/Field.spec.js
+++ b/src/__tests__/Field.spec.js
@@ -647,7 +647,7 @@ const describeField = (name, structure, combineReducers, expect) => {
       expect(store.getState()).toEqualMap({
         form: {
           testForm: {
-            registeredFields: [ { name: 'foo.bar', type: 'Field' } ]
+            registeredFields: { 'foo.bar': { name: 'foo.bar', type: 'Field', count: 1 } }
           }
         }
       })
@@ -674,7 +674,7 @@ const describeField = (name, structure, combineReducers, expect) => {
       expect(store.getState()).toEqualMap({
         form: {
           testForm: {
-            registeredFields: [ { name: 'foo.fighter.bar', type: 'Field' } ]
+            registeredFields: { 'foo.fighter.bar': { name: 'foo.fighter.bar', type: 'Field', count: 1 } }
           }
         }
       })
@@ -705,7 +705,7 @@ const describeField = (name, structure, combineReducers, expect) => {
       expect(store.getState()).toEqualMap({
         form: {
           testForm: {
-            registeredFields: [ { name: 'foo', type: 'Field' } ]
+            registeredFields: { foo: { name: 'foo', type: 'Field', count: 1 } }
           }
         }
       })
@@ -716,7 +716,7 @@ const describeField = (name, structure, combineReducers, expect) => {
       expect(store.getState()).toEqualMap({
         form: {
           testForm: {
-            registeredFields: [ { name: 'bar', type: 'Field' } ]
+            registeredFields: { bar: { name: 'bar', type: 'Field', count: 1 } }
           }
         }
       })
@@ -1275,7 +1275,7 @@ const describeField = (name, structure, combineReducers, expect) => {
             values: {
               age: 15 // number
             },
-            registeredFields: [ { name: 'age', type: 'Field' } ]
+            registeredFields: { age: { name: 'age', type: 'Field', count: 1 } }
           }
         }
       })
--- a/src/__tests__/FieldArray.spec.js
+++ b/src/__tests__/FieldArray.spec.js
@@ -694,11 +694,11 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
       expect(store.getState()).toEqualMap({
         form: {
           testForm: {
-            registeredFields: [
-              { name: 'foo.bar', type: 'FieldArray' },
-              { name: 'foo.bar[0].val', type: 'Field' },
-              { name: 'foo.bar[1].val', type: 'Field' }
-            ],
+            registeredFields: {
+              'foo.bar': { name: 'foo.bar', type: 'FieldArray', count: 1 },
+              'foo.bar[0].val': { name: 'foo.bar[0].val', type: 'Field', count: 1 },
+              'foo.bar[1].val': { name: 'foo.bar[1].val', type: 'Field', count: 1 }
+            },
             values: {
               foo: { bar: [ { val: 'dog' }, { val: 'cat' } ] }
             }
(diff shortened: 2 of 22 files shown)


huan086:
@erikras what to do with the `some` that isn't covered due to me using [].some instead?

huan086:
Don't merge yet. I think it makes more sense to make registeredFields an object instead, so that there won't be any need for loops. Should make my 100 fields form perform better

erikras:
Okay, let me know when you feel more confident about this PR.

huan086:
Ops, I've accidentally mixed some changes in #2473 into this pull request. Will you be merging #2473? If you are, this pull request will have a merge conflict. I'll update this pull request to resolve the conflict.

If you are not, then I'll update this pull request to remove the changes in #2473.

huan086:
Fixed merged conflict, removed changes meant for #2473.

Ready to merge!

erikras:

thuringia:
@erikras Is the breaking change regarding the type of registeredFields intended? It used to be an Array but now it's an Object.

huan086:
I change it to an object as an optimization. Previously when using array, it needs to loop through all the elements to find the registered field. That is O(n^2) and isn't good for large forms.

The final commit message daying backwards compatible in master is misleading though because it is a squash of my first commit that was backwards compatible with the breaking change afterwards

erikras:
@thuringia Hmm... I don't really consider changing the internal structure, not exposed via any API, as a breaking change. You had some code that was looping through `registeredFields`?

thuringia:
@erikras We're looping through `registeredFields` in order to filter for inactive or invisible fields to prevent faulty errors from being displayed due to our rather large validation schemas.

gustavjf:
> Edit: added link
It is shown in the official docs though: redux-form 6.5.0 docs

This is a valid point, could this be updated to reflect the change?

yoadsn:
@erikras I thought the same but life is hard sometimes - Server side package is V6.6.0 - client side bundled package uses V6.4.1.
Server side generated the redux state and it is hydrated on the client.
Client chokes on the new type present in the initial state it got from the server.

If this was considered a breaking change, semantic versioning would not allow the server to have a different package compared to the one the bundler used (different machine building the client side code - same package.json).

So, yeah - since the redux state *is* the API between the server side and the client side - this indeed is a breaking change and not an internal detail. I would never anticipate this though.

(Just consider multiple machines, each one NPM installs a different semantic version which is actually not compatible). A client get's a bundle from any of those machines - and it's lottery time.

Thank god I'm migrating to Yarn.

yoadsn:
Oh, just realized this breaks SSR in a way #2742

PR #2473, opened 2017-01-17
Do not set FieldArray as touched on submit

### LINKED ISSUE #2014: [6.1.1] Bug: ARRAY_* reducers incorrectly uses splice on `state.fields` objects
redux-form state.fields properties are objects. Fields for arrays are stored as { 0: { prop1: { touched: true }}, touched: true}. However, in reducer.js, ARRAY_\* operates on this object as if it were an array [result = doSplice(result, 'fields', field, index, removeNum, nonValuesValue)](https://github.com/erikras/redux-form/blob/master/src/reducer.js#L59).

state.fields is thus corrupted after that.

huan086:
as it causes its entry in storeState.fields to become an object instead of an array. The ARRAY_* reducers expects the fields entry to be an array. Fixes #2014.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -1563,7 +1563,46 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
 
       expect(stub.fieldList).toContainExactly([ 'sec.bar', 'sec.barArray' ])
     })
+    it('should not set FieldArray as touched on submit', () => {
+      const store = makeStore({})
+      const onSubmit = createSpy()
+      const noopRender = () => <div/>
 
+      class Form extends Component {
+        render() {
+          const { handleSubmit } = this.props
+          return (
+            <form onSubmit={handleSubmit}>
+              <FieldArray name="fooArray" component={noopRender} type="text"/>
+              <button type="submit">Submit</button>
+            </form>
+          )
+        }
+      }
+      const Decorated = reduxForm({ form: 'testForm' })(Form)
+
+      const dom = TestUtils.renderIntoDocument(
+        <Provider store={store}>
+          <Decorated onSubmit={onSubmit}/>
+        </Provider>
+      )
+
+      const form = TestUtils.findRenderedDOMComponentWithTag(dom, 'form')
+      TestUtils.Simulate.submit(form)
+      expect(onSubmit).toHaveBeenCalled()
+      expect(store.getState()).toEqualMap({
+        form: {
+          testForm: {
+            anyTouched: true,
+            registeredFields: {
+              fooArray: { name: 'fooArray', type: 'FieldArray', count: 1 }
+            },
+            submitSucceeded: true
+          }
+        }
+      })
+    })
+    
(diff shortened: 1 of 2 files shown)


gustavohenke:
Where is this problem happening, exactly?
I thought it was something to do with `setIn`, but I couldn't find evidence.

huan086:
Not related to setIn.

- The problem is that ARRAY_* reducers expects that the store state fields to be array.
- Given a form with `FieldArray name='foo'`, and 2 elements in the array, the registeredFields are `[{name: 'foo', type: 'FieldArray'}, {name: 'foo[0]', type: 'Field'}, {name: 'foo[1]', type: 'Field'}]`.
- When the form is submitted, `reduxForm.submit` iterates over the registeredFields and mark them as `touched`
- Since store state fields for `foo` has not been created yet, it sets `foo.touched = true`

Problem 1: Subsequent setting of `foo[0].touched = true` fails, as `foo` is an object

Problem 2: Subsequent swapping of `foo[0]` and `foo[1]` fails, as ARRAY_* reducer uses `splice` on `foo`, which is invalid for object

To see the weird behavior, go to http://redux-form.com/6.4.3/examples/fieldArrays/ and click submit immediately. See that `At least one member must be entered` although it uses the undocumented FieldArray.meta.touched prop.

Then refresh the page, this time click Add Member, focus on the first name text box, delete member and click Submit. See that `At least one member must be entered` do not show, which is the correct behavior. The reason being on focus of `foo[0]`, the field for `foo` is initialized as array. On submit, setting of foo.touched = true fails, as expected

gustavohenke:
I could not check this behavior looking at this example.

1. The `validate` function actually requires you to add at least one member, so it's correct to show that message;
2. Although `fields.members` is `{ touched: true }` at this moment, as soon as I click "Add member" it becomes `[{}]`;
3. If I remove such member, then the state becomes `[]` and the error message won't appear again.

huan086:
This is so hard to replicate. Let me try again

- Add 2 members
- Click submit
- at this point members become an object
- type the first name to set members[0].valid
- call fields.swap to swap the 2 members
- the valid prop is now incorrect, as splicing the fields fail

stunaz:
was able to reproduce the issue

erikras:
@huan086 Can you fix this merge conflict?

huan086:
done

erikras:

EvilOatmeal:
I understand that this fix was needed but it was a breaking change for me and I've spent about an hour now searching for the cause behind why my FieldArray error messages didn't show up.

It would be helpful if you categorized/highlighted bug fixes that are potentially breaking changes in a clearer way in the release notes. And any bug fix that changes a behaviour that people rely on as a feature ought to be listed as a potentially breaking change if they don't warrant a major SemVer release.

There is a warning in this release's notes ("This last one could potentially..."), but this bug was the second to last and I don't always pay much attention to the bug fixes. Also the warning triangle emoji should probably prefix each list item that the warning note is for and not just the note itself so there is a clearer association.

PR #2497, opened 2017-01-24
Clarify what to do when StackOverflow questions get no answers

gustavohenke:


gustavohenke:
It seems like CodeClimate need some love in order to understand a documentation PR...

erikras:
Yeah, I connected code climate, but never got around to doing the work to actually make it work...

Issue #883, opened 2016-04-29
v6 form onChange


aight8:
I would like to listen if the form changes some value. Then submit myself the form at every change (there are not text inputs but checkboxes etc.) I look at the easiest solution.. but I figured out that I have to listen at Field component level on changes, wrap the field props onChange and execute the handleSubmit method after it executes the fields onChange method one.

Maybe we can introduce a onChange method in the reduxForm configuration?

So #806 this issue could be solved too with this

erikras:
So the form does a deep-equal check in `componentWillReceiveProps` and if the values have changed and it has an `onChange` prop, it calls `this.props.onChange(values)`?

Seems doable.

aight8:
yap when it not eats to much performance. Don't know exactly the inner architecture, but already analyzing it :)

Currently this is my workaround:
```
<Radio
   {...field}
   onChange={value => {
      field.onChange(value);
      handleSubmit();
   }}
   options={options}
 />
```
To every field.

aight8:
Oh no, it not works this way neither. When handleSubmit executes the onSubmit in the reduxForm it receives the old values before ```field.onChange(value);```is called.

An idea how to solve that? 🎃

erikras:
This is a common mistake. I really should make a FAQ about this. It looks reasonable and declarative, but `handleSubmit` submits the values as they currently are in the `props`, and the `props` don't get updated from Redux until the next process tick. You have two options; pick that which least disgusts you.
```js
<input {...field} onChange={event => {
  field.onChange(event)
  setTimeout(handleSubmit)
}}/>
```
or
```js
<input {...field} onChange={event => {
  field.onChange(event)
  handleSubmit(values => {
    this.props.onSubmit({
      ...values,
      myField: event.target.value // <--- manually insert this field's value into submitted json
    })
  })
}}/>
```

paulshen:
Thanks for the reply! In case you're following along (like I was), the last snippet should be

```javascript
handleSubmit(values => {
  this.props.onSubmit({
    ...values,
    myField: event.target.value
  });
})(); // add parentheses here
```

erikras:
Thanks. Updated my comment.

nktssh:
@erikras thanks for sharing this. You should definitely add this in FAQ!
But what if I want to fire callback only if the value is valid (e.g. no `error` passed to component via `validate` function)? How I can handle this situation?
Manually calling the validate function seems a little bit overhead...

calvinfroedge:
[...]

PR #2530, opened 2017-01-30
Proposal: `Form`-level `onChange` implementation

egeste:
This code is very simple, and it should be easy for anyone to understand. The unit testing framework in place, however, is daunting. If this proposal is acceptable, I would appreciate help getting coverage for this behavior. It is based off of https://github.com/erikras/redux-form/issues/883#issuecomment-215732308

### DIFF
--- a/src/Form.js
+++ b/src/Form.js
@@ -16,10 +16,8 @@ class Form extends Component {
   componentWillReceiveProps(nextProps) {
     if (nextProps.onChange) {
       const currentValues = this.context._reduxForm.getValues()
-      if (this.previousValues) {
-        const currentValuesDiffer = !deepEqual(currentValues, this.previousValues)
-        if (currentValuesDiffer) nextProps.onChange(currentValues)
-      }
+      const currentValuesDiffer = !deepEqual(currentValues, this.previousValues)
+      if (currentValuesDiffer) nextProps.onChange(currentValues)
       this.previousValues = currentValues
     }
   }
--- a/src/Form.js
+++ b/src/Form.js
@@ -1,3 +1,4 @@
+import deepEqual from 'deep-equal'
 import React, { Component, PropTypes } from 'react'
 
 class Form extends Component {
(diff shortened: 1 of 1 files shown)


egeste:
In practice, it seems like `onComponentWillReceiveProps` only gets called for certain fields. For example, in my application, I have several `Field` wrapped components that invoke `this.props.input.onChange` directly with their `value`. For simple component wrappers like `WrappedInput` it works fine, but for more complex ones like `WrappedFileUpload` or `WrappedSelect` (which I'm sorry, I can't share the implementation for yet), `componentWillReceiveProps` does not get invoked. It could be an implementation problem on my end, but I don't think so.

egeste:
ping @erikras

src/Form.js
```diff
@@ -12,6 +13,17 @@ class Form extends Component {
     this.context._reduxForm.registerInnerOnSubmit(this.props.onSubmit)
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.onChange) {
+      const currentValues = this.context._reduxForm.getValues()
+      if (this.previousValues) {
+        const currentValuesDiffer = !deepEqual(currentValues, this.previousValues)
+        if (currentValuesDiffer) nextProps.onChange(currentValues)
```
gustavohenke: I would call `onChange` even for the first call (when `this.previousValues` is falsy)

egeste:
@gustavohenke if you can hack the unit tests in for this, I would be grateful.

egeste:
Another thought - it's probably advisable to always cache `this.previousState`, as adding an `onChange` prop during some point in the component lifecycle will dispatch an errant `onChange` invocation from `undefined` to `currentValues`

erikras:
This is not a bad idea, but you've implemented it in the wrong place.

I've implemented it in #2576.

erikras:
Fix published in [`v6.6.0`](https://github.com/erikras/redux-form/releases/tag/v6.6.0).

Issue #2466, opened 2017-01-17
FieldArray remove and push immediately cause view not rerender


sckoh:
I have a method that remove a field from FieldArray and immediately add a field into FieldArray but the view is not rerender. 
The state in redux is already updated to the new field when I look at the redux logger log.

```
  fields.remove(0);
  fields.push({ postId: 1 });
```

When I press on one of the input in the FieldArray, the FieldArray will only rerender.

After I read through the source code of ConnectedFieldArray.js,
I found that the shouldComponentUpdate method is preventing the FieldArray to rerender

```
const propsToNotUpdateFor = [
  '_reduxForm',
  'value'
]
```

```
    shouldComponentUpdate(nextProps) {
      const nextPropsKeys = Object.keys(nextProps)
      const thisPropsKeys = Object.keys(this.props)
      return nextPropsKeys.length !== thisPropsKeys.length || nextPropsKeys.some(prop => {
        // useful to debug rerenders
        // if (!plain.deepEqual(this.props[ prop ], nextProps[ prop ])) {
        //   console.info(prop, 'changed', this.props[ prop ], '==>', nextProps[ prop ])
        // }
        return !~propsToNotUpdateFor.indexOf(prop) && !deepEqual(this.props[ prop ], nextProps[ prop ])
      })
    }
```

the value of the FieldArray's Field has changed, but it is returning false due to this part" !~propsToNotUpdateFor.indexOf(prop)"

sckoh:
the value of the FieldArray's Field has changed, but it is returning false due to this part" !~propsToNotUpdateFor.indexOf(prop)"

sckoh:
@erikras 
May I know what is the reason not rerender the FieldArray when value is changed?
I think that if the value is an object and it is changed, we should rerender the FieldArray right?

dtipson:
I think I'm having a similar issue: deleting the item at index 0 of a FieldArray using ArrayRemove is resulting in the right input in the redux store, but wacky, mismatched display on the page (deleted item remains, items below it end up with mismatched props!)

ashyap:
I'm also experiencing the same issue. Any workaround for this ?

ashyap:
I found a workaround for this one. Just dispatch the arrayPush function instead of using fields.push

```
fields.remove(index)
dispatch(arrayPush('form', 'field', productToAdd))
```

irvinlim:
[...]

Issue #1578, opened 2016-08-22
Why does FieldArray have to iterate all fields on every keypress?


JakeLin1024:
Take the sample code as an example(slightly modified)

```
const renderMembers = ({ fields }) => {
  return (<ul>
    <li>
      <button type="button" onClick={() => fields.push({})}>Add Member</button>
    </li>
    {fields.map((member, index) =>{ 
      console.log("It's iterating members",index)  <==== here
      return (
        <li key={index}>
        <button
          type="button"
          title="Remove Member"
          onClick={() => fields.remove(index)}/>
        <h4>Member #{index + 1}</h4>
        <Field
          name={`${member}.firstName`}
          type="text"
          component={renderField}
          label="First Name"/>
        <Field
          name={`${member}.lastName`}
          type="text"
          component={renderField}
          label="Last Name"/>
        <FieldArray name={`${member}.hobbies`} component={renderHobbies}/>
       </li>)
    }
    )}
  </ul>
)}
```

lets say we have two members. Once you type the first name, console is like:
```
It's iterating members,0
It's iterating members,1
```
Since I dont know that much of the codebase, could you explain it a little bit?

erikras:
Excellent catch! This was a bug. Fixed by #1590. I will comment here again when the fix is released.

erikras:
Published in [`v6.0.1`](https://github.com/erikras/redux-form/releases/tag/v6.0.1).

PR #2544, opened 2017-02-02
Fix shouldComponentUpdate in FieldArray

irvinlim:
This fixes what is mentioned in #2466.

Because of fixes introduced in #1578, FieldArray does not re-render when the `value` changes. However this causes the FieldArray to not render updates for swapping, insertion, or other operations on the `value`, which is undesired behaviour.


erikras:

nayemmajhar:
No. fields.swap or fields.move do not rerender fields with new value.

nayemmajhar:
Sorry, it's working fine. (y)

vdanylov:
@nayemmajhar I'm still has this error...I'm using d&d and store changes but component doesn't(

PR #2575, opened 2017-02-10
Fix for issue 2567

### LINKED ISSUE #2567: pure prop passed to a redux form at runtime not working.
The documentation states:

> IMPORTANT: All of these configuration options may be passed into reduxForm() at "design time" or passed in as props to your component at runtime.

However, the `shouldComponentUpdate` method of Form (line 241 in [reduxForm.js](https://github.com/erikras/redux-form/blob/master/src/reduxForm.js) in version 6.5.0)
reads: `if (!config.pure) return true;`, completely ignoring the pure prop.

d-leeg:
reduxForm.js ShouldComponetUpdate now uses this.props.pure instead of config.pure to decide whether to not update.

Fixes #2567.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -238,7 +238,7 @@ const createReduxForm =
           }
 
           shouldComponentUpdate(nextProps) {
-            if (!config.pure) return true
+            if (!this.props.pure) return true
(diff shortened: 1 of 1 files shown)


erikras:

Issue #2427, opened 2017-01-10
validation fails after reinitialization


lechneal:
Hi, first of all congrats to this great component. Great work!

I'm combining the two website examples "Initializing from State" and "Submit Validation".
I'm using the submitted data to reinitialize the form.

The form is correctly reinitialized with the submitted data. 
However, after reinitializing, the validation function is called with the **old** initialValues. I end up with sync errors, although the data is valid.

The action sequence is this:
```
@@redux-form/START_SUBMIT
@@test/UPDATE_DATA (here I'm updating the state, which is used as initialValues)
@@redux-form/INITIALIZE
(in between these 2 actions, the validation function is called, with the **old** initialValues, resulting in sync errors)
@@redux-form/UPDATE_SYNC_ERRORS
@@redux-form/STOP_SUBMIT
@@redux-form/SET_SUBMIT_SUCCEEDED
```

It seems that the wrong values are passed in the nextProps in validateIfNeeded:
[reduxForm.js#L164](https://github.com/erikras/redux-form/blob/f8b9dcb86395427c3b2123f97131925e479ca2e4/src/reduxForm.js#L164)

- nextProps.initialValues is correctly set to the updated initalValues.
- The problem: nextProps.values equals to this.props.initialValues. 
nextProps.values should either be empty or equal to the updated initialValues

<img width="876" alt="screen shot 2017-01-10 at 11 28 17" src="https://cloud.githubusercontent.com/assets/17518288/21803627/2bd8b204-d72b-11e6-8f63-3c65581ce8e2.png">

<img width="947" alt="screen shot 2017-01-10 at 11 28 40" src="https://cloud.githubusercontent.com/assets/17518288/21803609/1839c6e8-d72b-11e6-83f1-1d8001807be3.png">

<img width="943" alt="screen shot 2017-01-10 at 11 28 55" src="https://cloud.githubusercontent.com/assets/17518288/21803633/2f737f84-d72b-11e6-9318-c88bb2ee297c.png">

Code

```
const initialTestState = {
    data: {
        username: ''
    }
};

const validate = values => {
  const errors = {}
  if (!values.username) {
    errors.username = 'Required'
  }
  return errors;
}; 

function submit(values, dispatch) {
  return sleep(1000) // simulate server latency
    .then(() => {
      dispatch(updateTestDataAction(values));
    })
};

let SyncValidationForm = reduxForm({
  form: 'syncValidation',  
  validate,                
  enableReinitialize: true
})(FormComponent);

[...]

Issue #2153, opened 2016-11-21
[Regression 6.2.0] initialValues + sync validation results in validating stale data


AVVS:
https://github.com/erikras/redux-form/blob/ac3b6da74b6939f621b7e4c659eec0af451395d9/src/reduxForm.js#L207-L208

Looking at this, when we have sync validate & initialValues setup the order of actions is the following:

1. redux-form/INITIALIZE dispatched, scheduling state update
2. redux-form/UPDATE_SYNC_ERRORS triggered on existing this.props.values, which are not the same as initialValues populated in the next rerender cycle

At this point if you press submit - form will fail with existing sync errors and will not run validation again at all (!), which prevents form from submitting

AVVS:
You can reproduce this with settings, apparently if I turn on touchOnBlur: true the above would work

```
    touchOnBlur: false,
    touchOnChange: true,
```

AVVS:
After downgrading to version 6.1.1 I can confirm that this problem does not persist any longer

AVVS:
@erikras this is still the case, sadly I cant find an easy way to fix past 6.2.0, can you suggest anything?

vintikjeny:
i have the same issue. the only workaround i've found so far is to explicitly invoke form `reset()` after submit success

shikelong:
+1

gustavohenke:
Because this issue saw almost no activity for almost a year, I'm closing it.

PR #2576, opened 2017-02-10
Added form-wide onChange functionality

### LINKED ISSUE #883: v6 form onChange

erikras:
Was suggested and attempted in #2530.

Fixes #883.

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -235,6 +235,11 @@ const createReduxForm =
             this.validateIfNeeded(nextProps)
             this.warnIfNeeded(nextProps)
             this.submitIfNeeded(nextProps)
+            if (nextProps.onChange) {
+              if (!deepEqual(nextProps.values, this.props.values)) {
+                nextProps.onChange(nextProps.values)
+              }
+            }
           }
 
           shouldComponentUpdate(nextProps) {
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -2887,7 +2887,7 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
         form: 'testForm'
       })(Form)
 
-      const dom = TestUtils.renderIntoDocument(
+      TestUtils.renderIntoDocument(
         <Provider store={store}>
           <Decorated onChange={onChange}/>
         </Provider>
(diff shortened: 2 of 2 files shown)


fzaninotto:
Where is this fantastic new feature documented?

gustavohenke:
Uh, nowhere, at the moment!

jscinoz:
This has some odd interaction with initialValues when the new global onChange event is used to submit the form on change.

In the reduxForm decorator's mapStateToProps, there is the following:

```javascript
const shouldUpdateInitialValues = enableReinitialize && initialized && !deepEqual(initialValues, stateInitial)
const shouldResetValues = shouldUpdateInitialValues && !keepDirtyOnReinitialize

let initial = initialValues || stateInitial || empty

if (shouldUpdateInitialValues) {
  initial = stateInitial || empty
}

let values = getIn(formState, 'values') || initial

if (shouldResetValues) {
  values = initial
}
```

Specifically:


```javascript
if (shouldUpdateInitialValues) {
  initial = stateInitial || empty
}
```

This always uses the state values, which are stale as initialize() is not called until reduxForm's componentWillReceiveProps (i.e. the component rendered by the connect() HOC). The end result is an infinite update loop (and the tab eventually being killed by the browser) as values & initialValues will continuously alternate.

src/reduxForm.js
```diff
@@ -235,6 +235,11 @@ const createReduxForm =
             this.validateIfNeeded(nextProps)
             this.warnIfNeeded(nextProps)
             this.submitIfNeeded(nextProps)
+            if (nextProps.onChange) {
```
egeste: 👍

timhwang21:
@erikras just wondering if there is a minor version release planned soon. This patch in particular would be really useful to me. As usual, thanks for the great library!

erikras:

j-machado:
I'm having the same issue as @jscinoz in 6.6.0, can't use `enableReinitialize` with `onChange`. 

For now, I'm using an ugly flag to skip unwanted calls to `onChange`. Did anyone had a better solution for this?

ravicious:
@jscinoz @j-machado: I think it may be related to #2427 and #2153.

PR #2584, opened 2017-02-13
Avoided unnecessary call to sync validation

### LINKED ISSUE #2583: Validation is triggered twice on form init
Bug ?

### What is the current / expected behavior?
In `Synchronous Validation Example` validate function is trigged twice, before form is mounted and after it's rendered.

Since nothing was changed in the form, why do we need to trigger it twice, is it by design?

### What's your environment?
Redux form 6.5.0

erikras:
Fixes #2583.

### DIFF
--- a/src/__tests__/FieldArray.spec.js
+++ b/src/__tests__/FieldArray.spec.js
@@ -925,9 +925,9 @@ const describeFieldArray = (name, structure, combineReducers, expect) => {
 
       renderArray.calls[ 0 ].arguments[ 0 ].fields.push('rat')
 
-      // should validate twice, once for new field, and again to validate new field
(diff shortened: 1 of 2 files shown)


erikras:

KatFishSnake:
@erikras can you please look into @stream7 referenced issue its breaking for multiple people

dizlexik:
I'll second that, @KatFishSnake.

@erikras, issue #2971 is a critical failure of the reset function and @stream7 seems to have tracked it down to this PR. I'm not in any position right now to rollback/build/test but when you get a chance could you please take a look?

yohnstoppable:
Anyone have a workaround to fix this? Still doesn't work in 7.0.3

PR #2586, opened 2017-02-13
Added submitFailed to FieldArray meta prop

### LINKED ISSUE #2581: Field Arrays bug example
Bug in this [example](http://redux-form.com/6.5.0/examples/fieldArrays/)

The `members` Fields Array doesn't display the global error(`'At least one member must be entered'`) after submit.

It should display the global error after submit.

### What's your environment?
Redux Form version: v6.5.0
OS/browser: Mac OS/ Chrome

### Other informations
In the new release the FieldArray is not touched after submit(https://github.com/erikras/redux-form/pull/2473), this is the reason why is not displaying the global error.
Maybe passing the prop `submitFailed` to the `ConnectedFieldArray` could solve the problem, displaying the error only if there is an error and the form was submitted.

erikras:
Fixes #2581, caused by #2473.

### DIFF
--- a/examples/fieldArrays/src/FieldArraysForm.js
+++ b/examples/fieldArrays/src/FieldArraysForm.js
@@ -12,11 +12,11 @@ const renderField = ({ input, label, type, meta: { touched, error } }) => (
   </div>
 )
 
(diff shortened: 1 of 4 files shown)


erikras:

PR #2592, opened 2017-02-14
Update asyncValidating PropType to match documentation

Limess:
This is currently set to `bool` whereas it is a string if a field triggers async validation.

See
http://redux-form.com/6.5.0/docs/api/Props.md/#-asyncvalidating-string-boolean-

Current PropType violation:
propType violation


erikras:

Issue #2176, opened 2016-11-28
selector for getFormSyncWarnings


jole78:
Any ideas how I would write such as selector or if it could be exported?

gustavohenke:
If it was baked into the library, it would be much easier... 💭

erikras:
Fix published in [`v6.6.0`](https://github.com/erikras/redux-form/releases/tag/v6.6.0).

PR #2624, opened 2017-02-23
Added selector for getFormSyncWarnings.

evarouss:
Added a new selector to the list: `getFormSyncWarnings` as suggested [here](https://github.com/erikras/redux-form/issues/2176). Implementation is the same as `getFormSyncErrors`.

### DIFF
--- a/docs/api/Selectors.md
+++ b/docs/api/Selectors.md
@@ -21,6 +21,7 @@ import {
   getFormInitialValues,
   getFormSyncErrors,
   getFormAsyncErrors,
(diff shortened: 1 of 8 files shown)


erikras:
Just when I think there's nothing more to have a selector for... 😆

Can you modify the docs, too?

evarouss:
@erikras ahah, thanks for reviewing! I've documented the new method in `docs/api/Selectors.md` as part of this PR. Is there more docs where the selectors are referenced?

erikras:

PR #2645, opened 2017-03-02
(v5) Changes to members of top-level array field should change instance identity

asazernik:
React components that take just the array field still need to re-render.
Existing tests covered this case, but explicitly asserted the incorrect
behavior.

Specifically "top-level" array fields, because array fields below the
top level change their identity based on the return value rather than
modifications to the fields object.

Copied the control flow from the object-field path, which Does The Right
Thing.

### DIFF
--- a/src/__tests__/readField.spec.js
+++ b/src/__tests__/readField.spec.js
@@ -570,7 +570,7 @@ describe('readField', () => {
     const afterArray = fields.foo;
     const after1 = fields.foo[0];
     const after2 = fields.foo[1];
-    expect(beforeArray).toBe(afterArray); // array should be same instance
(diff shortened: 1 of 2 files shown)


erikras:
I'm not really too concerned with `v5` anymore, so I'm not examining this in detail, but I'll merge it and perhaps release it eventually.

morgante:
@erikras How can you not be concerned with v5 anymore? v6 is a huge migration and is literally impossible on many projects because it breaks with so many other libraries.

gustavohenke:
Erik is probably out of time to maintain an old branch which barely gets requests nowadays.
If you get enough significative contributions landed on the v5 branch, we'll very likely publish new releases.

PR #2648, opened 2017-03-03
Fix #2647 ("pristine" not calculated using deep equality)

### LINKED ISSUE #2647: "pristine" calculation only does a shallow comparison
bug report

### What is the current behaviour?
[This line](https://github.com/erikras/redux-form/blob/master/src/ConnectedField.js#L216) only does a "shallow" comparison between `value` and `initial` and yet the docs state that values can be any value.

If the values are object-based (e.g. Object, Array, etc.), then the calculation produces incorrect behavior, because a pointer equality is performed and not a deep equality operation.

Perform deep equality operation.

### What's your environment?
N/A

### Other informations
N/A

begincalendar:
This is a fix for the issue: #2647.


begincalendar:
I don't know why the Travis build says it failed, when it seems to have completed successfully in the logs.

erikras:
Great catch. 👍

erikras:

Issue #2667, opened 2017-03-08
Submit validation errors in FieldArray are initialised with empty


a-xin:
### Are you submitting a **bug report** or a **feature request**?
Bug

### What is the current behavior?
- Create a `FieldArray` with the possibility to add fields (like in the FieldArray example)
- Add a field
- Use Submit Validation and throw an error for the submitted field
  ```
  submitErrors: {
    subfields: ['invalid input']
  }
  ```
- Add an additional field to the beginning of the `FieldArray`
=> The additional field receives a submit error `empty` (ie. `{ }`)
  ```
  submitErrors: {
    subfields: [{ }, 'invalid input']
  }
  ```

### What is the expected behavior?
The additional field receives a submit error `undefined`
```
submitErrors: {
  subfields: [undefined, 'invalid input']
}
```

gustavohenke:
Fixed by #2668

erikras:
Fix published in [`v6.6.0`](https://github.com/erikras/redux-form/releases/tag/v6.6.0).

PR #2668, opened 2017-03-08
Fix initial submitError state for FieldArrays

a-xin:
When adding a new field to a `FieldArray` (which already has a field with a `submitError`), the `submitError` for the new field is initialised with `empty`.

This breaks form validation, as the new field has not been submitted yet and therefore should not have an initial `submitError`.

Therefore changing the initial value for `submitError` and `asyncError` of new fields to `undefined`.

Issue: https://github.com/erikras/redux-form/issues/2667

### DIFF
--- a/src/__tests__/reducer.arraySplice.spec.js
+++ b/src/__tests__/reducer.arraySplice.spec.js
@@ -13,6 +13,63 @@ const describeArraySplice = (reducer, expect, { fromJS }) => () => {
       })
   })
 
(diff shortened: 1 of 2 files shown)


a-xin:
Hey @erikras, can you take a look at this one? It's currently causing quite some issues for me, would be great to get this fixed

erikras:
Looks great. What a simple fix. 👍

erikras:

PR #2680, opened 2017-03-13
Bound dispatch to blur and change in props passed to onSubmit

### LINKED ISSUE #2663: `change` prop is unbound
bug report

In a `handleSubmit` function which looks like:
```js
handleSubmit(data, dispatch, { updateAccount, change, untouch }) {
  return updateAccount(data)
    .then(() => {
      dispatch(change("password", null));
      dispatch(change("current_password", null));
      untouch("password", "current_password");
    })
    .catch(parseFormErrors);
}
```
I had to dispatch `change` function in order to actually fire it. Just calling `change("current_password", null)` does nothing, moreover while [docs](http://redux-form.com/6.5.0/docs/api/Props.md/) stats that `change` is a bound function which returns nothing, `console.log(change("password", null))` outputs:
```
Object
  meta: {form: "AccountForm", field: "password", touch: false, persistentSubmitErrors: false}
  payload: null
  type: "@@redux-form/CHANGE"
```

`change` should be bound, so I can just call it to fire an action, like how I do with `untouch`.

### What's your environment?
```
"redux-form": "^6.5.0"
Node: v7.4.0
Mac OS latest, with latest versions of Firefox & Safari
```

erikras:
Fixes #2663.


erikras:

PR #2695, opened 2017-03-15
Update GettingStarted.md

xmas11:
There are two different 'things' named `handleSubmit` in this example.
First one is a custom handler defined on `ContactPage` component, whereas second one is name of property injected by `reduxForm` hoc.
This may cause confusion, especially due to structure of this example:
`<ContactForm onSubmit={this.handleSubmit} />` is only slightly different then `<ContactForm handleSubmit={this.handleSubmit} />`.
My proposal is to change name of custom handler method from `handleSubmit` to `submit`.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -85,13 +85,13 @@ function.
 import ContactForm from './ContactForm';
 
 class ContactPage extends React.Component {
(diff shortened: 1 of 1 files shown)


erikras:
Sure. Seems okay.

PR #2707, opened 2017-03-19
Update propTypes.js

mqklin:
I tried to use `forbidExtraProps` from airbnb-prop-types and it gave me this errors:
`The prop initialValues is marked as required in MyComponent, but its value is undefined.`
`Failed prop type: MyComponent: unknown props found: pure, triggerSubmit, clearSubmit, submit`
So I found these in docs except  `triggerSubmit` and `clearSubmit`. If you say me what they does, I'll add them to PR.

### DIFF
--- a/src/propTypes.js
+++ b/src/propTypes.js
@@ -13,8 +13,9 @@ const propTypes = {
   form: string.isRequired,            // the name of the form
   invalid: bool.isRequired,           // true if there are any validation errors
   initialized: bool.isRequired,       // true if the form has been initialized
-  initialValues: object.isRequired,   // the initialValues object passed to reduxForm
+  initialValues: object,              // the initialValues object passed to reduxForm
   pristine: bool.isRequired,          // true if the values are the same as initialValues
+  pure: bool.isRequired,              // if true, implements shouldComponentUpdate
   submitting: bool.isRequired,        // true if the form is in the process of being submitted
   submitFailed: bool.isRequired,      // true if the form was submitted and failed for any reason
   submitSucceeded: bool.isRequired,   // true if the form was successfully submitted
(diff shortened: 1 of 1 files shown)


erikras:
Thanks!

erikras:

albermav:
I am getting this error now:
Failed prop type: The prop `triggerSubmit` is marked as required in `Form`, but its value is `undefined`.
in my forms.
Does anyone has the same problem?

jeffberry:
Was this PR reviewed? I'm curious as to how the author of this PR did not run into the same errors as @albermav - we as well are now getting the same warning as @albermav all of our codebase. 

This `triggerSubmit` prop is defined two other places in the code as `triggerSubmit: PropTypes.bool` Note that it is defined as an optional boolean. This PR defines it as a required function. 

This PR should be reverted and/or amended to mark these as optional booleans.

mqklin:
Of course I tried this code in my project, and it worked well. 
I wrote in the first comment that I hadn't found what those props make.
I created another PR that removes `isRequired` from `triggetSubmit`.

jeffberry:
That's interesting - what function are you providing as a prop to your form components then? The intended use of this property seems to be a boolean, so marking it as a function (required or not) is still incorrect from what I can see. Please correct me if I'm wrong.

mqklin:
Oh, thank you, changed it to bool. I don't provide `triggerSubmit` or `clearSubmit` function, but my code works well. Maybe `validation` function makes this. 
Seems like you are better in this problem - maybe you create better PR?

jeffberry:
Your other PR looks fine! Thanks!

erikras:
Published other fix in [`v6.6.1`](https://github.com/erikras/redux-form/releases/tag/v6.6.1).

PR #2720, opened 2017-03-23
add babel transform runtime to fix ie 11 issues

jptconiconde:


jerrysu:
Another solution would be to avoid the use of `for-of` in the library if there's any interest in keeping the library smaller and more universal without the runtime polyfill.

Like, I think using `Array.prototype.find` works in `getError`:

```
const getError = (value, values, props, validators) => {
  return toArray(validators).find((validator) => {
    const error = validator(value, values, props)
    if(error) {
      return error
    }
  });
}
```

jptconiconde:
I agree, in that case, there are other places that we have to update then.

erikras:
Thanks for this. Will replace `for...of`'s. #2724

PR #2721, opened 2017-03-23
Fix #2718 (Incorrect way of accessing syncErrors, syncWarnings in ConnectedField, ConnectedFieldArray, ConnectedFields for redux-form/immutable)

### LINKED ISSUE #2718: Incorrect way of accessing syncErrors, syncWarnings in ConnectedField, ConnectedFieldArray, ConnectedFields for redux-form/immutable
**bug report**

1. Create simple field with _Field-level validator_ using `redux-form/immutable`.

```
import { Field } from 'redux-form/immutable';

const required = value => (value ? undefined : 'Required');
const renderField = ({ input, type, meta: { error } }) => {
  return (
    <div>
      <div>
        <input {...input} type={type} />
        {error && <span>{error}</span>}
      </div>
    </div>
  );
};

<Field name="username" type="text" component={renderField} validate={required} />
```

2. Type something
3. Remove content from input
4. `@@redux-form/UPDATE_SYNC_ERRORS` event appears with `syncErrors: {username:'Required'}` payload
5. **No error message is visible**

`Required` string is assigned to `meta.error` variable.

Currently `error` is `undefined`.

### What's your environment?
`redux-form: 6.5.0`

### Other informations
In `ConnectedField.js` during the `getSyncError` call, `syncErrors` contains valid `{username:'Required'}` object as an `Immutable.Map`

### What's your environment?
`redux-form: 6.5.0`

### Other informations
In `ConnectedField.js` during the `getSyncError` call, `syncErrors` contains `{username:'Required'}` **Immutable.Map* object.

Meanwhile in https://github.com/erikras/redux-form/blob/master/src/ConnectedField.js#L15 you are extracting error using `plain.getIn(syncErrors, name)` instead of `getIn(syncErrors, name)`

The same issue occurs for `syncWarnings` in all of `ConnectedField*` files:

- `ConnectedField.js`
- `ConnectedFieldArray.js`
- `ConnectedFields.js`

This breaks validation in general when using immutable.

hsz:
Use `getIn` instead of `plain.getIn` in case we're using `redux-form/immutable`.

Ref. https://github.com/erikras/redux-form/issues/2718

### DIFF
--- a/src/ConnectedFields.js
+++ b/src/ConnectedFields.js
@@ -11,14 +11,14 @@ const propsToNotUpdateFor = [
 const createConnectedFields = ({ deepEqual, getIn, toJS, size }) => {
 
   const getSyncError = (syncErrors, name) => {
-    const error = plain.getIn(syncErrors, name)
+    const error = getIn(syncErrors, name)
     // Because the error for this field might not be at a level in the error structure where
     // it can be set directly, it might need to be unwrapped from the _error property
     return error && error._error ? error._error : error
   }
 
   const getSyncWarning = (syncWarnings, name) => {
-    const warning = plain.getIn(syncWarnings, name)
+    const warning = getIn(syncWarnings, name)
     // Because the warning for this field might not be at a level in the warning structure where
     // it can be set directly, it might need to be unwrapped from the _warning property
     return warning && warning._warning ? warning._warning : warning
@@ -70,7 +70,7 @@ const createConnectedFields = ({ deepEqual, getIn, toJS, size }) => {
         }, {})
       }
     }
-    
+
     shouldComponentUpdate(nextProps) {
       const nextPropsKeys = Object.keys(nextProps)
       const thisPropsKeys = Object.keys(this.props)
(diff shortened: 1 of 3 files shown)


src/ConnectedFields.js
```diff
@@ -70,7 +70,7 @@ const createConnectedFields = ({ deepEqual, getIn, toJS, size }) => {
         }, {})
       }
     }
-
```
hsz: Just extra whitespaces removed from the end of the line by the IDE.

src/ConnectedFields.js
```diff
@@ -159,7 +159,7 @@ const createConnectedFields = ({ deepEqual, getIn, toJS, size }) => {
       return {
         _fields: names.reduce((accumulator, name) => {
           const initialState = getIn(formState, `initial.${name}`)
-          const initial = initialState !== undefined ? initialState : (initialValues && getIn(initialValues, name))
```
hsz: Just extra whitespaces removed from the end of the line by the IDE.

erikras:
Nice work. This was vestigial from before we put the sync errors into the Redux state, and we allowed the validation function to return a regular object, even if you were using immutable.

hsz:
@erikras Thanks. And here comes the question - when do you plan to do the next release? I've migrated our product from 2.x to the current redux-form version, but the validation case is the latest one that blocks us from going live. I've tested this fix and previous validators work out of the box.

erikras:

PR #2731, opened 2017-03-24
Fix #2591: stopSubmit removes pending promise from reduxForm.

### LINKED ISSUE #2591: stopSubmit or setSubmitFailed should delete submitPromise
**bug report**

When using with redux-saga a good trick is to return an empty Promise from the onSubmit function, like so:
```javascript
const mapDispatchToProps = (dispatch) => {
    return {
        onSubmit: (value) => {
            dispatch({type: 'LOGIN_WITH_EMAIL', payload: {email: value.email, password: value.password}});

            return new Promise(() => { }); // sets form state to SUBMITTING
        }
    }
}
```

Then in the saga, if the api call on the server fails, one would simply dispatch `stopSubmit('LoginForm', { _error: error})`.

All of this works for one submit. Subsequent attempts to submit are ignored because the submit promise returned by onSubmit remains undeleted (https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L407). Since the submit has stopped, shouldn't the submitPromise be cleared too?

This happens in any redux-form 6.* version.

gretro:


gretro:
Please wait before approving the PR. I'm working on some unit tests to ensure this works properly.

gretro:
Ok, tests have been comited and are passing.

erikras:

PR #2738, opened 2017-03-26
Fix triggerSubmit/clearSubmit propTypes problems

### LINKED ISSUE #2744: Can we remove `triggerSubmit: func.isRequired` from `src/propTypes.js`?
`triggerSubmit` prop is undocumented ("// didn't find in the docs what is does"), but at the same time, it is marked as required. It is generating false warnings in the developer console since developers don't use this prop (at least I know of no actual, documented usage in the docs and no one mentioned it in the Issues yet.)

We should display no warnings for code developers can't do anything about. Perhaps we should remove `.isRequired` from props that are not required for the form to function properly.

My environment:
Redux-Form 6.5.0, 6.6.0, Chrome (OS X), node 7.7.4

mqklin:
fix https://github.com/erikras/redux-form/pull/2707 problems

Fixes #2744

### DIFF
--- a/src/propTypes.js
+++ b/src/propTypes.js
@@ -48,9 +48,9 @@ const propTypes = {
   submit: func.isRequired,            // action to trigger a submission of the specified form
   untouch: func.isRequired,           // action to mark fields as untouched
   
-  triggerSubmit: bool, // didn't find description in the docs
-  clearSubmit: func // didn't find as well
-  
+  // triggerSubmit
+  triggerSubmit: bool,                // if true, submits the form on componentWillReceiveProps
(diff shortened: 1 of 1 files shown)


btimo:
**triggerSubmit** seems to be a **bool** reporting the state of a form when submitting it.
Multiple proofs available in the source: [](https://github.com/erikras/redux-form/search?utf8=%E2%9C%93&q=triggerSubmit&type=)

Empact:
Oh hey didn't check and made: #2747 you could copy the doc comments from there if you like.

mqklin:
Added your comments.

erikras:

PR #2748, opened 2017-03-28
Change the form-level onChange to accept the same args as onSubmit

Empact:
This enables common use cases such as submitting the form on change, and
specifically re-using the same action for submit and change.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -2903,6 +2903,8 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
       expect(onChange.calls.length).toBe(1)
 
       expect(onChange.calls[0].arguments[0]).toEqualMap({ foo: 'dog' })
+      expect(onChange.calls[0].arguments[1]).toBeA('function')
+      expect(onChange.calls[0].arguments[2].values).toEqualMap({ foo: 'dog' })
 
       changeBar('cat')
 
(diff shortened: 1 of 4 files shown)


Empact:
Incidentally it seems a reasonable api refinement would be to just pass `props` into `onChange` and `onSubmit`, as the others values can be extracted from props with destructuring.

src/__tests__/reduxForm.spec.js
```diff
@@ -2911,22 +2913,28 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
...
+      })
 
       changeFoo('dog')
 
       // onChange NOT called since value did not change
       expect(onChange.calls.length).toBe(2)
-      expect(onChange.calls[1].arguments[0]).toEqualMap({
-        foo: 'dog',
-        bar: 'cat'
-      })
```
Empact: My thought was that this is not necessary to check as the calls have not grown.

src/__tests__/reduxForm.spec.js
```diff
@@ -3227,6 +3235,8 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
       expect(onSubmit).toHaveBeenCalled()
       expect(onSubmit.calls.length).toBe(1)
       expect(onSubmit.calls[ 0 ].arguments[ 0 ]).toEqualMap({})
+      expect(onSubmit.calls[ 0 ].arguments[ 1 ]).toBeA('function')
+      expect(onSubmit.calls[ 0 ].arguments[ 2 ].values).toEqualMap({})
```
Empact: Added testing of `onSubmit` args as well.

benigeri:
dope!

erikras:
Fantastic work, @Empact.

jaspersorrio:
How exactly do we use this new functionality?
Is there an example? Using 

```
// in SomeFormName class render method

const {
  handleSubmit,
  onChange
} = this.props;

<Form
onChange={
  
  handleSubmit((values)=>{ 
    console.log(values); //<-- gives me the old values
  })();

  onChange((v)=>{})() // <-- error. onChange is undefined
}>

...
...
...

</Form>
```

```

// in SomeFormName class render method
const {
  handleSubmit,
  onChange
} = this.props;

<Form
onChange={onChange}>

...
...
...

</Form>

// outside class

const change = (values)=>{
  console.log(values); // <-- returns event. As expected 
}

export default reduxForm({
  onChange:change
})(SomeFormName)
```

jaspersorrio:
In the end, i managed to get the latest form state by doing this. Not sure if it this was the intended way to use this api, but just sharing the way I was able to use it.
```
// inside class render method
const {
  handleSubmit
} = this.props;

return (
<Form>
  onChange={(event)=>{
    handleSubmit((values, dispatch, props)=>{

      const newValues = values.set(event.target.name, event.target.value);
      this.submitForm(newValues);

    })()
  }}>

  ...
  ...
  ...
</Form>
)
```

PR #2750, opened 2017-03-28
Fix 'renderHobbies' was used before it was defined eslint error

Empact:
Found by codeclimate.

I just switched the two method definitions.


gustavohenke:
Thanks!

PR #2755, opened 2017-03-29
Fix: hasError used plain.getIn(syncErrors, key) for Immutable instead of getIn(syncErrors, key)

hsz:
Because of using `plain.getIn`, the syncErrors are not respected during the form submit operation and form is submitted even if it is not valid.

Can we release it ASAP?


erikras:

PR #2772, opened 2017-04-04
clearSubmitErrors not exposed in lib

Nickman87:


erikras:

PR #2776, opened 2017-04-04
pass meta.form to Fields component

chirag04:
meta.form is undefined for `Fields` component. This pr adds the failing test case and the fix.


chirag04:
cc @erikras

PR #2779, opened 2017-04-05
Fixed inconsistent treatment of sync errors in redux-form/immutable

### LINKED ISSUE #2774: ARRAY_PUSH action with Immutable give Invalid keyPath error
I think this is a bug related with validate function and immutable, because if I don't pass validate function ARRAY_PUSH actions works fine.

screen shot 2017-04-04 at 16 04 40

At the moment I try to push on field array new data. In this case I am trying to initialize new condition with a default operator. So idea of final result in plain JS will be:
```js
{
  title: 'title',
  conditions: [
    {
       attribute: 'bla',
       operation: '===',
       value: 'true',
       operator: 'AND',
    },
    {
       attribute: 'bla',
       operation: '===',
       value: 'true',
       operator: 'AND',
    },
....
  ],
}
```

I expect when I try to add new data to list of conditions doesn't crash

### What's your environment?
redux-form: 6.6.1
redux: 3.6.0
redux-immutable: 4.0.0
immutable: 4.0.0-rc.2
material-ui: 0.17.0

Let me know if you need more info about it

### Other informations
After some research I "fix" it if I add fromJs to line 390 on reducer.js file
result = setIn(result, 'syncErrors', fromJS(syncErrors))
I detect that state store syncErrors as Immutable.Map but all inside of it is plain object so when you use getIn and want to access deeper this error shows because deeper key is plain object not Immutable data. So thats it is reason I add that fromJS, but after run tests I saw I break 7 tests so I have feeling that maybe it isn't real problem.

This is killing me because I can't continue until this is fixed, let me know how I could help you or if you need anything from my side.

erikras:
Fixes #2774.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux'
 import createFieldProps from './createFieldProps'
 import onChangeValue from './events/onChangeValue'
 import { dataKey } from './util/eventConsts'
+import plain from './structure/plain'
 
 const propsToNotUpdateFor = [
   '_reduxForm'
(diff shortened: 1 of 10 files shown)


erikras:

Issue #2785, opened 2017-04-05
wizard screen submission failing with no errors


mpham-r7:
i have a redux-form@6.6.1 that implements the wizard pattern and i'm having an issue... when submitting a screen (no syncErrors present), submission fails... i think i've tracked down the issue to the submit function in reduxForm.js

it seems the problem is that the value for (the closed over variable) validExceptSubmit at https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L445 could be different from the ~equivalent value in this.props.validExceptSubmit https://github.com/erikras/redux-form/blob/master/src/reduxForm.js#L442 ... i find that this is the case /w my code

i haven't managed to reproduce the behavior in a (simplified) fiddle as i'm not exactly sure what the form state causing this is, but in terms of the code itself, the callback on line 445 should reference the current props, rather than the closed over variable (which over time could diverge from the current props) right?

mpham-r7:
i tracked it down to here, looks like the intent was to refactor the code...

https://github.com/erikras/redux-form/commit/81ce486788bba5485b1d580c1c0fad359adc41cd#diff-d749bb1257f721af37259e0346e623beR403

mpham-r7:
made a pr for this: #2786

erikras:
Great 🕵️ work on this, @mpham-r7!

erikras:
Fix published in [`v6.6.3`](https://github.com/erikras/redux-form/releases/tag/v6.6.3).

PR #2786, opened 2017-04-05
fix (possibly stale) reference to validExceptSubmit variable

mpham-r7:
instead of referring to a possibly stale reference to validExceptSubmit, refer to this.props.validExceptSubmit... opened an issue /w more details here: #2785

### DIFF
--- a/src/reduxForm.js
+++ b/src/reduxForm.js
@@ -408,7 +408,7 @@ const createReduxForm =
           }
 
           submit(submitOrEvent) {
-            const { onSubmit, blur, change, dispatch, validExceptSubmit } = this.props
+            const { onSubmit, blur, change, dispatch } = this.props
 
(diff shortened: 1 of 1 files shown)


mpham-r7:
made a new issue with more detail and a plunkr demoing the issue here #2799 
@erikras i would appreciate if you took a look

erikras:
I just spent an hour debugging this problem and came to the exact same conclusion. Hard to believe such a complex problem is solved with such a tiny change. Programming is fun! 😀

Luckily it wasn't time completely wasted, as I've written a great test to demonstrate the bug and also demonstrate that this fixes it.

chirag04:
@erikras while we are at it, can you also release a new version?

erikras:
Published in [`v6.6.3`](https://github.com/erikras/redux-form/releases/tag/v6.6.3).

Issue #2768, opened 2017-04-03
[blur event] Why does handleBlur updates field value?


mereskin:
### What is the current behavior?

onBlur and onFocus callbacks are required to mark field as touched to display per-field validation. onBlur and onFocus assume DOM event is passed and current field value corresponds to value of the underlying DOM element.

That behavior is not working for custom components. If custom component is used in place of input the `value` property of DOM element that originates event can not be equal to that of a component itself. 

For example, we use `CheckBoxGroup` which has `value` proptype as `arrayOf(string)`. This component still emits onFocus and onBlur events when individual checkboxes are checked/unchecked and passes DOM event through. This DOM event is used by blur, value is extracted from it and this value is no longer an array but string `true`, since this value is read from `checked` property of a checkbox.

### What is the expected behavior?

onBlur event handler shouldn't update the value of the field and shouldn't assume that event object passed in is actually a DOM event. onBlur handler should update `touched` property of the field so that validation is still displayed.

Is there any reason why onBlur event updates form state? Can we use `onChange` events exclusively to update form state?

### What's your environment?
All environments

carlsverre:
I am also having an issue with Blur modifying the value of the field.  The problem for me is that onBlur assumes that the event contains the current value - which is not true when you batch react updates and the form is updated via autocomplete.  In this case the browser fires `onChange` and `onBlur` in the same frame and thus the `onBlur` will see the "last" value of the field since the HTML input hasn't actually had a chance to update.

Making `onBlur` not update the form value will resolve my problem.

carlsverre:
I am currently getting around this issue by calling the `onBlur` handler with no event argument.  This is supported by the tests and seems to do exactly what we want.  You can see the test here:

https://github.com/erikras/redux-form/blob/master/src/__tests__/reducer.blur.spec.js#L70

shodanuk:
Another workaround is to use the parse (and / or possibly normalize) prop of Field to force the type value of that the blur action sets into state.

In my case, redux-form was converting number values of my fields into strings, causing issues with custom react-numeric input fields. [Line 99 of ConnectedField](https://github.com/erikras/redux-form/blob/892f45b5325fa7274cb6f8f0e7b3ea617b960c29/src/ConnectedField.js#L99) calls onChangeValue which "normalises" the value of the field using a function passed into the parse or normalize prop of ConnectField. In the absence of these functions being specified, the value seems to default to a string.

For what's worth, this is all I passed into the parse prop:

```
<Field
[...]

PR #2791, opened 2017-04-07
[onBlur] Breaking: do not emit value in onBlur

mereskin:
Fixes https://github.com/erikras/redux-form/issues/2768.

Some Complex fields cannot emit DOMevent with correct value of e.target. 
Also, we couldn't identify a use-case where changing value on blur makes sense.

This PR cleans up `blur` action and removes value from the payload.  It also removes a bunch of tests. For asyncValidate it always emits with the last rendered value of the input field assuming that there was `onChange` preceding `onBlur`.

to: @erikras

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -95,15 +95,7 @@ const createConnectedField = ({ deepEqual, getIn, toJS }) => {
     }
 
     handleBlur(event) {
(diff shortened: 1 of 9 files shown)


mereskin:
PTAL @erikras

erikras:
Closing for reasons explained here: https://github.com/erikras/redux-form/issues/2768#issuecomment-292770517

PR #2796, opened 2017-04-07
Ignore non-RF actions

### LINKED ISSUE #2790: [6.6.1] reducer function registers form when non-redux-form action type dispatched to store
bug report

We have our own specialized react form implementation but we also use redux-form in our app. Our form implementation uses `action.meta.form` (just like redux-form) to send our form's name to our reducer. When we dispatch our `ecrForm\INIT_FORM` action type to the redux store with `action.meta.form = 'generatedForm'` redux-form's reducer also "registers" `generatedForm` as its form in its 'form' state slice.

We think the problem is in `src/reducer.js` file's `byForm` method. It retrieves the form name from `action.meta.form` and in the last line it  sets `generatedForm` to an empty Map in redux-form's state because `result` is empty (because we dispatched non-redux-form action) and `formState` is `undefined`.
```
  const byForm = (reducer) =>
    (state = empty, action = {}) => {
      const form = action && action.meta && action.meta.form
      // ...
      const result = reducer(formState, action)
      return result === formState ? state : setIn(state, form, result)
    }
```

Do not change state when non-redux-form action type dispatched to store.

### What's your environment?
Redux Form 6.6.1, Immutable store (3.8.1), React 15.4.2

erikras:
Fixes #2790.


src/reducer.js
```diff
@@ -446,13 +451,13 @@ const createReducer = structure => {
...
           .keys(reducers)
           .reduce((accumulator, key) => {
-            const previousState = getIn(accumulator, key)
-            const nextState = reducers[ key ](previousState, action, getIn(state, key))
-            return nextState === previousState ?
-              accumulator :
-              setIn(accumulator, key, nextState)
-          },
-          this(state, action)))
+              const previousState = getIn(accumulator, key)
```
mereskin: Is that whitespace intentional?

erikras: No consciously, no. Was added by my IDE, and ESLint doesn't mind.

Personally, I think I prefer that wrapped continuations of parameters (in this case an arrow function) be indented....but not very strongly.

Thank you for the eagle eye, though. I'd be open to hearing arguments against it.

PR #2797, opened 2017-04-07
Enable initialize in componentWillMount()

### LINKED ISSUE #1385: Validation errors not passed to fields after re-initialize when enableReinitialize is turned on
When passing the `enableReinitialize` flag all of the fields are getting being marked as valid even though there are errors present for them. It appears to be that when initialize is called a second time it does not trigger the `redux-form/UPDATE_SYNC_ERRORS` action and as such all of the fields don't get updated with their error state again.

When this flag is passed, I think the `redux-form/UPDATE_SYNC_ERRORS` action always needs to get triggered in this scenario and maybe a change in the logic of `updateSyncErrorsIfNeeded` is needed?

erikras:
Fixes #1385.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -1,16 +1,11 @@
+import { createSpy } from 'expect'
+import { noop } from 'lodash'
 /* eslint react/no-multi-comp:0 */
 import React, { Component } from 'react'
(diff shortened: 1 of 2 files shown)


erikras:
Published in [`v6.6.3`](https://github.com/erikras/redux-form/releases/tag/v6.6.3).

PR #2811, opened 2017-04-13
Update Field.md

rupertrutland:
Fix the text, it seemed like a mistake.

### DIFF
--- a/docs/api/Field.md
+++ b/docs/api/Field.md
@@ -115,7 +115,7 @@ passing them into `props`.
 #### `parse : (value, name) => parsedValue` [optional]
 
 Parses the value given from the field input component to the type that you want stored in the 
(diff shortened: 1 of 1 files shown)


docs/api/Field.md
```diff
@@ -115,7 +115,7 @@ passing them into `props`.
 #### `parse : (value, name) => parsedValue` [optional]
 
 Parses the value given from the field input component to the type that you want stored in the 
-Redux store. Common use cases are to parse currencies into `Number`s into currencies or 
+Redux store. Common use cases are to parse currencies into `Number`s or
```
gustavohenke: Why would you want to convert numbers into dates?

rupertrutland: I don't know the use case for the example but the text "parse currencies into `Number`s into currencies" does not make sense in English.

erikras: 👍

PR #2812, opened 2017-04-13
Migrate from React.PropTypes to prop-types (v5)

oziniak:
PR should eliminate annoying warning in console about migration to prop-types package https://github.com/reactjs/prop-types/blob/master/README.md#difference-from-reactproptypes-dont-call-validator-functions


oziniak:
Is there a chance this gets merged any time soon? :sweat_smile:

PR #2823, opened 2017-04-19
Migrated to v5 to prop-types

### LINKED ISSUE #2822: Use PropTypes from 'prop-types' on v5

Bug report.

PropTypes are being imported from react.  This has been fixed in the v6 branch but not on v5.

PropTypes should be imported from `prop-types` as it is in v6.

### What's your environment?

N/A

### Other information

Obviously there's a limit to how long you can continue to support the v5 branch, but hopefully this is a trivial enough change to do.  I'm happy to submit a PR against the `v5` branch if it's something that would be merged.  Unfortunately for us the migration to v6 is a massive task and we're still putting it off behind higher priority work.

erikras:
Fixes #2822.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "redux-form",
-  "version": "5.3.4",
+  "version": "5.3.5",
   "description": "A higher order component decorator for forms using Redux and React",
   "main": "./lib/index.js",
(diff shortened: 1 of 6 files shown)


erikras:
Published in [`v5.3.5`](https://github.com/erikras/redux-form/releases/tag/v5.3.5).

Issue #2832, opened 2017-04-20
passing the last initialValues into initialize action if it is reinitializing the form


sckoh:
### Are you submitting a **bug report** or a **feature request**?
**feature request**

### What is the current behavior?
initialize is not passing the last initialValues if it is reinitializing the form.

```

initIfNeeded(nextProps) {
  const { enableReinitialize } = this.props;
  if (nextProps) {
    if (
      (enableReinitialize || !nextProps.initialized) &&
      !deepEqual(this.props.initialValues, nextProps.initialValues)
    ) {
      const keepDirty = nextProps.initialized &&
        this.props.keepDirtyOnReinitialize;
      this.props.initialize(nextProps.initialValues, keepDirty);
    }
  } else if (
    this.props.initialValues && (!this.props.initialized || enableReinitialize)
  ) {
    this.props.initialize(
      this.props.initialValues,
      this.props.keepDirtyOnReinitialize,
    );
  }
};
```

### What is the expected behavior?
expect the last initialValues is passed into the initialize action creator so that we can know what what is the difference, what values are changed between the next initialValues and last initialValues.

I would like the previous initialValues to be passed into the initialize action like:

```
this.props.initialize(nextProps.initialValues, keepDirty, {
  lastInitialValues: this.props.initialValues,
});
```

### What's your environment?
redux-form: 6.6.3

erikras:
Published fix as [`v6.7.0`](https://github.com/erikras/redux-form/releases/tag/v6.7.0).

PR #2835, opened 2017-04-20
passing the last initialValues into initialize action if it is reinitializing the form

sckoh:
issue #2832


erikras:
I assume this is just for some _other_ reducer or middleware in your system to use?

sckoh:
adding this allow us to know the changes between new and old initialValues
- what trigger the reinitialise
- which field value is changed during reinitialise
- is reinitialise or fresh initialise

erikras:

PR #2836, opened 2017-04-20
Remove meta.touched prop from FieldArray

### LINKED ISSUE #2830: Remove meta.touched from FieldArrays to avoid confusion since it's always false

Bug.

`meta.touched` is included in FieldArrays even though `touched` is no longer set, which is confusing when debugging the application and you expect `touched` to behave just like it does in the Field component.

`meta.touched` should not be included since it's always `false` anyway.

### What's your environment?

Applicable since Redux Form v6.5.0.

### Other information

`meta.touched` was added to FieldArrays in PR #1715 and released in v6.0.3, but then PR #2473 and release v6.5.0 stopped updating touched on FieldArray to fix a bug.

I spent an hour trying to figure out which regression caused touched to no longer be set when in fact it was a bug fix.

erikras:
Fixes #2830.

It had never ever been set to anything, was never documented, and was only causing confusion.

### DIFF
--- a/examples/fieldArrays/src/FieldArraysForm.js
+++ b/examples/fieldArrays/src/FieldArraysForm.js
@@ -34,11 +34,11 @@ const renderHobbies = ({ fields, meta: { error } }) => (
   </ul>
 )
 
-const renderMembers = ({ fields, meta: { touched, error, submitFailed } }) => (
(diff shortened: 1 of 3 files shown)


erikras:

leojh:
Love the "never used." I was using it.

Issue #2838, opened 2017-04-20
Field validation not triggered after reset


bbenezech:
Very easy to reproduce:

https://codesandbox.io/s/o22j1lOZB

Modify age, then reset, then modify age, then submit.
Field validation on `Username` and `Email` are bypassed.
`UPDATE_SYNC_ERRORS` is not triggered after reset like it is when a validated field is dirty.

It looks like the conditions `updateSyncErrorsIfNeeded` are incorrects.

This is a major issue IMO. I have no easy workaround. 
@erikras I know you must be terribly busy but can you have a look?

irvingv8:
Hello there @bbenezech!

Where did u got that sandbox from ? 

I certainly can reproduce the behavior on it, but I can't do it on the codesanbox provided in the docs:

https://codesandbox.io/s/PNQYw1kVy

perhaps someone screwed up while setting it up ?

bbenezech:
@irvingv8 The only difference is the absence of validation on `age`.
If the modified field is validated, it works, if not, it doesn't. Sorry I should have made that clearer.

I looked at the conditionals in `updateSyncErrorsIfNeeded`, not sure I can do it, I have no clear idea what is going on here.

irvingv8:
Ok! I Believe I fixed it! I'll submit a pull request, will be my first contribution to OSS <3! hopefully I did it well, even though this was an easy one

irvingv8:
@erikras is there any guide/steps on how to contribute to redux-forms? It was little bit hard for me to get to test what i modified, I find the rebuild examples script, but it was going over all of the examples taking a LOT of time to rebuild each of them and I just wanted to test a single one. 

I ended up deleting all the other examples, running the script and testing...

Maybe the script could take the name of a example to rebuild ?

timhwang21:
I actually ran into this exact issue myself a while back. Really appreciate the PR, and I'm willing to help get this moved along!

You probably want to add a test [here](https://github.com/erikras/redux-form/blob/master/src/__tests__/reducer.reset.spec.js). Specifically, something like `it should not erase sync errors if they exist.` The test structure is pretty self-explanatory.

One question I have: I'm wondering about the order of things happening. 

1. [Validation seems to happen when new props are received](https://github.com/erikras/redux-form/blob/809c8414121ba2c3de1a5797d259ac4ebb56193e/src/reduxForm.js#L244). This function checks to see if values are deeply equal, and skips validation if so.
[...]

PR #2847, opened 2017-04-24
Added syncErrors to reset reducer

irvingv8:
Fixing issue #2838 

Validations were not triggered after reset


erikras:
I don't think this is the solution. You don't want to just persist the previous sync errors, as `RESET` has most likely changed the values, so the sync errors might be wrong.

PR #2855, opened 2017-04-26
Add missing props param in the field-level validations example documentation

kevinzwhuang:
Summary:
-----
Field level validation functions take 3 params -> `(value, allValues, props)`.
This is accurately documented in the API documentation: http://redux-form.com/6.6.3/docs/api/Field.md/#-validate-value-allvalues-props-error-optional-

However, the Field-Level Validation example only documents 2 params (http://redux-form.com/6.6.3/examples/fieldLevelValidation/).

Changes:
------
- [x] Add `props` explanation to Field-Level Validation example params documentation


gustavohenke:
Thanks!

PR #2856, opened 2017-04-26
Add active prop to reduxForm

vitbokisch:
Adding new prop passing to form via reduxForm. The value shows up when any field is active, but it is not passing to form at all. The prop, for example, helps doing more ux stuff like showing form as a modal when user is interacting withing the form and so on. Hope this proposal is worthy for you as well.

### DIFF
--- a/docs/api/Props.md
+++ b/docs/api/Props.md
@@ -22,6 +22,10 @@ class SimpleForm extends Component {
 }
 ```
 
+### `active : string`
+
+> Returns name of the active field (if any of the fields has focus), `undefined` otherwise.
+
 ### `anyTouched : boolean`
 
 > `true` if any of the fields have been marked as `touched`, `false` otherwise.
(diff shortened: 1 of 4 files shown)


vitbokisch:
You are right, I'm gonna update it

erikras:
Pro: definitely seems useful
Con: adds another reason for the entire form to rerender

When you tab from input `'foo'` to input `'bar'` form to another, does the `active` prop ever `=== undefined` in its way between the two values?

erikras:
[`v5` had this](https://redux-form.com/5.3.3/#/api/props) and it was removed as a conscious decision in the rewrite of `v6` to minimize form rerenders. I think this should be [a selector](http://redux-form.com/6.6.3/docs/api/Selectors.md/). It would be trivial to write (use #2859 as a guide). I'd accept that PR. Going to close this one.

PR #2858, opened 2017-04-27
Add a Field's initial value to props.meta.initial

bmv437:
This PR passes the initial value for a field down to the Field component as `props.meta.initial`

This is useful for having to compare the current value to the initial value to display a pending change. In my specific case, this enables checking a number never goes below the initial value, but can be increased and decreased to any value >= the initial value.


src/createFieldProps.js
```diff
@@ -88,6 +89,7 @@ const createFieldProps = ({ getIn, toJS }, name,
       dispatch,
       error,
       form,
+      initial: initial,
```
idrumgood: any reason for not just `initial`?

bmv437:
@gustavohenke Does this feature look alright? Can we get it merged? Thanks for your efforts on redux-form!

erikras:
Looks great!

erikras:

Issue #2109, opened 2016-11-10
Expose Input/Meta proptype shape


Neophy7e:
As Input and Meta props can be passed to own 'user' components it would be nice if their PropTypes could be exposed by redux-form as it's done by react-int eg: https://github.com/yahoo/react-intl/blob/master/src/types.js

niksajanjic:
You can access values and errors using selectors:

http://redux-form.com/6.2.0/docs/api/Selectors.md/

Not sure how to access other meta data per each input, not sure if that is possible, although it is stored inside redux state. I'm not contributor but I think redux-form decorator will not expose each input and meta data outside Filed component by design and I don't think they will implement that.

Neophy7e:
@niksajanjic I wasn't speaking about input meta data but about the PropTypes for examples : 

```jsx
<Field
  name="email"
  component={Input}
  placeholder="This is placeholder"
  type="email"
/>
```

```jsx
const Input = (props) => {
  const { input, meta, ...rest } = props;

  return (
      <input {...input} {...rest} />
  );
};

Input.PropTypes = {
   input: InputShape.isRequired, /* InputShape would be imported from redux-form 
 as it can be done for example with react-intl */
}
```

niksajanjic:
Aha, sorry I didn't understand you. There used to be:

    import { propTypes } from 'redux-form'; 

And then you were able to do:

   static propTypes = {
      ...propTypes,
      otherProps: PropTypes.object.isRequired,
    };

I don't use this anymore in v6 as it breaks my Eslint rules but I have a feeling this can't be used inside Field component, only outside, so I think you're right and this is an issue. I'm not a contributor though, so I can't help you any further.

ghostd:
Here is a proposal #2862

gustavohenke:
Because this issue saw almost no activity for a few months, I'm closing it.

PR #2862, opened 2017-04-28
Add PropTypes for the Field components

ghostd:
Proposal for #2109


erikras:
Missing:
- Add to `src/createAll.js`
- Add to `src/index.js`
- Add to `src/immutable.js`

You've got to actually export it from the library. Look at how `propTypes` is done in each of these files.

ghostd:
Thx for the hint. I updated the PR

Issue #2857, opened 2017-04-27
Checkbox value breaking pristine state


vitbokisch:
Hi,
I found out that using true/false value within checkboxes is breaking the pristine state value. The case is: Checkbox initial value is probably an empty string. When I set a checkbox to checked state and then uncheck it, the pristine value changes. The flow is '' -> true -> false. You can even check it in the simple example here: http://redux-form.com/6.6.3/examples/simple/ 

I fixed it by using normalize prop in Field for checkbox component like: `<Field normalize={value => value || ''} ... />`. Another approach could be setting default initial values for all the checkboxes in each form, which doesn't make sense at all.

I propose to set as a default behavior true || '' for checkboxes. Then the checkboxes will behave the same way as any other types of inputs.

erikras:
Published fix as [`v6.7.0`](https://github.com/erikras/redux-form/releases/tag/v6.7.0).

dmeenhuis:
Not sure if intended, but this change now breaks the application I'm currently working on. I'm submitting boolean values from my form to an API which expects the relevant property to be an actual boolean value (``true``/``false``).

However, after this change, when I uncheck the box, it now submits an empty string while previously it submitted ``false``. The API cannot convert the empty string to a boolean, so I'm getting an error.

vitbokisch:
@dmeenhuis if it worked before and you use Field component, you can just use the opposite approach I've been using (I mean using normalize). Could work like this:
```
<Field 
    component={Control} 
    name={name} 
    normalize={value => value || false}
    {...props} 
  /> 
```
Also using props `parse` or `format` may help.

dmeenhuis:
Sure, there are plenty of workarounds to think of. I'm just not sure if the current default behaviour introduced by this change is really sensible.

It doesn't feel right that something that worked like it should before (checkbox checked means a ``true`` value, unchecked is a ``false`` value), now requires additional code to make it work logically again. It's also pretty easy to forget, causing any receiving system that expects an actual boolean value to break. Perhaps @erikras would like to chime in?

Is there any way we can get the change introduced by this PR to work together with the previous behaviour?

potty:
Shouldn't be this change marked as BC break? Suddenly my app stopped working after update.

geminiyellow:
This change is actually a fuck BC

theKirt:
The solution posted by @vitbokisch works 👍

kgram:
[...]

PR #2863, opened 2017-04-28
Checkbox default behavior proposal

vitbokisch:
instead true/false to be true/'' as the current true/false breaks pristine prop - problem described here: https://github.com/erikras/redux-form/issues/2857


erikras:
This makes perfect sense. Thanks! 👍

erikras:

Issue #2875, opened 2017-05-03
"Invalid Keypath" error when editing FieldArray after submit has failed.


dagstuan:
### Are you submitting a **bug report** or a **feature request**?
Bug report

### What is the current behavior?
https://codesandbox.io/s/MPzRZV21

If you submit the form, and then (after the submit has failed) try to add/delete members and edit their fields. An "Invalid KeyPath" error will be logged to the console.  This is only while using the immutable.js-version of redux-form.

### What is the expected behavior?
No "invalid keypath"-message

### Other information
I have narrowed this problem down to the code for `ArraySplice` which inserts `undefined` into the state for `submitErrors` and `asyncErrors`. When editing one of the fields inside a FieldArray the code `deleteInWithCleanUp(result, `submitErrors.${field}`)` is called, which tries to update the submit errors for the field inside the array, however, the array element is set to "undefined" in the state, and a call to immutable.js such as `state.updateIn(['form', 'formName', 'submitErrors', '0', 'fieldName'], /*  rest */)` fails with the "Invalid KeyPath" error.

erikras:
Published fix as [`v6.7.0`](https://github.com/erikras/redux-form/releases/tag/v6.7.0).

PR #2876, opened 2017-05-03
deleteInWithCleanUp needs to check if there is a value in the state before deleting.

dagstuan:
This is related to #2875. I'm not sure if this is the best way of fixing the issue, but it solves the problem atleast. Adds a check to see if there is anything in the state before calling `deleteIn` when `deleteInWithCleanUp` is called. With this code I no longer get `Invalid KeyPath` errors when using Immutable.js and editing FieldArrays after submit has failed.

### DIFF
--- a/src/deleteInWithCleanUp.js
+++ b/src/deleteInWithCleanUp.js
@@ -9,7 +9,12 @@ const createDeleteInWithCleanUp = ({ deepEqual, empty, getIn, deleteIn, setIn })
       const parent = getIn(state, pathTokens.join('.'))
       return parent ? setIn(state, path, undefined) : state
     }
-    const result = deleteIn(state, path)
+
+    let result = state
+    if (typeof getIn(state, path) !== 'undefined') {
+      result = deleteIn(state, path)
+    }
+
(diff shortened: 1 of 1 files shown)


src/deleteInWithCleanUp.js
```diff
@@ -9,7 +9,12 @@ const createDeleteInWithCleanUp = ({ deepEqual, empty, getIn, deleteIn, setIn })
       const parent = getIn(state, pathTokens.join('.'))
       return parent ? setIn(state, path, undefined) : state
     }
-    const result = deleteIn(state, path)
+
+    let result = state
+    if (typeof getIn(state, path) !== 'undefined') {
```
gustavohenke: `typeof myvar != "undefined"` is usually only useful for checking if global vars exist - for vars you know exist, use simply `myvar !== undefined`.

Not a big deal here, though, so no need to bother changing it!

erikras: Not a big deal, but I prefer avoiding magic strings whenever possible.

erikras:

PR #2888, opened 2017-05-04
Fixed field-level validation reset bug

### LINKED ISSUE #2838: Field validation not triggered after reset

erikras:
Fixes #2838.

The `this.lastSyncErrors` logic added in #2797 was flawed.


erikras:

PR #2889, opened 2017-05-04
Better cleanup on UNREGISTER

### LINKED ISSUE #2884: syncErrors still persist in state after fields have been unmounted and unregistered
I have required fields that only get mounted when a checkbox field is true.
When the checkbox is checked, `syncErrors` are now present in state (makes sense)
When the checkbox is then unchecked, `syncErrors` are still there even though the fields are not in `registeredFields`

When the fields are unmounted and unregistered, they should also be removed from `syncErrors` because they stop the form from being submitted successfully.

### What's your environment?

v6.6.3
Chrome latest
Mac OS X 10.11.4
Node v7.2.0

If a code example is required please let me know. I'm working in a private codebase so I'd have to something up.

Thank you

erikras:
Cleaned up the following on field unregister:

- sync errors
- sync warnings
- submit errors
- async errors

Fixes #2884.

### DIFF
--- a/src/__tests__/reducer.unregisterField.spec.js
+++ b/src/__tests__/reducer.unregisterField.spec.js
@@ -13,6 +13,66 @@ const describeUnregisterField = (reducer, expect, { fromJS }) => () => {
       })
   })
 
+  it('should remove sync errors', () => {
+    const state = reducer(fromJS({
+      foo: {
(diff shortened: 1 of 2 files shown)


ArjanJ:
Thank you!

erikras:

PR #2891, opened 2017-05-05
Exports defaultShouldValidate and defaultShouldAsyncValidate

### LINKED ISSUE #2890: Export default shouldValidate for external use
Feature request (happy to implement myself as my first PR to the project!

Recently, I was working on a project where I have a form that can exist within several states. This is controlled by a property that is passed to the component which renders the redux-form.

We use this prop to control the validation that's performed (in one state, we allow for certain fields to be left blank, in another we don't). However when this prop changed, I noticed that my validation function wasn't being called.

I fixed this by providing a custom implementation of shouldValidate - just checking if the current prop !== next prop. However, as the default shouldValidate function isn't exposed externally, I couldn't just simply provide a function that wrapped the default behaviour.

It would be nice if the default shouldValidate (and by extension the default async shouldValidate) were exported externally, to reduce duplication.

sh1989:
These default functions are exported for external use - particularly so
that a user can wrap the default functionality with specialised behaviour.

To resolve #2890


erikras:
Can you add them to `src/immutable.js`, too?

erikras:

Issue #2882, opened 2017-05-03
Immutable deepEqual function returning wrong value when comparing Map's with different structure


Inucoder:
### Are you submitting a **bug report** or a **feature request**?

bug report

### What is the current behavior?

https://github.com/erikras/redux-form/blob/master/src/structure/immutable/deepEqual.js function returns true for Immutable.Map's where the first Map has a key that is not present in the second Map and it's value is null, '' or false.

For example
```javascript
const A = fromJS({a: ''});
const B = fromJS({b: 'Some String'});
deepEqual(A, B); //returns true
```

### What is the expected behavior?
Above snippet should return false because the Map's are not deeply equal.

### What's your environment?

Redux Form 6.63
Mac OSX
Node 7.5.0

### Other information

I think this piece of code from https://github.com/erikras/redux-form/blob/master/src/structure/immutable/deepEqual.js
```javascript
if (Iterable.isIterable(obj) && Iterable.isIterable(other)) {
  return obj.count() === other.count() && obj.every((value, key) => {
    return isEqualWith(value, other.get(key), customizer)
  })
}
```
should validate than 'other' object have 'key' before the recursive call, like this
```javascript
if (Iterable.isIterable(obj) && Iterable.isIterable(other)) {
  return obj.count() === other.count() && obj.every((value, key) => {
    return other.has(key) && isEqualWith(value, other.get(key), customizer)
  })
}
```

erikras:
Published fix as [`v6.7.0`](https://github.com/erikras/redux-form/releases/tag/v6.7.0).

Inucoder:
Awesome! Thanks @erikras

PR #2892, opened 2017-05-06
Fix deep equal on maps with different structures

tyleriguchi:
Fixes issue #2882. Deep equal wasn't checking to see if the key exists on the map it was comparing itself to.


erikras:
👍

erikras:

Issue #2315, opened 2016-12-20
How to reduce lib size?


0rangeT1ger:
I'm using redux-form and my app.xxx.js is 90kb bigger（uglified, no gzip）.
Any thoughts on how to reduce the lib size？

gustavohenke:
Usually, libraries offer a "function per file" approach, so you're only packing what you really use.

redux-form even does separate each component, selector and etc separated in its own file, but the default export of those files is a factory function. 
So you can't really use such approach to decrease your JS bundle size :(

maslianok:
Yeah, I'm also found that redux-form the biggest library in my bundle

<img width="539" alt="screen shot 2016-12-25 at 10 48 17 am" src="https://cloud.githubusercontent.com/assets/3485490/21469635/d248b61e-ca8f-11e6-897e-c9f7f8e45f49.png">

I

gustavohenke:
webpack 2 does in fact increase the size of your bundle.
See webpack/webpack#2451 for some reference.

0rangeT1ger:
Used System.import() to split out the redux-form part.
That part is 93KB.
It seems huge....

erikras:
Some background reading: #890 #859 #1481 

A big chunk comes from `lodash`. I have tried to implement some of the `setIn()` features from `lodash` in the past, and the truth is that it's just a complicated task that takes a lot of code. However, this lib is using `lodash` in a way that _should_ share well if you are using `lodash` in your project that is also using `redux-form`.

estaub:
@erikras I believe you'd get a big win from cherry-picking lodash in your imports, e.g., replacing 

`import { isEqualWith } from 'lodash'`

with

`import isEqualWith from 'lodash/isEqualWith'`

I just went through all my dependencies using lodash to find what's dragging in all of lodash, and all the others are using this idiom.

To reduce the editing, make a `myLodash.js` with a line like

`export isEqualWith from 'lodash/isEqualWith'`

for each function you use anywhere.  Then just retarget your lodash imports at that file.

mizchi:
@erikras I used webpack-bundle-analyzer to measure my project and found lodash and lodash-es include 260kb and those with redux-form itself becomes 430kb. It's big somewhat as one dependencies. For your reference.

https://github.com/th0r/webpack-bundle-analyzer

gustavohenke:
> However, this lib is using lodash in a way that should share well if you are using lodash in your project that is also using redux-form.

[...]

PR #2893, opened 2017-05-06
Allow direct imports to reduce bundle size.

ruiaraujo:
Webpack 2 does not remove a lot of unused modules from my bundle so direct import remain the best solution to reduce bundle size.

The doc page was heavily inspired by https://github.com/ReactTraining/react-router/blob/v3/docs/guides/MinimizingBundleSize.md


gustavohenke:
Can you check #2315, please?
Over there guys think the bundle size problem is related to lodash.

BTW, what a huge PR!

ruiaraujo:
For me since I was also using lodash, it is shared so that is not a problem. The problem was the factory index files that prevented direct imports.

erikras:
Wow! That's an impressive PR!

Well done. Thank you for your time.

docs/faq/BundleSize.md
```diff
@@ -0,0 +1,61 @@
...
+```js
+import reduxForm from 'react-router/es/reduxForm'
+import Field from 'react-router/es/Field'
+import FieldArray from 'react-router/es/FieldArray'
+```
+
+
+Use `es` if you are using a bundler that can process ES modules like `webpack@2` or `rollup`, otherwise use `lib`.
+
+The public API available in this manner is defined as the set of imports available from the top-level `react-router` module.
```
erikras: Oops. Missed a `react-router`. 😛

ruiaraujo: god dammit. Do you fix it or should I?

ruiaraujo: Next time I 'inspire' myself on the work of others, do cmd+R. 10 year old me would not have done this kind of mistake. 😃

ruiaraujo: Fixed in #2898

erikras:
@ruiaraujo Is there some advantage to doing

```jsx
import { default as Foo } from './somedir/Foo'
```
over
```jsx
import Foo from './somedir/Foo'
```
??

The latter is less explicit, but much more readable, in my opinion.

erikras:
Oops, nevermind. They are *EXPORT* statements. There, now we've both made absentminded mistakes on this PR. :-)

erikras:
Ugh. ☹️ In retrospect, you really should have `git mv`'d `reduxForm.js` to `createReduxForm.js`, and then created a new `reduxForm.js`, as this PR has wiped all annotation history from the file, since all the complex code now exists in a brand new file. Oh well... live and learn...

(I probably would not have thought to do this either.  I'm just only realizing this in retrospect.)

ruiaraujo:
Hmm, does git mv actually help with that problem?

git mv is just a mv + git add which was what I did.

erikras:
Hmm... [So it is](http://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv). I was under the impression that it would preserve the history of the file. I swear I've seen it do that before. Perhaps it only works if you don't immediately add another file with the old name...? Oh well.

Nevermind then. 😄

gajus:
Is there a timeline for this getting released. :-) ?

docs/faq/BundleSize.md
```diff
@@ -0,0 +1,61 @@
...
+                "redux-form": {
+                  "transform": "redux-form/es/${member}",
+                  "preventFullImport": true
+                }
+            }]
+        ]
+    }
+```
+
+##Caveat (Action Creators)
```
gajus: From the documentation, it is not clear whether this step is automated by `babel-plugin-transform-imports` or do I still need to use `react-router/es/actions` when using the plugin.

ruiaraujo: I will try to be more explicit about the problem​ then.

erikras: LOL, even the follow-up question didn't notice the wrong library name. 😆

erikras:

piuccio:
screen shot 2017-05-11 at 16 23 56

Interesting PR, two minor things
* You need a space in the markdown after `##`, currently it looks like in the picture above
* There's no mention on how to use this together with `immutable.js`

ruiaraujo:
@piuccio It is the same thing to use it with immutable, just add immutable.

piuccio:
> just add immutable

where?

`redux-form/es/whatever/immutable`
`redux-form/es/immutable/whatever`

I suppose it's the first one, but it would be nice to have it in the doc

timhwang21:
Hi, I'm noticing that the Babel plugin doesn't work well with actions:

```
WARNING in ./src/foo/bar.js
23:23-30 "export 'default' (imported as 'actions') was not found in 'redux-form/es/actions'
```
My import:

```javascript
import { actions, Field, reduxForm } from 'redux-form';
const { reset } = actions;
```

This can be worked around from the user's end by changing the import statement, but I was wondering if it would be a good idea to add a default export to `es/actions` that contains all the actions. I think the import pattern I'm using is fairly standard, and the change would help users optimize their bundles with minimal refactoring.

Alternatively, if this is happening because I'm doing something wrong, please let me know. Thanks!

brneto:
@erikras What is that createValues.js and values.js? Why it's not documented? Can I use it do access the form field values or you plan to remove that on the future?

PR #2904, opened 2017-05-09
Added initialized prop

### LINKED ISSUE #2901: Calling initialize(values) does not set props.initialized to true
### **Bug report**
version: 6.6.0

Calling initialize(values) does not set props.initialized to true

this.props.initialized === true

erikras:
Fixes #2901.

Although this does add an additional form rerender. It will only ever happen once during the form's lifecycle.


erikras:

PR #2912, opened 2017-05-10
Added webpack-bundle-analyzer script

erikras:
...also yarn.lock.


package.json
```diff
@@ -10,9 +10,10 @@
     "url": "https://github.com/erikras/redux-form"
   },
   "scripts": {
+    "analyze": "webpack src/index.js dist/redux-form.js -p --bail --profile --json > stats.json && webpack-bundle-analyzer stats.json",
```
gustavohenke: When I want to analyze the webpack results somehow, I usually do it like `webpack --profile --json | webpack-bundle-analyzer`.

Just a minor idea, in case you like it 😃

erikras: The current versions of these tools seem to require having an actual `stats.json` on the filesystem.

PR #2914, opened 2017-05-11
Added webpack-bundle-analyzer script

phaniacube:


erikras:
I'm confused. Isn't this the same as #2912?

Issue #2118, opened 2016-11-12
Bulk change event support


melinath:
We have a medium-sized form - ~60 fields - with values that _can_ be auto-generated from the values in another part of the form. So there's a button, and if you click it, it calculates the values and fills them in for you.

This ends up being slow, because it means we have to fire off ~60 separate change events, which each has to be handled individually. This is _helped_ by turning off dev tools, but is still slow enough to be inconvenient during development and feel sluggish in production.

Related: #2023

erikras:
Several times, I have considered including some action batching code, but then I remember that there are libraries like [`redux-batched-actions`](https://www.npmjs.com/package/redux-batched-actions) that do that for you. Could something like that help you?

brentmulligan:
I am seeing a similar issue on a payment form with auto-completion. After entering an email address Chrome autofills the shipping address, etc. The auto fills causes redux-form to fire off a FOCUS, BLUR and CHANGE for each field. When you combine this with async validations where all fields are validated for each BLUR it gets hairy fast. All that said, I have not run this on a production build yet so it may be less noticeable when I get there.

I will take a look at redux-batched-actions and see if that might be a good fix.

brentmulligan:
@erikras @melinath redux-batched-actions does seem to work pretty good if you also add a small piece of middleware to catch/debounce the redux-form actions you'd like to tame. ([middleware example](https://gist.github.com/brentmulligan/21922996bd9a46dfd611a3dd314b6942)).

christopherscott:
+1, this would be nice/helpful. We're currenly using something like this:

```
export function changeBulk (form, obj = {}) {
  return (dispatch) => {
    const fields = Object.keys(obj)
    fields.map(field => {
      dispatch(change(form, field, obj[field]))
    })
  }
}
```

Industrial:
@erikras I think that dispatching `REGISTER_FIELD` at `O(n)` is a terrible idea. Why not make it `O(1)` by adding an event `REGISTER_FIELDS` and only dispatch it once?

My form has 200 inputs. That's 200x `REGISTER_FIELD` dispatched to my store. 200x React render and THEN I see my page. That's 6000 milliseconds React render time right there. Compared to my 15 milliseconds HTML load time it's terrible.

erikras:
@Industrial I'll be curious to see your PR.

Industrial:
@erikras I hadn't looked at the source of redux-form at the time, but I will make an attempt ^_^'

[...]

PR #2928, opened 2017-05-14
Attempt to add a new action to batch registration: REGISTER_FIELDS

Industrial:
Hi.

I have a big form with 206 inputs (https://github.com/erikras/redux-form/issues/2118#issuecomment-300196560). Currently it takes a lot of time to render the page. It's even worse (6 to 10 seconds) if you have the devtools open. This is largely due to the fields registering one by one.

This is my attempt at adding an action to batch the registering of actions on the initial load.

I am trying to move the registering of the fields from the `Field` component to the `Form` component by collecting all the fields at `Field#componentWillMount` and doing the register in `Form#componentDidMount`.

The page load is now 1.2 seconds for me, down from 7 with devtools performance analyzer.

All the tests are passing except for the FieldArray ones. I might need some help with this. Would you be interested in adding this functionality @erikras ?

gr,

Tom


src/createField.js
```diff
@@ -29,6 +29,22 @@ const createField = ({deepEqual, getIn, setIn, toJS}) => {
...
+          'Field',
+          () => this.props.validate,
+          () => this.props.warn
+        )
+        return
+      }
+
+      if (registeredFields[this.name]) {
+        this.context._reduxForm.unregister(this.name)
+      }
```
erikras: Why are we unregistering on `cWM`? Because it will be reregistered on `cDM`?

Industrial: I'll remove it, since it has no function I think.

Industrial:
@erikras: There's still some tests failing. I'm not sure how to fix FieldArray to make them pass.

HankMcCoy:
FYI, I am working on a very similar PR here: https://github.com/HankMcCoy/redux-form/pull/1/commits/ad50754, which I hope to have in a reviewable state by the end of the week. Just wanted to double check that the idea of adding batched registrations/unregistrations as an opt-in is something that you'd be potentially be interested in merging @erikras?

erikras:
Sorry for the delay. Yes, I would be interested in merging such functionality, @Industrial, @HankMcCoy.

tyscorp:
Does this PR work in React 16 also?

erikras:
Sadly, due mostly to neglect on my side, this PR is fairly out of date. It might be quicker to resubmit it on the current `master`.

Industrial:
I have not used redux-forms (nor had a chance to) in quite some time.

I made the PR to explore the possibility of this functionality but I don't
think I will be developing it further.

I hope Hank McCoy will have an attempt

HankMcCoy:
I'm hoping to get https://github.com/HankMcCoy/redux-form/commit/ad50754 polished into an actual PR, based off of 7.1.0 this weekend.

bertho-zero:
Having a 'CHANGE_FIELDS' action would also be useful for components that manage multiple fields at once.

erikras:
I'm closing this due to git-rot, but I'd be open to the same thing being implemented fresh against the latest codebase.

PR #2939, opened 2017-05-17
Submit - Avoid strack trace using native submit

renatoagds:
Using the new `Form` wrapper, I catch a problem that if you're using the wrapper but you want to use the native submit passing `onSubmit` into config param, will enter in a stack trace.

Here a code example:

```
import {Form} from 'redux-form';

class MyComponent extends React.Component {
  render() {
    const {handleSubmit} = this.props;
    return (
       <Form onSubmit={handleSubmit}>
         <input type="submit" value="aaa" />
       </Form>
    );
  }
}

reduxForm({
  form: 'test',
  onSubmit: (value) => value
})(MyComponent);
```

I just added a verification to not call `innerSubmit` if is already the `submit` function.


erikras:

PR #2954, opened 2017-05-23
Added validation for min length, alphanumeric and phone number.

LagunaISW:


erikras:
What value does this add?

LagunaISW:
@erikras I just used those on my app, on fields with "required" anyone can send just a "space" if there is no min value, also usually Username doesn't allow especial characters so, I just add those validation examples.

PR #2964, opened 2017-05-25
Add safety checks for events in ConnectedField

hook:
When you pass a value to a Field's Component's onChange, ConnectedField passes that raw value as the event.

Using the Field's `onChange` or other handler overrides lets us `event.preventDefault()`, but if the event is a value, we get an error.

This does a safety check before calling `event.preventDefault()`.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -8,6 +8,24 @@ import plain from './structure/plain'
 
 const propsToNotUpdateFor = ['_reduxForm']
 
+const isObject = (entity) => entity && typeof entity === 'object'
+const isFunction = (entity) => entity && typeof entity === 'function'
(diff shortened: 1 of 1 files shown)


erikras:
@hook, if you had actually run the tests, you'd have spotted this simple bug I just fixed. https://github.com/erikras/redux-form/pull/2964/commits/495444c40bc4ce6dd5d1d5e0ee28bf9af6c7ae47

erikras:
It's always nice when a unit test actually finds a bug. :-)

hook:
My bad on missing the bug, I sent the PR from work and then forgot to revisit it later (I saw the test failed but the firewall at work was blocking Travis).

Excellent library by the way!

erikras:

PR #2984, opened 2017-05-30
updated PropTypes for initialValues to support an array or an object

erikrestificar:
initialValues can be an array or an object.  For example a form with fields: 
`<Field name="0.firstName />
<Field name="1.firstName" />`


erikras:
That's funny. It never once occurred to me that someone might do that. Sure, why not! 😄

erikras:

Mario-Eis:
What to pass to ArrayField`s name attribute when using arrays as initialValues? An empty name string doesn't seem to work.

Issue #2852, opened 2017-04-26
FormSection and nested Field component bug


Kestutis:
### Are you submitting a **bug report** or a **feature request**?
Bug with FormSection and nested Field

### What is the current behavior?
When using FormSection with nested Field components, input names get prefix with every nested reuse. 
```javascript 

const fieldComponent = ({input}) => {
	console.log(input.name); // as expected form1.foo
        return <Field {...input} component={nextFieldComponent} />
}

const nextFieldComponent = ({input}) => {
	console.log(input.name); // becomes form1.foo.foo
	return <input {...input} />
}

const Form = () => (
  <div>
    <FormSection name='form1'>
      <Field name='foo' component={fieldComponent} />
    </FormSection>    
  </div>
)
```
fiddle with alert on this: https://jsfiddle.net/86v1yuuz/2/

### What is the expected behavior?
It should not append FormSection name on nested Field components

### Sandbox Link
https://jsfiddle.net/86v1yuuz/2/

### What's your environment?
Redux v. 6.6.3, React 15.5.3

cyberpluto:
We are seeing the same issue with `<FormSection>` nested inside `<FormSection>`

wmertens:
This causes problems when you have e.g.

```js
const Section = ({input: {name, value}}) => (
  <FormSection name={name}>
      <Field Component={Toggle} name="enableSub"/>
      {value.enableSub && <Field Component={Section} name="sub"/>}
  </FormSection>
)
```

Nesting like this will work ok for the first Section (say it has the name "root"), but every subSection will add the full path to the `sectionPrefix`, because `Field` will pass the full `"root.sub"` name to Section instead of the given `"sub"`.

It seems to me that either:
* Field components should always get their `name` relative to `sectionPrefix`
* or FormSection should accept an `absolute` parameter so it can replace `sectionPrefix` entirely, e.g. `<FormSection absolute name={name}>` in the example

The former is more intuitive IMHO but is an API change that could cause issues. The latter is a simpler solution, and can be used in other situations as well.

wmertens:
Another option would be to have to add a `.` in front of the name to make the `FormSection` relative.

```js
<FormSection name="root">
  <FormSection name=".sub"/> // sectionPrefix becomes root.sub
  <FormSection name="otherRoot"/> // sectionPrefix becomes otherRoot
</FormSection>
```

erikras:
[...]

PR #2985, opened 2017-05-30
FormSection: add 'absolute' parameter

wmertens:
This is a possible solution for #2852


erikras:
Closing this after discussion on #2852 lead to #2996.

PR #2994, opened 2017-05-31
Correct grammatical error in API README.md

michfarr:
Fixing an extraneous _to_ that was squirreled away under the descriptive text for `Fields`.


erikras:
Good eye for _to_ squirreling!  :shipit:

PR #3000, opened 2017-06-01
Export formPropTypes as an alias for propTypes

Empact:
These propTypes are those for the form component and labelling them as such
disambiguates the variable for all importers, saving them from having to
redefine the name 'as' something in the base case.

### DIFF
--- a/src/__tests__/immutable.spec.js
+++ b/src/__tests__/immutable.spec.js
@@ -43,6 +43,7 @@ import {
   hasSubmitSucceeded,
   hasSubmitFailed,
   propTypes,
+  formPropTypes,
   reducer,
   reduxForm,
   registerField,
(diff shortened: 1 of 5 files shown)


erikras:
This adds extra API surface area (which needs to be maintained forever) for not a whole lot of reason, it seems to me. Can you explain why `import { propTypes } from 'redux-form'` is so cumbersome?

Empact:
Because propTypes is IMO an overly general name so in practice I rename it:

import { propTypes as formPropTypes } from 'redux-form'

This also make more sense now that the fieldPropTypes are exported - the name more explicitly distinguishes the propTypes from any others.

You could optionally deprecate and remove the propTypes export to remove the extra maintenance burden.

erikras:

PR #3001, opened 2017-06-01
Fix fieldPropTypes to have shapes for input and meta

Empact:
Otherwise we get errors on use, e.g.:
```
  Failed prop type: inputField: prop type `meta` is invalid;
  it must be a function, usually from React.PropTypes.
```


erikras:

PR #3002, opened 2017-06-01
Using initial values if values haven't been initialized yet

Darijusch:
I changed for the connectedFields how the value is retrieved, test run through and according to my testing it didn't break anything and fixes the first rendering, so server side rendering is working now (which didn't work before)

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -248,7 +248,8 @@ const createConnectedField = ({deepEqual, getIn, toJS}) => {
       const initial = initialState !== undefined
         ? initialState
         : initialValues && getIn(initialValues, name)
-      const value = getIn(formState, `values.${name}`)
+      const values = getIn(formState, 'values') || getIn(formState, `initial`) || initialValues
+      const value = getIn(values, name)
       const submitting = getIn(formState, 'submitting')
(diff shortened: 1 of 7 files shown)


Darijusch:
hi @gustavohenke,
whats the best approach for testing serverside rendering inside of the unit test, since this issue only happens on server side rendering?

Darijusch:
so i used for now real server side rendering since i couldn't reproduce it with the client side rendering, let me know if there is something unclear or if i should change anything

gustavohenke:
Just immitate the behavior of the server 😉

gustavohenke:
> so the issue with server rendering ist that actions do not modify anything in redux store

Hmm, this doesn't feel right. There are probably some problems in your setup, since Redux is environment-agnostic.
Perhaps you should create a repo that reproduces this problem and then post a issue detailing what happens?

Darijusch:
the thing is you can only define an initial state, you can dispatch actions but updates in the store  would result in a rerendering of the components, which you don't have on server side rendering. If you have any other suggestion how to test this behavior.

Darijusch:
ok i tested this weekend from home with a sample setup and we had an issue that server side the reducer was missing :) so i close the pull request :)

PR #3006, opened 2017-06-02
updated getting started docs - diagram and descriptions

fjaskolski:
Hi there!

I've happened to use `redux-form` and I really dig it! As a little give back, I'd like to help with docs and submit a few proposals. Feel free to totally disagree, those are just suggestions!

**Summary:**
1. Removed information on Action Creators from the spotlight, as it's not crucial to get started,
2. Switched class components into functional ones within examples for better readability,
3. Removed `decorator` terminology, as it may be confusing; AFAIK there is no actual ES7 decorator within the codebase, so I'd stick with HOC – what do you think?
4. Also the `<Field/>` is used in the similar fashion, so probably could be called HOC as well,
5. Created a diagram of the basic data flow.

This is just a quick draft. Let me know if you'd like me to go on!


docs/GettingStarted.md
```diff
@@ -1,104 +1,160 @@
...
-3. A `Field` component to connect your individual field inputs to the Redux store.
-4. Various Action Creators for interacting with your forms throughout the application.
+## Overview
 
-## Implementation Guide
+To connect your React form components to your Redux store you'll need following pieces:
 
-### Step #1
+* Redux Reducer: `formReducer`,
+* React HOCs: `reduxForm()` and `<Field/>`.
```
erikras: Hmm.... I never considered `Field` to be a higher order component, but I suppose it sort of is....

docs/GettingStarted.md
```diff
@@ -1,104 +1,160 @@
...
+
+Take a look at the simplified diagram below. Note that in most cases you don't need to worry about the [action creators](http://redux-form.com/6.7.0/docs/api/ActionCreators.md/) for yourself, as they're already bound to dispatch for certain actions.
+
+<div style="text-align: center;">
+  <img src="https://cloud.githubusercontent.com/assets/5114422/26726540/c2f60e8c-47a3-11e7-8dd4-13ebd00d656a.png" width="500" style="max-width: 100%;"/>
+</div>
+
+The flow is pretty simple. Let's go through a simple example:
+* User clicks on the text input wrapped with `<Field/>` component,
+* Form component wrapped with `reduxForm()` dispatches an action,
```
erikras: In my mind, it's the `Field` component that dispatches the action.

erikras:
Overall, this is excellent! I ❤️❤️❤️ the flow chart!

fjaskolski:
Hi there guys,

Thanks for the feedback, glad you like it! 🎉 

Just pushed some updates:
- removed "HOC" wording around `<Field/>`, totally agree that it's stretching it too much; much better to call it by name (which would be the [class] component),
- URL to the diagram follows the URL pattern used at [value lifecycle](http://redux-form.com/6.7.0/docs/ValueLifecycle.md/)
- minor wording tweaks and a grammar check.

Cheers!

erikras:

Issue #3008, opened 2017-06-02
FieldArray not updates on own field change


vtaits:
### Are you submitting a **bug report** or a **feature request**?
Bug

### What is the current behavior?
https://jsfiddle.net/vadim91/nduqzsst/1/

### What is the expected behavior?
In this example "Remove" button should disable other inputs and "Restore" button should enable them. But this doesn't happen after each button click. 

### What's your environment?
Ubuntu 16.04.2 LTS
node 6.10.1
npm 3.10.10

"react": "^15.5.4"
"react-dom": "^15.5.4"
"react-redux": "^5.0.5"
"redux": "^3.6.0"
"redux-form": "6.7.0"

### Other information
Looks like a bug in shouldComponentUpdate method of ConnectedFieldArray
```js
    shouldComponentUpdate(nextProps) {
      // Update if the elements of the value array was updated.
      const thisValue = this.props.value
      const nextValue = nextProps.value

      if (thisValue && nextValue) {
        if (
          thisValue.length !== nextValue.length ||
          thisValue.every(val => nextValue.some(next => deepEqual(val, next))) // <--------------
        ) {
          return true
        }
      }
```

Now component updates if each prev value equals each next value and not otherwise.

erikras:
This is not a bug; it's a feature! 😀 The `FieldArray` component (`renderMembers` in this example) does not rerender every time one of the values inside it changes. That would be horribly inefficient for large arrays.

That said, the line that you highlighted makes no sense to me; I don't recall writing it, and all the tests pass when it is removed. Why on earth we would want to rerender the `FieldArray` when all the values are equal? It certainly seems misplaced.

As for your particular use case, I don't have a great answer. Perhaps some sort of prop like `rerenderOnEveryChange` to `FieldArray` might work....?

erikras:
Fix published in [`v6.8.0`](https://github.com/erikras/redux-form/releases/tag/v6.8.0).

vtaits:
Thank you!

vtaits:
I caught a bug.
https://jsfiddle.net/nduqzsst/3/
If there are multiple fields groups with same values, "remove" works incorrectly.

A little later I'll fix.

nihaux:
yep same problem happen when you change the order of the FieldArray.  
In my use case, I have an array of tags in which the order matters.   
If you switch tag 1 with tag 3 shouldComponentUpdateReturns false.  
[...]

PR #3009, opened 2017-06-02
Added rerenderOnEveryChange

vtaits:
#3008 

1. Added `rerenderOnEveryChange` prop to `FieldArray`
2. Added test `should rerender when a value changes if rerenderOnEveryChange setted`
3. Added docs for `rerenderOnEveryChange`


erikras:

Issue #2996, opened 2017-05-31
Feature Proposal: formValues() HOC decorator that works inside a form component


erikras:
From discussion on #2852 with @wmertens. The problem being solved is that [`formValueSelector()`](http://redux-form.com/6.7.0/docs/api/FormValueSelector.md/) cannot be used in a reusable way inside a [`<FormSection>`](http://redux-form.com/6.7.0/docs/api/FormSection.md/) because it doesn't know what the prefix defined in the `<FormSection>` is. General API syntax would look like:

```jsx
@formValues({
  destProp: 'nameOfField'
})
class MyReusableFields extends Component {
  render() {
    const { destProp } = this.props
    // destProp is the value of the field called 'nameOfField'
    return (
      <div>
        <Field name="nameOfField" component="input"/>
        ... other fields

        ... some logic involving value of nameOfField:
        {destProp === 'specialValue' && <div>Value is SPECIAL!!</div>}
      </div>
    )
  }
}
```
There might also be an alternative syntax where, if your keys and values were all going to be the same, rather than pass...
```jsx
@formValues({
  firstField: 'firstField',    // key and value are ===
  secondField: 'secondField'   // key and value are ===
})
class MyReusableFields extends Component {
  ...
}
```
...you could just pass a list of strings. e.g.
```jsx
@formValues('firstField', 'secondField')
class MyReusableFields extends Component {
  ...
}
```

wmertens:
Alright, I took a stab at implementing it. It's using `@connect` to listen to store updates, but it uses the `getValues()` from `context._reduxForm` to get at the values. See #3011

erikras:
Merged #3011. Will comment again when released.

erikras:
Fix published in [`v6.8.0`](https://github.com/erikras/redux-form/releases/tag/v6.8.0).

jasan-s:
I am sure I'm using the feature incorrectly. Instead of  reading from Redux directly I thought I could use `Values = formValues('email','password')(Auth)`  the` Auth` is the form name and the Field names are `email `and `password`.

jasan-s:
I wanted to access forms values within the component.  I know I can pass it down by using connect but though this may be cleaner.

wmertens:
No, what you did there is create a component named Values that gives
`email` and `password` as props to the Auth component

jasan-s:
@wmertens  so, what is the recommended way to accesses the form field values within the same form component?

wmertens:
[...]

PR #3011, opened 2017-06-04
formValues: add decorator

wmertens:
You can either use `formValues(path1, path2, ...)` or `formValues({prop1: path1, prop2: path2, ...})` to get values from the current form as props.

Implements #2996


wmertens:
Updated with more tests.

erikras:
Looks good. Can you write the docs for this?

src/immutable/formValues.js
```diff
@@ -0,0 +1,4 @@
+import createValues from '../createValues'
```
erikras: Should this not be `'../createFormValues'`?

wmertens: Wow! How do the tests even pass? 😕

wmertens: ah yes the tests create their own. Not ideal…

wmertens:
Where do I go for writing docs?

wmertens:
@erikras Alright, it's now testing the actually created `formValues`, and docs are added.

erikras:

Issue #2798, opened 2017-04-08
Redux-form compatibility with RichText editor (draft-js) control


buscemrx:
Hello:
I am using a rich text editor control (draft-js)  on a redux form. The issue I am having is that the object  returned to redux-forms is a ContentState  type by the OnChange function which causes an iodash deprecation error. Returning this ContentState object value is the only way to get Redux-forms to work for me using this control but getting warning errors.

Does redux-forms support data value types other than strings, numbers, and arrays? It appears the Rich TextEditor (draft-js) is a ContentState object thus causing parsing errors within redux-forms.

This is the warning that appears everytime the OnChange or a key is pressed within the RichTextEditor field:

immutable.js:4655 iterable.length has been deprecated, use iterable.size or iterable.count(). This warning will become a silent error in a future version. Error
    at ContentState.get (webpack:///C:/Users/bbuscemi/Documents/AE2S/Origin-Development/ae2s/AE2S/~/draft-js/~/immutable/dist/immutable.js?:4650:21)
    at baseIsTypedArray (webpack:///C:/Users/bbuscemi/Documents/AE2S/Origin-Development/ae2s/AE2S/~/lodash/_baseIsTypedArray.js?:57:19)
    at arrayLikeKeys (webpack:///C:/Users/bbuscemi/Documents/AE2S/Origin-Development/ae2s/AE2S/~/lodash/_arrayLikeKeys.js?:26:47)
    at keys (webpack:///C:/Users/bbuscemi/Documents/AE2S/Origin-Development/ae2s/AE2S/~/lodash/keys.js?:34:32)

buscemrx:
I saw a previous post on this issue that someone recommended react-rte which is an enhancement to the draft-js component. That enhancement returns a string format instead of an object. I used that and it works great with redux-forms. I do like redux-forms and appreciate Erik's work very much. Thanks.

PR #3019, opened 2017-06-05
add support for props with nested immutables

mattkrick:
Background:
Sometimes, a prop might be a POJO that contains a nested immutable structure. Draft-js' `editorState` is a great example. Within sCU, the plain `deepEqual` function is used on it & breaks. 
 
Why this is needed:
- Per #2798 the only way to make this work with draft-js is to serialize the editorState.
- The beauty of immutables is that we don't have to do a deep equals, comparing references is enough
- folks may use different immutable libraries so there is no safe way to ducktype an immutable
- folks may have large POJOs that they want to exclude from being deep equaled in sCU. This is a cheap, unobtrusive way to add performance. 

Example:
`reduxForm({form: 'linkChanger', immutables: ['editorState']})`

can bump the docs if you like it

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -528,12 +528,10 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
         dom,
         'button'
       )
-        TestUtils.Simulate.click(initButton);
-
-      // no need to rerender form on initialize
+      
+      TestUtils.Simulate.click(initButton);
+      
       expect(formRender.calls.length).toBe(2)
-
-      // check rerendered input
       expect(inputRender.calls.length).toBe(1)
     })
     
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -475,6 +475,68 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
       expect(spy.calls.length).toBe(9)
     })
 
+    it('should strict equals props in immutableProps', () => {
+      const store = makeStore({})
+      const inputRender = createSpy(props => (
+        <input {...props.input} />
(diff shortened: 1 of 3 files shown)


erikras:
From your description, this sounded like a massive change, and then it only turned out to be two little lines.

So all you're doing is allowing additional "props not to update for" to be specified. It's unclear to me what this has to do with immutability or how it would help with Draft JS. Why are you passing `editorState` to your form as a prop?

erikras:
The problem I see, though, is that you are totally ignoring changes to the props listed. It seems like you should be `===`ing them, right?

Let's call it `immutableProps`.

mattkrick:
you're correct, i was being selfish & just solving for my use case.
Fixed & documented! if a unit test is needed lemme know where.

erikras:
Can you get the code coverage back up to 100%?

`npm run test:cov`
`open coverage/lcov-report/index.html`

src/__tests__/reduxForm.spec.js
```diff
@@ -475,6 +475,68 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
...
+      expect(inputRender.calls.length).toBe(1)
+
+      // initialize
+      const initButton = TestUtils.findRenderedDOMComponentWithTag(
+        dom,
+        'button'
+      )
+        TestUtils.Simulate.click(initButton);
+
+      // no need to rerender form on initialize
```
gustavohenke: This comment is kinda misleading, but IMO fix it if you want 😁

mattkrick: good catch, i was just blindly cut & pasting :smile:

erikras:
Fantastic!

mattkrick:
oh super duper, thank you!

erikras:

PR #3022, opened 2017-06-06
add support for plain objects in keys extractor from immutable structure

adam187:
Reason for this change is that for immutable structure `valid` prop  from `reduxForm()` is always `false` 

That is happening because `isValid` selector in this place
https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L705
is using `getFormState` that returns plain object so `registeredFields` here
https://github.com/erikras/redux-form/blob/master/src/selectors/isValid.js#L36
are plain object and `keys(registeredFields)` returns empty list from https://github.com/erikras/redux-form/blob/master/src/structure/immutable/keys.js#L3

After this change it seams to work as expected.

### DIFF
--- a/src/structure/immutable/__tests__/keys.spec.js
+++ b/src/structure/immutable/__tests__/keys.spec.js
@@ -25,4 +25,15 @@ describe('structure.immutable.keys', () => {
       List(['a', 'b', 'c'])
     )
   })
+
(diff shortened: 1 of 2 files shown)


erikras:
Looks good. Thanks!

erikras:

PR #2363, opened 2016-12-29
Implement `getFormAsyncErrors` selector

mistadikay:
It was requested in #2201 but was never implemented. The implementation and tests are pretty much similar to `getFormSyncErrors` unless I'm missing something. Tested on the application I'm working on.

PR #3023, opened 2017-06-06
use immutable structure for selectors

adam187:
I'm not quire sure if this is bug or not because i'm not using this selectors for anything, but I'm assuming that they should use `immutable` structure like other selectors. 

I ran on it while debugging issue from this PR https://github.com/erikras/redux-form/pull/3022


erikras:
Great catch, thanks! Was a mistake on #2363.

erikras:

PR #3033, opened 2017-06-08
INITIALIZE should not insert null versions of objects into arrays for deleted indicees

ncphillips:


ncphillips:
I'm not entirely convinced that this is the right solution for the actual problem I'm facing. It actually seems to introduce a bug where new pristine values might not be added to the current working document.

The better solution might be for the ARRAY_REMOVE reducer to go through `registeredFields` and remove all fields whose name starts with `'${fieldName}.${index}'`. This may not actually solve the problem, and could result in a similar bug.

ncphillips:
We're working with some potentially complex forms and having this reconciliation work properly is pretty important. That said, I may be able to help add more tests to the spec.

ncphillips:
Sorry I didn't pay much attention to the code style because I thought it was setup to not commit unless es-lint passed...that must have been removed.

And yes, that is correct, it is the only test that covers the bug I am experiencing. The other two tests were to confirm that the change I made didn't introduce new bugs.

ncphillips:
We're running a fork with this fix in it until this is out is out. Is there an expected date for 6.7.1 to be released?

erikras:

Issue #3041, opened 2017-06-10
Documentation (Live Web)  Immutable JS Example v6.8.0 Webpack Bundle Error 


N0taN3rd:
### Are you submitting a **bug report** or a **feature request**?

**Bug Report**

### What is the current behavior?
![reduxformexamplesimmutablebundleerror](https://user-images.githubusercontent.com/8648317/27005330-8110de82-4dea-11e7-8d93-58c04c534aa2.png)

### What is the expected behavior?
![reduxformimuttableexpected](https://user-images.githubusercontent.com/8648317/27005354-2ea66ec2-4deb-11e7-8065-a53cefcef6e2.png)

### Sandbox Link
http://redux-form.com/6.8.0/examples/immutable/

### What's your environment?

Firefox 53.0.3 and Chrome 59.0.3071.86 (Official Build), Ubuntu 16.04 

### Other information
Just saw that you updated the project to 6.8.0 and added ``immutableProps`` to the form config :+1: 
So I went to the Immutable JS example to see if it was updated (habitual check if examples changed when the docs have changed) thus this issue.

Hope this helps in bringing a quick fix. But if no time is available to give to it I would be more than happy to address it when I myself get some time later today or tomorrow.

Love the project, thank you @erikras and contributors for the continued development

erikras:
Yep, working on it. Same as #3038.

That's what I get for releasing late on a Friday. 🤦‍♂️

Issue #3038, opened 2017-06-09
Simple Form example fails to build locally 


milliechan:
Simple Form never loads after installing dependencies. I get this error on localhost:3030
```
Uncaught Error: Cannot find module "./App.scss"
    at eval (bundle.js:1)
    at Object.o (bundle.js:1)
    at t (bundle.js:1)
    at Object.defineProperty.value (bundle.js:22)
    at t (bundle.js:1)
    at Object.e.exports (bundle.js:1)
    at t (bundle.js:1)
    at eval (bundle.js:1)
    at eval (bundle.js:1)
    at eval (bundle.js:1)

```

I also had to manually install webpack-dev-middleware as that seemed to be missing.. 
Looks like that file is coming from the website-template module and that might be the reason why it's failing 

Steps to reproduce:
cd examples/simple 
npm i 
npm start

timhwang21:
Ran into this just now as well, for what it's worth, using the scripts at the root `package.json` works: `npm example:immutable`

erikras:
This is a known problem, thanks. I'm working on it.

Some webpack upgrade broke it, I think.

erikras:
This should be fixed now. Thanks!

Issue #2258, opened 2016-12-10
Sync validation 'warn' not executed - redux-form/immutable


somghosh:
It seems that the warning validation function for redux-form/immutable is not working. my console log line is never printed (or debugger never hit).I am using `"redux-form": "^6.3.2",`

	import {Field} from 'redux-form/immutable';
	import {reduxForm} from 'redux-form/immutable'

	const warn = (values, props) => {
		console.log('in warning')
		const warnings = {}

		debugger
		if (!values.get('email') =="XX") {
			warnings.email = 'XX'
		}

		return warnings
	}

	// Decorate the form component
	AccountSettingsBase = reduxForm(
		{
			form: 'accountSettingsForm', // a unique name for this form
			validate,
			warn: warn,
			enableReinitialize: true,

		})(AccountSettingsBase);

timhwang21:
Hmm, validation with the same function works properly? I'll take a look when I have a chance.

oakley808:
@timhwang21 Here's a codepen showing the standard redux-form/immutable 6.7.0 example with warn added in... and `warn` never firing like @somghosh noted. 

https://codesandbox.io/s/73v1N67M8

Am I missing something?

timhwang21:
@oakley808 for what it's worth, I fixed the incorrect import of `warn` in the code sandbox example, and the warning fired correctly. Is it still not working correctly in your app?

oakley808:
@timhwang21 Awesome! Thank you.

gustavohenke:
@timhwang21 thanks for the help!

@somghosh are you still tracking this?

gustavohenke:
Closing due to inactivity.

PR #3042, opened 2017-06-10
Add `warn` to Immutable demo form

timhwang21:
Had a discussion a while back in #2258 where there was some confusion regarding using `warn` in immutable forms, and thought this might help to reduce future confusion.

However I haven't verified my changes yet because of the issues described in #3041 and #3038.


timhwang21:
@gustavohenke Thanks for the approve! I do want to check that I didn't make any dumb errors once the issues I linked re: the docs are fixed :)

PR #3044, opened 2017-06-12
Updating documentation

sarah-due:
By defining ContactForm with 'const' on line 73, it cannot set using reduxForm below, line 82.


gustavohenke:
Thanks, but your change is the same as #3043.

PR #3045, opened 2017-06-12
Fix naming discrepancy

jschneier:
`handleSubmit` is what appears in `SimpleForm.js`.


gustavohenke:
`handleSubmit` is a prop provided by the `reduxForm` HOC.
It wraps external `onSubmit` prop and triggers necessary Redux changes.

Therefore, the example is correct.

jschneier:
Thanks for the explanation.

PR #3064, opened 2017-06-15
Do not clear errors unless field is being removed from the form

tikotzky:
Currently errors will be cleared every time an `UNREGISTER` action is fired even if the count is greater than 1.

This would cause the errors to get cleared even when there is still a form field on the page.
_you can see a demo of the error [here](https://codesandbox.io/s/ADjL8EGD1)_

This PR changes the behavior so that errors will only be cleared if the count is <= 0 and the field is actually being removed from the `registeredFields`

### DIFF
--- a/src/__tests__/reducer.unregisterField.spec.js
+++ b/src/__tests__/reducer.unregisterField.spec.js
@@ -32,6 +32,28 @@ const describeUnregisterField = (reducer, expect, { fromJS }) => () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


tikotzky:
Thanks!
Not aware of any specific issue. I uncovered this bug while looking into https://github.com/erikras/redux-form/issues/2742

erikras:
Excellent work. Man, I really ❤️ CodeSandbox. It was so easy for you to clearly demonstrate this bug.

erikras:

Issue #2953, opened 2017-05-23
Bug: STOP_ASYNC_VALIDATION erases previous asyncErrors if currently blurred field passes validation


davedx:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

1. Build a form with two inputs with asyncValidation
2. Invalid one of the inputs
3. Enter something in the second input that is valid (causes no errors)
4. STOP_ASYNC_VALIDATION does this: result = deleteIn(result, 'asyncErrors') https://github.com/erikras/redux-form/blob/master/src/createReducer.js#L341

### What is the expected behavior?

asyncErrors still contains the error from the first input so the form can continue showing this field is invalid.

### What's your environment?

Redux Form 6.6.3

### Other information

I'm opening this as a bug report to start a discussion to at least confirm this behaviour is a bug. If so I'm happy to help work on a PR to fix it once I understand it properly. Is asyncErrors *supposed* to be state shared between fields like this?

nktssh:
Any updates on this?

davedx:
Funny you should ask, I just made a PR. Looking forward to feedback.

nktssh:
@davedx you can find workaround for this here: https://github.com/erikras/redux-form/issues/980#issuecomment-278360610

Just merge previous errors and current error (or delete the key, if no errors for current input)

erikras:
Fix published in [`v7.0.0`](https://github.com/erikras/redux-form/releases/tag/v7.0.0).

davedx:
I made a PR that was merged

PR #3068, opened 2017-06-16
Don't clear all asyncErrors on stopAsyncValidation

davedx:
Fixes https://github.com/erikras/redux-form/issues/2953

The reason this makes sense to do, is that the field specific asyncErrors are cleared already by events like CHANGE, so it's not necessary to clear all asyncErrors in the STOP_ASYNC_VALIDATION action.

This allows us to have more than one field with async validation in the same form. When one field fails async validation but then another field passes, it no longer clears asyncErrors for the whole form, therefore setting the first field to a valid state incorrectly.

### DIFF
--- a/src/__tests__/reducer.stopAsyncValidation.spec.js
+++ b/src/__tests__/reducer.stopAsyncValidation.spec.js
@@ -111,56 +111,6 @@ const describeStopAsyncValidation = (reducer, expect, { fromJS }) => () => {
     })
   })
 
-  it('should unset field async errors on stopAsyncValidation', () => {
-    const state = reducer(
-      fromJS({
-        foo: {
-          values: {
-            bar: ['dirtyValue', 'otherDirtyValue']
-          },
-          initial: {
-            bar: ['initialValue', 'otherInitialValue']
-          },
-          asyncErrors: {
-            bar: ['async error 1', 'async error 2']
-          },
-          fields: {
-            bar: [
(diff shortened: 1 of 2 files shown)


davedx:
@gustavohenke so we need to run the validation for every field in one place? OK, I'll look into that (I'm back at work on Tuesday), closing the PR until I've tried it.

If that's true it would be great to update the docs to reflect it, it's a common use case. I can maybe help out there. I'll get back to you.

davedx:
@gustavohenke I'm now confused when I think about this.

Do you mean my asyncValidate function should re-run the async validation API calls for every field every time it's called? This seems very wasteful. Or do you mean I should cache somewhere the results of validation for every field except the one being changed? That seems overly complex. Then we would be storing the state of each field separately outside of redux form? Are there any examples of how this should work?

gustavohenke:
I agree with most of your points here (complex, wasteful, etc).

A not-so-complex caching solution would be to use redux-form selectors: https://codesandbox.io/s/1jEp9JYBP

Anyway, I still think this PR could be merged if @erikras decides it's a good addition.
I'm not against it 😊

davedx:
OK, thanks again - looking forward to hearing what he thinks. :)

erikras:
At first I was ambivalent, but the case laid out in #2953 convinced me. 👍

erikras:

PR #3072, opened 2017-06-17
Use same empty object for syncErrors and syncWarnings

JeffreyATW:
I noticed that my form and its fields were unnecessarily re-rendering upon any changes to my store. I realized that React Redux's shallow comparison was failing when comparing `syncWarnings` between states. This is because Redux Form's connector returns a new object every time when there are no `syncWarnings` or `syncErrors`.

Changing this code to use the `empty` object makes shallow comparison work properly, reducing unnecessary re-renders significantly.


erikras:
Great catch! 👍

erikras:

Issue #3061, opened 2017-06-14
Component not reflecting state in store


helmus:
### Are you submitting a **bug report** or a **feature request**?

Bug Report

### What is the current behavior?

I created a very simple yes / no component mapped to true / false values.
When clicking "No" the value is updated in the store, but the component does not update.
Only after clicking Yes first the component updates when clicking No.

### What is the expected behavior?

The component should reflect the store in the state.

### Sandbox Link

https://codesandbox.io/s/r06VKjB4K

### What's your environment?

redux-form 6.8
redux 3.6
react-redux 5.0.4

### Other information

I apologize if this is not a bug, but this seems extremely counter intuitive for users trying to build custom components. My hunch feeling says this is a bug because somewhere the state isn't being strictly compared.

### Update:

Adding a formatter or a parser that does nothing ( `format={value=>value}` ) has a side effect of fixing the problem. This amplifies my feeling that this is in fact a bug.

erikras:
It's interesting that if you made the values `0` and `1`, it works. So it's not about being "falsy".

erikras:
[Here's the bug](https://github.com/erikras/redux-form/blob/1f08e34d0f9809898396b3e40ab1e1ba0966377c/src/structure/plain/deepEqual.js#L5-L8).

I'm trying to figure out why this was written like this.

erikras:
This is related to the hotly contested #1762, #1774, and #1993.

I think that @clayne11 is probably correct on this issue. `false` should be a valid value, and if you need a checkbox to go back to `pristine` when it gets unchecked, you will have to initialize it to `false`.

And fixing this will require a breaking change release.

helmus:
agree 100% that true, false should be valid values. I also think that null, undefined and `''` should be treated as distinct valid values to.

erikras:
>  I also think that `null`, `undefined` and `''` should be treated as distinct valid values too.

That's a harder case to make, as an empty field comes back as `''`, and I think it _is_ important for a form with no initial values to still report as `pristine` if you type into a field and erase the value. This is a personal opinion of mine.

zcherries:
I'm also having this problem.

erikras:
⚠️ __IMPORTANT__ ⚠️  As of [`v7.0.0`](https://github.com/erikras/redux-form/releases/tag/v7.0.0), `undefined` and `false` are no longer equal. [Read more](https://github.com/erikras/redux-form/releases/tag/v7.0.0).

PR #3081, opened 2017-06-19
BC: Treat undefined and false as NOT equal

erikras:
# ⚠️ BREAKING CHANGE ⚠️

Addresses #3061, #1762, #1774, and #1993.


gustavohenke:
`v7.0.0` incoming?? 😀

erikras:
Not merging yet until I'm ready to pull the trigger, but yes. This lib needs to be more respectful of semver.

helmus:
@erikras Any idea when this would be released ?

erikras:

dwick:
@erikras shouldn't this also treat `null` and `false` as not equal. upgraded and immediately got bit by that.

PR #3089, opened 2017-06-21
Fixed rerenderOnEveryChange for FieldArray with same groups (#3008)

vtaits:
#3008 

1. Fixed behavior in `shouldComponentUpdate` method of `ConnectedFieldArray`
2. Added test `should rerender when a value changes if rerenderOnEveryChange is set and FieldArray has multiple groups with same values`


nihaux:
This also solves the following problem: when you change the order of items in field array, it does not update. cf https://github.com/erikras/redux-form/issues/3008#issuecomment-311063934

erikras:

Issue #3095, opened 2017-06-23
Initialize with keepDirty and empty array creates an empty object in form values 


percevalw:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

Hi, I'm working on an array structure form. When I delete a field in the array list that was before set to a 'empty-like' value `{name: ""}`, and triggers initialize with keepDirty option and an new empty initial value to make it pristine, the ReduxForm reducer creates an empty object in the state values.

### What is the expected behavior?
The values array should not be altered by the reducer as the keepDirty parameter is set to true and the field indeed dirty.

### Sandbox Link

https://www.webpackbin.com/bins/-KnHcmQvY3yKJ4ZF8m0C

### What's your environment?

Any environment

### Other information
```javascript
import { reducer, initialize } from 'redux-form';
import structure from 'redux-form/es/structure/plain';
const { deepEqual, fromJS } = structure;

const before = {
  myForm: {
    registeredFields: {
      // field has been deleted in array before so fields have been unregistered
      myList: { name: 'myList', type: 'Field', count: 0 },
      'myList.0.name': { name: 'myList.0.name', type: 'Field', count: 0 },
    },
    values: {
      myList: [], // field has been deleted in array before
    },
    initial: {
      myList: [{ name: '' }], // but was still present in initial values, with an empty string value
    },
  },
};

// let's submit the data to the server, and the form will update its initialValues
// either programmatically with an action or with enableReinitialize for example
// so the new initialValues should now be
const actionInitialValues = [];
const actionKeepDirty = true;

const state = reducer(
  fromJS(before),
  initialize('myForm', actionInitialValues, actionKeepDirty),
);

console.log(JSON.stringify({
  myForm: {
    registeredFields: {
      // this part is not affected
      myList: { name: 'myList', type: 'Field', count: 0 },
      'myList.0.name': { name: 'myList.0.name', type: 'Field', count: 0 },
    },
    values: {
      // this is the problematic part, as it should be empty
      myList: [],
    },
    initial: {
      // new initial values are indeed modified to empty array
      myList: [],
    },
  },
}, null, 2));

console.log(JSON.stringify(state, null, 2));
```
```diff
       }
     },
     "values": {
-      "myList": []
+      "myList": [
+        {}
+      ]
     },
     "initial": {
       "myList": []
```

In the `createReducer/js` file :
[...]

PR #3105, opened 2017-06-24
Fix initialize with keepDirty and empty array creates an empty object in form values (#3095)

percevalw:
Hi, I made this fix for #3095.

It prevents the initialize action with keepDirty set to true to create 'empty objects' `{}` uselessly if the field to keep pristine has to change to `undefined`. This happens because of the `setIn` function that creates all the objects required for the nested path.

Example : the initialize action in the new test sets the value of `myList.0.myField` to same as the one in `{myList: []}`.
Before the fix, it created an empty object in the values `{myList: [{}]}`, which for ReduxForm is the same as `{myList: [{myField: undefined}]}`.
Now it leaves the values like this `{myList: []}`

### DIFF
--- a/src/__tests__/reducer.initialize.spec.js
+++ b/src/__tests__/reducer.initialize.spec.js
@@ -446,6 +446,52 @@ const describeInitialize = (reducer, expect, { fromJS }) => () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


erikras:
Nice one. 👍

erikras:

Issue #2842, opened 2017-04-22
Remove peerDependencies with `react@^15.0.0`


giautm:
Because i using `react@16.0.0-alpha.6`, peer dependency with v15 make a copied of react and cause a problem.
Current workaground is remove `node_modules/redux-form/react`
https://github.com/erikras/redux-form/blob/master/package.json#L109

Please move react to `dependencies` or set `peerDependencies` to `"react": "^15.0.0 || 16.0.0-alpha.6"`

gustavohenke:
👍 for the second alternative

giautm:
Workaround steps:
Step 1:
```
yarn add rimraf -D
```

Step 2: Add following script to `package.json` under `scripts`
```
    "postinstall": "rimraf ./node_modules/redux-form/node_modules/react"
```
Now, after install any module, `./node_modules/redux-form/node_modules/react` will auto remove.

<img width="455" alt="screen shot 2017-05-05 at 23 23 31" src="https://cloud.githubusercontent.com/assets/12751435/25754687/5da3c3ae-31ea-11e7-8b14-e1965013a104.png">

PR #3111, opened 2017-06-26
Extend React Peer dependency to version 16

cliedeman:
Redux-form creates a duplicate react install in react-native due to the React 15 peer dependency

Taken from [react-redux](https://github.com/reactjs/react-redux/blob/master/package.json#L112)

Tested on React Native 45

Ciaran


giautm:
I will close https://github.com/erikras/redux-form/issues/2842 after this merged. 👍

Issue #2763, opened 2017-04-02
Feature request: Async field level validation


lloydh:
It would be great to be able to use promise-based validation libraries like yup for field level validation but it seems that redux-form only supports sync validation functions on fields.

There's a [stack overflow post](http://stackoverflow.com/questions/40363850/is-it-bad-practice-to-repurpose-asyncvalidate-for-integrating-a-promise-returnin) and [this article](http://www.davidboyne.co.uk/2016/04/29/redux-form-validation-with-yup.html) discussing use of the asyncValidate config option but I don't think there are any options for async field level validation.

jorrit:
I would like this feature as well. It is a bit cumbersome to set the async function and asyncBlurFields on the form level. It would be nicer to set asyncValidate on a field and get the same effect, just like with synchronous validation.

roytz:
Will be very useful to me too! Will this feature be available anytime soon?
Thanks!

jorrit:
Perhaps a maintainer can make an estimate of the effort needed. If it's not too much I could take a look.

shai32:
thats will be great

yussinsharp:
Wrote some code that allowed me to compose asyncValidator functions into one. Thought I'd share here:

```js
function composeAsyncValidators(validatorFns) {
  return async (values, dispatch, props, field) => {
    let errors = {};
    for (const validatorFn of validatorFns) {
      try {
        await validatorFn(values, dispatch, props, field);
      } catch (err) {
        errors = Object.assign({}, errors, err);
      }
    }

    throw errors;
  };
}

// Example Usage - need to call multiple API endpoints to validate data

function uniqueEmail() {
  return axios.post('/users/phoneNumber', { phoneNumber: values.phoneNumber })
    .then((response) => {
      if (response.data.exists) {
        throw { phoneNumber: 'There is already a user with that phone number' };
      }
    });
}

function uniquePhoneNumber() {
  return axios.post('/users/phoneNumber', { phoneNumber: values.phoneNumber })
    .then((response) => {
      if (response.data.exists) {
        throw { phoneNumber: 'There is already a user with that phone number' };
      }
    });
}

const asyncValidate = composeAsyncValidators([uniqueEmail, uniquePhoneNumber]);
```

Field-level async validation would've been preferred but this also solves the case of composing functions that depend on multiple fields (or throw errors for multiple fields).

jorrit:
Isn't something involving `Promise.all()` appropriate here?

[...]

PR #3134, opened 2017-06-29
Async validation support for multiple fields

bitstrider:
Addressing issue #2763


bitstrider:
@gustavohenke appreciate the feedback,

While this PR doesn't yet implement field-level async validation via a prop on the `Field` component, it does take a step towards it by fixing how the `asyncErrors` state slice (which is suppose to contain a map of errors by field names) is updated with incoming async validation errors: the incoming async validation errors should be merged into the existing map, rather than replacing it entirely. 

Without this behavior, async validations for multiple fields is broken in general, because you're lose existing async validation errors every time there's a new one from a different field.

I'm a bit unclear how my code merges submit validation errors with async validation errors, since the former are stored in the `submitErrors` state slice (and the latter in the `asyncErrors` state slice). Maybe you could clarify?

If what you meant was how to account for `asyncValidate` also being called in `handleSubmit`,  I'm currently using this workaround, using code from @yussinsharp in #2763, which doesn't touch `handleSubmit`:

```javascript
function composeAsyncValidators(validatorFns) {
  return async (values, dispatch, props, field) => {
 
    if(field) { // then this is an asyncValidate call from onBlur for that field
      const validatorFn = validatorFns[field]
      return validatorFn(values, dispatch, props, field);

    }else{ // field arg is undefined, indicating this is an asyncValidate call from onSubmit
      // OPTION 1: perform all async validations, delaying submission
      let errors;
      for (const validatorFn of Object.values(validatorFns)) {
        try {
          await validatorFn(values, dispatch, props, field);
        } catch (err) {
          errors = Object.assign({}, errors, err);
        }
      }

      if (errors) throw errors;
      else return new Promise((resolve)=> {resolve()}) // since all async validations passed

      // OPTION 2: skip async validations and just return a resolved promise
      //return new Promise(resolve=> resolve())
  };
}


const usernameValidate = (values /*, dispatch, props, field */) => { /* ... */ }
const companynameValidate = (values /*, dispatch, props, field */) => { /* ... */ }


const asyncValidate = composeAsyncValidators({
  username: usernameValidate,
  companyname: companynameValidate
});


MultipleAsyncForm = reduxForm({
  form: 'multipleAsync',
  asyncValidate, /* composed of many field-level validators */
  asyncBlurFields: ['username', 'companyname']
})(MultipleAsyncForm)
```

PR #3138, opened 2017-06-30
🎉 Flow Typing 🎉

erikras:
This was a pretty huge project, and I'm certain that mistakes have been made, but we might need to rely upon the community to find them all.


gausie:
@erikras the action creators need to indicate which arguments are optional.

For example

```javascript
const change: Change = (
  form: string,
  field: string,
  value: any,
  touch: boolean,
  persistentSubmitErrors: boolean
): ChangeAction => ({
  type: CHANGE,
  meta: { form, field, touch, persistentSubmitErrors },
  payload: value
})
```
should be
```javascript
const change: Change = (
  form: string,
  field: string,
  value: any,
  touch?: boolean,
  persistentSubmitErrors?: boolean
): ChangeAction => ({
  type: CHANGE,
  meta: { form, field, touch, persistentSubmitErrors },
  payload: value
})
```
at least according to the API docs

gustavohenke:
Hi @gausie, please open a new issue for this.

PR #3154, opened 2017-07-06
Add missing initial prop to Fields

hswhite33:
fieldProps.meta.initial was not being set for Fields components.


erikras:
Looks good. Thanks!

PR #3170, opened 2017-07-11
More Flow typing prep

erikras:
Preparing for full flow support in v7.


src/actions.js
```diff
@@ -607,3 +400,41 @@ export const updateSyncWarnings: UpdateSyncWarnings = (
...
+  setSubmitFailed,
+  setSubmitSucceeded,
+  touch,
+  unregisterField,
+  untouch,
+  updateSyncErrors,
+  updateSyncWarnings
+}
+
+export default actions
```
huan086: @erikras I think you missed out documenting this breaking change in v7. No longer `import { actionName } from '../actions';`, but `import actions from '../actions'; const { actionName } = actions`. Not sure if it's possible to edit the v7 release notes

Issue #2872, opened 2017-05-01
_error key inaccessible outside of form


TrevorHinesley:
As the title suggests, if stopSubmit is called, I can call it like:

`stopSubmit(FORM, {_error: err.message})`

The form will receive the error in its props, but the wrapper component of my form (which uses the selector `getFormSubmitErrors`), will be undefined. If I change the key from `_error` to anything else, it will work properly.

mitas1:
Can you send a pull request @tgallacher please?

gustavohenke:
Duplicate of #3325

PR #3180, opened 2017-07-13
Add selector for form error (_error)

tgallacher:
Provides a selector to retrieve the "_error" key for a given form error.

This provides a solution to bug #2872


erikras:
Did you actually test this? The form-wide error is under `error`, not `_error`.

tgallacher:
Ah. So I take it that `_error` is aliased to the `props.error` key? 
If so, never saw that. This PR was inspired by the comment 

> Also note that a general form-wide error can be returned via the special _error key.

from http://redux-form.com/7.0.1/examples/submitValidation/. 

Closing as this is obviously not needed. 
thanks,
t

Issue #3184, opened 2017-07-14
Using FormValueSelector selector as mapStateToProps returns undefined (v7.0.0)


prattl:
### Are you submitting a **bug report** or a **feature request**?
**Bug report**

### What is the current behavior?
Using a selector created from `FormValueSelector` as `mapStateToProps` does not work as intended from the [API docs](http://redux-form.com/7.0.0/docs/api/FormValueSelector.md/). Item 3 under the Usage section indicates that this selector can be used directly as a `mapStateToProps` function, but it appears to return `undefined` when used in this manner.

### What is the expected behavior?
The selector itself should work just fine as mapStateToProps, or the docs should be changed to reflect this is not the case.

### Sandbox Link
https://codesandbox.io/s/nr9ZNnJnY

### What's your environment?
```
redux-form@7.0.0
react@15.6.1
redux@3.7.1
react-redux@4.4.8
```

```
ubuntu 16.04
Google Chrome 58.0.3029.81
node v.8.1.4
```

gustavohenke:
Hi @prattl!
From the docs we can infer that selecting a single field returns its value directly, without wrapping in an object. Therefore you were returning a string in your `mapStateToProps` 😀 
In this case, you should return an object, or select more fields (in case you need them...).

You can take a look at this working sandbox: https://codesandbox.io/s/wy8NYRMJ

*Additionally*, you could perhaps use [`formValues` decorator](http://redux-form.com/7.0.0/docs/api/FormValues.md/) instead of the selector 😉

prattl:
Ah I see now, thanks for the help! Apologies for the false bug report. I think it'd be handy to call out in the docs that using the selector as a `mapStateToProps` will only work if selecting more than one value. I'll open a PR for that.

PR #3191, opened 2017-07-17
Docs: Add note indicating that FormValueSelector can only be used as mapStateToProps if selecting multiple fields

prattl:
Might help others avoid a bit of confusion like I encountered in #3184.


gustavohenke:
Perfect! Thanks for _blaming_ the docs and for _actually fixing_ them! ❤️

PR #3199, opened 2017-07-18
Remove custom prop from fieldPropTypes

eemeli:
The `custom` object is not included in the props given to a decorated form component; instead custom props are included directly in the props object. As `custom` is nevertheless marked with `isRequired` in `fieldPropTypes`, the proptype isn't actually usable as it'll always fail.


erikras:

Issue #3196, opened 2017-07-18
[7.0.0] Deprecated warning from immutablejs (iterable.length)


giautm:
### **bug report**
Deprecated warning from immutablejs

### What is the current behavior?

Show Deprecated warning from immutablejs:
> iterable.length has been deprecated, use iterable.size or iterable.count()

### What is the expected behavior?
Don't show up warning.

### What's your environment?
```
"react": "16.0.0-alpha.12",
"redux": "^3.6.0",
"redux-form": "^7.0.0",
"immutable": "3.7.6",
```
### Other information
![image](https://user-images.githubusercontent.com/12751435/28296814-a9e00dde-6b94-11e7-9234-13586610307f.png)

erikras:
I went to test this, and installed the latest version of `immutable`, which is `3.8.1`, and I got no stack trace. So I tried your specific version, `3.7.6`, and saw that the error says, **This warning will become a silent error in a future version**. 🤦‍♂️ What sense does that make? Is it an error or not? Seems like questionable library management, but perhaps they have their reasons.

Regardless, `redux-form` should not be doing a plain `deepEqual` check on an immutable `Map`, so this really is a bug.

erikras:
Okay, my previous conclusion was not entirely correct. Sometimes this error is being caused by lodash's `deepEqual` checking if the value is "array like", by seeing if it has a numeric `.length` attribute. That hardly seems like a sin worth throwing up a warning for. As such, it _does_ sort of make sense that immutable.js went through one version where it warned everybody who might have been checking `.length` the old way, but that it _should_ "fail" silently in future versions.

I did find one bug in my investigation, so that's good. However, lodash is certainly not committing any crime by probing an attribute on a value.

### TL;DR:

This is not really a problem. Upgrade to immutable.js `3.8.x`.

erikras:
Semi-fix published in [`v7.0.1`](https://github.com/erikras/redux-form/releases/tag/v7.0.1).

giautm:
So fast. 😄

PR #3201, opened 2017-07-18
Sync warnings and errors should default to {}

erikras:
This is a bug found whilst investigating #3196.

Even when using `redux-form/immutable`, the sync errors and warnings are plain javascript objects.

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -869,8 +869,8 @@ const createReduxForm = (structure: Structure<*, *>) => {
 
           const pristine = shouldResetValues || deepEqual(initial, values)
           const asyncErrors = getIn(formState, 'asyncErrors')
-          const syncErrors = getIn(formState, 'syncErrors') || empty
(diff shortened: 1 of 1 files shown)


erikras:

simonify:
Sorry for the bump on this, but this change seems to revert the optimization made in https://github.com/erikras/redux-form/pull/3072. Is there a specific reason not to cache the empty object or would a PR be welcomed to do that?

gustavohenke:
I'm not a lot into this issue, but I'm thinking `empty` (imported from the structure) doesn't fit well the immutable case.

We could try using the same `{}` reference instead of creating a new object every time and instead of importing that `empty`. Would you like to try that? PRs are _always_ welcome, by the way! 😃

simonify:
It looks like `empty` from `structure/plain` is just an empty object literal so I think it shouldn't be an issue if defining an object here is OK.

I'll whip up a PR so discussion can take place there instead.

Issue #3202, opened 2017-07-18
Updating the name of a FormSection does not update props of formValues properly


SBRK:
It seems when I update the `name` of my `FormSection`, the `formValues` always use the first `name` that was set as their prefix.

In the following example, if displaying `MyFormComponent` with `index = 0` then changing index to 1, it would always display 'one'
```js
let MyComponent= ({ bar }) => (<div>{bar}</div>)
MyComponent = formValues({
  'bar',
})(MyComponent)

const MyFormComponent = ({ index }) => (
  <FormSection name={`foo[${index}]`}>
    <MyComponent />
  </FormSection>
)

const myFormValues = {
  foo: [
    {
      bar: 'one'
    },
    {
      bar: 'two'
    }
  ]
}
```
Edit: could it be that the context used is the one from the constructor, and it is never updated ? https://github.com/erikras/redux-form/blob/master/src/createFormValues.js#L55

SBRK:
I did a fix for this. I'll try to add a test too https://github.com/erikras/redux-form/pull/3203

erikras:
Fix published in [`v7.0.1`](https://github.com/erikras/redux-form/releases/tag/v7.0.1).

SBRK:
I'm reopening this as I still have a problem, even after the fix. It works most of the time, but in my app there is one case where the `formValues` are not updated after `FormSection` `name` was updated.

In my app, the `FormSection` name is the element of an array, whose index comes from the route (ie: `name={foo[${index}]}` with route being `/something/:index`)

When loading the page `/someting/1`, then going to another page, then hitting back in the navigator, the formValues are updated (with the wrong index), then `FormSection` name is updated, and the `formValues` are not. I think the `formValues` should be updated on `componentWillReceiveProps` with new `sectionPrefix` in the `context`, but I'm not sure how to do that without creating a new `this.Component`, which is not a good thing because it would unmount/remount whatever component is wrapped.

SBRK:
Did another fix here : #3210 
I pass the sectionPrefix from the context to the props of the connected component, and added a second argument to the selector (the props), so that connect calls it when props change

SBRK:
Also added a test

erikras:
Published yet another fix in [`v7.0.2`](https://github.com/erikras/redux-form/releases/tag/v7.0.2).

PR #3203, opened 2017-07-18
Fixed context not up to date in createFormValues

SBRK:
This fixes the problem described in #3202


erikras:

PR #3210, opened 2017-07-19
Formvalues sectionprefix update

SBRK:
Since the sectionPrefix is in the context, the connected Component would not update the values of the selector.
Passing the sectionPrefix to the props then adding props to the arguments of the selector fixes that.


SBRK:
Closes #3202 for good, hopefuly !

erikras:
Published in [`v7.0.2`](https://github.com/erikras/redux-form/releases/tag/v7.0.2).

Issue #3206, opened 2017-07-18
Using Immutable Map With GetIn


jmrapp1:
### What is the current behavior?

The bug occurs in the `getIn` method. When passing in an immutable map from the redux state, the code fails on the line with the comment `// HERE`. Immutable maps don't support the ability to get a value with the syntax used: `result = result[<key>]`. You can only get a value from an immutable map with its `get(key)` method. 

Direct Reference: https://github.com/erikras/redux-form/blob/1f08e34d0f9809898396b3e40ab1e1ba0966377c/src/structure/plain/getIn.js
```
var getIn = function getIn(state, field) {
  if (!state) {
    return state;
  }

  var path = _toPath(field);
  var length = path.length;
  if (!length) {
    return undefined;
  }

  var result = state;
  for (var i = 0; i < length && result; ++i) {
    result = result[path[i]];               // HERE
  }

  return result;
};

```
### What is the expected behavior?
The expected behavior is to be able to get a value giving some state from the redux store and a specific state name passed in as the field.

### What's your environment?

Any browser.

### Other Information
If it was just this function that caused the issue I would just recreate my own function locally, but this stems a bit deeper. I am trying to use the `formValueSelector` function to create a selector and then use that selector to find a value within an immutable state. The selector always returned `undefined` and after digging through it for a few hours I found this bug within `getIn` because the selector uses it.

jmrapp1:
Solution: Use `redux-form/immutable`

PR #3212, opened 2017-07-19
GetIn Immutable State Support

jmrapp1:
- Fix for the issue I submitted yesterday: #3206 
- The `getIn` function now has support for states to be passed in that are Immutable.
- This is a large desire because redux stores created from the `redux-immutable` package will not work with `redux-form` when trying to pull values from individual form elements.
- This is because the immutable types don't support the syntax `immutableMap["some-key"]` and instead the developer must use: `immutableMap.get("some-key")`
- This fixes that issue.

### DIFF
--- a/src/structure/plain/__tests__/getIn.spec.js
+++ b/src/structure/plain/__tests__/getIn.spec.js
@@ -4,7 +4,7 @@ import { Map } from 'immutable';
 
 describe('structure.plain.getIn', () => {
 
-  describe('Immutable Tests', () => {
+  describe('Non-Immutable Object Tests', () => {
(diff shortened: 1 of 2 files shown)


erikras:
Why are you changing the **plain** `getIn` to work with Immutable?

I don't suppose you really meant to import the Immutable-enabled version of `redux-form` [from `redux-form/immutable`](http://redux-form.com/7.0.1/examples/immutable/), did you?

jmrapp1:
Yuuup, that's what I was looking for. Must have missed it on the doc. Ignore!

erikras:
No worries. Thanks for your willing to contribute! 👍

Issue #1297, opened 2016-07-07
Async validation failing causes Uncaught promise error


kristian-puccio:
The link in the docs to rc.3 is broken but you can see this in rc.2

Go to http://redux-form.com/6.0.0-rc.2/examples/asyncValidation/ and fail the validation.
You will get a `undefined:1 Uncaught (in promise) Object {username: "That username is taken"}` error

Ruzo:
Same error for me using axios.

mdramos:
+1

jmarceli:
This problem is still visible in `rc.4`, any update on that? Maybe some workaround?

alyrik:
Still the same. It may even be reproduced with [this demo](http://redux-form.com/6.0.0-rc.4/examples/asyncValidation/)
![screenshot_2016-08-17_00-48-33](https://cloud.githubusercontent.com/assets/10388022/17717113/afda8184-6414-11e6-8c7a-f6a7b40962fd.png)

joelmukuthu:
A workaround for this involves using a [reducer plugin](http://redux-form.com/6.0.0-rc.1/docs/api/ReducerPlugin.md/). Instead of throwing an error that rejects the promise, issue a new action and handle it in the reducer plugin. For example using the example from the docs:
```js
const reducers = {
  // ... your other reducers here ...
  form: formReducer.plugin({
    login: (state, action) => {   // <----- 'login' is name of form given to reduxForm()
      switch(action.type) {
        case AUTH_LOGIN_FAIL:
          return {
            ...state,
            asyncErrors: {
               password: 'Invalid email or password' // <----- add an async error 
            }
          }
        default:
          return state
      }
    }
  })
};
```

The problem (of course :/) I found with this approach is that if you also clear the field and happen to have sync validation on the field then the sync errors will be shown instead of the async error you added, and that's because (I guess) of this https://github.com/erikras/redux-form/blob/v6.0.0-rc.3/src/createFieldProps.js#L47

NeXTs:
Due to inspecting devtools console seems to be that this issue wasn't fixed in [rc.5](http://redux-form.com/6.0.0-rc.5/examples/asyncValidation/) either :(

Strato:
Having this issue with v6.0.2...

Seems to come from here:

```
var handleErrors = function handleErrors(rejected) {
    return function (errors) {
      if (errors && Object.keys(errors).length) {
        stop(errors);
        return Promise.reject(errors);
      } else if (rejected) {
[...]

PR #3227, opened 2017-07-21
fix of the uncaught exception in promise within asyncValidation

Roman-Maksimov:


gustavohenke:
Can you prove this?

Roman-Maksimov:
Read the issue https://github.com/erikras/redux-form/issues/1297, please. It's status is `closed`, but the problem still exists.

Roman-Maksimov:
About test case, the asyn validation is not covered within `src/__tests__/Form.spec.js` at all. It might be too coprehensive issue. I think, it will be better to implement the test cases by authors of the project.

gustavohenke:
Can you confirm whether your issue is with https://github.com/erikras/redux-form/issues/1297#issuecomment-277038603?

It's the only way I found to reproduce that issue consistently.

Roman-Maksimov:
Yep. Exactly. This is the case.
If the `sync` validation is OK and `async` fails, the `uncaught promise error` from `submit` method appears.

erikras:
Published in [`v7.0.3`](https://github.com/erikras/redux-form/releases/tag/v7.0.3).

miracle2k:
I believe this leads to errors being swalled in other cases: https://github.com/erikras/redux-form/issues/3303

Issue #3247, opened 2017-07-26
Flow typing is incorrectly exported


huangbi:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

* The Field and FieldArray exports on index.js.flow have FieldProps and FieldArrayProps respectively as their props. These props are for the component passes to Field or FieldArray, not for Field and FieldArray themselves.
* The build:flow does not export the flow definitions to the es folder
* The Dispatch in FieldProps.types.js.flow should have a wildcard generic type (Dispatch<*>) or else flow complains about ambiguous polymorphic type.

### What is the expected behavior?
* The Props for Field and FieldArray in index.js.flow should be FieldInputProps and FieldArrayInputProps instead.
* build:flow should also export the flow definitions to the es folder
* Change Dispatch to Dispatch<*> in FieldProps.types.js

### What's your environment?

redux-form@7.0.2

gabrielhpugliese:
I added `.*/node_modules/redux-form/.*\.flow` to section `[ignore]` inside `.flowconfig` in the meantime.

erikras:
Fantastic reporting, @huangbi! 👍 👍

erikras:
Published in [`v7.0.3`](https://github.com/erikras/redux-form/releases/tag/v7.0.3).

ravicious:
> The Dispatch in FieldProps.types.js.flow should have a wildcard generic type (Dispatch<*>) or else flow complains about ambiguous polymorphic type.

Correct me if I'm wrong, but I believe that `*` is not "a wildcard generic type". It's [an existential type](https://flow.org/en/docs/types/utilities/#toc-the-existential-type).

> An existential type is used as a placeholder to tell Flow to infer the type.

I saw a weird cases in my apps where having `*` in the type would use a different type depending on the order of calling the function with different arguments.

I believe a wildcard [generic type](https://flow.org/en/docs/types/generics/) looks like this in Flow:

```javascript
function identity<T>(value: T): T {
  return value;
}
```

I just wanted to point it out there, because from my experience, a liberal use of `*` resulted in many weird Flow errors down the road which I've been fixing with the generic type.

erikras:
@ravicious So what should it be? `Dispatch<any>` or just `Dispatch` with no type?

ravicious:
I'm not really an expert on this, so I'm afraid I can't tell you what should be there. :(

[...]

PR #3248, opened 2017-07-26
Fix flow type exports

huangbi:
Proposed fix for flow export issues in #3247


erikras:
Ha! Thanks for this, but I just did a slightly more thorough job on #3252 before I saw this PR.

huangbi:
Sure, this was aimed for resolving my use case anyway. Thanks for the quick turnaround!

PR #3296, opened 2017-08-07
Validation not working after reset

slim25:
copy "SyncErrors", "valid" and "invalid" on RESET action from previous state to result state

### DIFF
--- a/src/createReducer.js
+++ b/src/createReducer.js
@@ -361,10 +361,6 @@ function createReducer<M, L>(structure: Structure<M, L>) {
         result = setIn(result, 'values', values)
         result = setIn(result, 'initial', values)
       }
-      var syncErrors = getIn(state, 'syncErrors');
-      if(syncErrors){
-        result = setIn(result, 'syncErrors', syncErrors);
-      }
       var submitFailed = getIn(state, 'submitFailed');
       if(submitFailed){
         result = setIn(result, 'valid', false)
--- a/src/createReducer.js
+++ b/src/createReducer.js
@@ -361,6 +361,15 @@ function createReducer<M, L>(structure: Structure<M, L>) {
         result = setIn(result, 'values', values)
         result = setIn(result, 'initial', values)
       }
+      var syncErrors = getIn(state, 'syncErrors');
+      if(syncErrors){
+        result = setIn(result, 'syncErrors', syncErrors);
+      }
+      var submitFailed = getIn(state, 'submitFailed');
+      if(submitFailed){
+        result = setIn(result, 'valid', false)
+        result = setIn(result, 'invalid', true)
+      }
       return result
     },
     [SUBMIT](state) {


src/createReducer.js
```diff
@@ -361,6 +361,11 @@ function createReducer<M, L>(structure: Structure<M, L>) {
         result = setIn(result, 'values', values)
         result = setIn(result, 'initial', values)
       }
+      var submitFailed = getIn(state, 'submitFailed');
```
timhwang21: If the user is resetting changes, why would they want failures to be propagated? What's the use case for this?

gustavohenke: I think you have misread that line, it's `getIn` not `setIn` 😄

timhwang21: Oh sorry my intention was to comment on the entire block. This seems to be checking if `submitFailed`, and if so, propagating `valid` state to the post-reset form state. My question is mainly, why would we want to propagate `invalid` after resetting?

gustavohenke: Yes, I don't agree with that.
Perhaps @slim25's idea is that every form start as invalid - which is not true, because this state depends on the validation functions.

slim25: Hi all, 
I noticed that when form is rendered and there is form validation errors so 'SyncErrors' field is present. In case if i understand correctly - form is invalid from start.

timacs12:
this one redux form issue so overide the sync error updation
----------------------------------------------------------------------------------------
import * as phenomicReducers from "phenomic/lib/redux/modules";
import { reducer as formReducer } from "redux-form";
import { combineReducers } from "redux";
import selectedApplyForm from "./SelectedApplyForm";
function callUpdateSyncErrors(state, action) {
  console.log("formReducer is called", state, action.type);
  // action.type === "@@redux-form/CHANGE" &&
  // action.meta.field === "select_your_profession1";
  if (state && action.type) {
    console.log("state>>>>>>>>>>>>", state);
    let flag = 0;
    let { registeredFields, syncErrors, values } = state;
    if (syncErrors && registeredFields) {
      let regFieldKeys = Object.keys(syncErrors);
      regFieldKeys.map(key => {
        flag = 1;
        console.log("key>>>>>>>>,", key, registeredFields);
        if (!registeredFields[key]) delete syncErrors[key];
      });
    }
    if (registeredFields) {
      flag = 1;
      syncErrors = {};
    }

    let new_synError = [];
    let new_registeredField = [];
    let difference = [];
    if (registeredFields)
      new_registeredField = Object.keys(registeredFields);
    if (syncErrors)
      new_synError = Object.keys(syncErrors);
    difference = new_registeredField.filter(x => new_synError.indexOf(x) === -1);
    difference.map((key) => {
      if (!values[key])
        if (syncErrors)
          syncErrors[key] = "Required *";
        else {
          syncErrors = {};
          syncErrors[key] = "Required *";
        }
    })


    if (difference.length !== 0 || flag === 1)
      return {
        ...state,
        syncErrors: {
          ...syncErrors
        }
      };
    return state;
  }
  else return { ...state };
}
const rootReducer = combineReducers({
  ...phenomicReducers,
  selectedApplyForm,
  form: formReducer.plugin({
    iBLeadApplyForm: (state, action) => {
      return callUpdateSyncErrors(state, action);
    }
    ,
    IotBootcampThankyouform: (state, action) => {
      return callUpdateSyncErrors(state, action);
    }
  })
});
// const rootReducer = combineReducers({
//   ...phenomicReducers,
//   form: formReducer,
//   selectedApplyForm
// })
export default rootReducer;

erikras:
I don't think I agree with this PR, and it isn't very well argued. @slim25, if you'd like to advocate for this, giving good reasons, I'd love to hear it, but as is, I'm going to have to close this PR.

PR #3308, opened 2017-08-10
Switched from isNan to typeof

morsecodist:
I noticed an issue when using `redux-form`. I was creating a form with field names that were IDs that were long numbers. This stores `fields` as an array rather than an object. The trouble with this is that the IDs were large enough to overflow the max array size.Cause crashing within `deleteIn` because there is a loop dependent on the array lengths and the array lengths for the arrays created from these IDs were incredibly high. I am not sure of the motivation of using `isNaN` here but switching to `typeof` resolves this issue.

### DIFF
--- a/src/structure/plain/setIn.js
+++ b/src/structure/plain/setIn.js
@@ -17,7 +17,7 @@ const setInWithPath = (
   const next = setInWithPath(firstState, value, path, pathIndex + 1)
 
   if (!state) {
-    if (isNaN(first)) {
+    if (typeof first !== 'number') {
       return { [first]: next }
     }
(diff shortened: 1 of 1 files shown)


chirag04:
can you also add a test?

morsecodist:
Sorry about the tests issue, I had trouble testing locally at first. My `initialValues` is an object and the problem persists. This is not as easy of a fix as I initially thought. I will close the PR but could you explain the motivation behind storing fields in an array in some cases? I want to try to make a more cohesive fix. At the very least large numbers should not be used for arrays like this.

PR #3321, opened 2017-08-15
Allow Fields to add arbitrary 'fieldData' to the redux state

bradywatkinson:


bradywatkinson:
hey, i wanted to use the wizard form features and server side validation but i did not have a good way to return to the page in which the errors occurred. I could replicate the redux-form state and add page identities in my own managed part of redux but i thought it would be more simple to add it to the registered fields themselves. If i forceUnregisterOnUnmount: false, then i can persist the page identity data until after my submission returns validation and simply find which page to redirect to

erikras:
I'm not convinced that this is something `redux-form` should do. You should use your own reducer, I think.

bradywatkinson:
If I use my own reducer to handle this case, then will essentially have a clone of the `redux-form` state which I will have to manage and keep in sync. By allowing meta data on the fields, `redux-form` can manage the lifecycle of the field but I can use that data to very effectively determine the page on which the field with the error is and redirect to it

PR #3333, opened 2017-08-17
Accept function as first argument in form value selector

delijah:
We run into a situation where we needed to retrieve a form field value of a form field value we didn't know, but was passed in by prop. To handle this in an appropriate way, we suggest to also accept functions as a first argument of the formValues selector which will resolve the according form field values.

(This code has not been tested, it should act as a proposal. I can try to run it and adjust tests if we see that there is a chance to have this in the master.)

### DIFF
--- a/src/createFormValues.js
+++ b/src/createFormValues.js
@@ -7,29 +7,9 @@ import type { Structure, ReactContext } from './types'
 import type { FormValuesInterface, PropPath } from './formValues.types.js.flow'
 
 const createValues = ({ getIn }: Structure<*, *>): FormValuesInterface => (
-  firstArg: string | Object,
+  firstArg: string | Object | Function,
   ...rest: string[]
 ) => {
-  let valuesMap: PropPath[]
-
-  if (typeof firstArg === 'string') {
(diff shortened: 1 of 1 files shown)


erikras:
Seems okay to me. Probably needs a test or two and the docs also need to be updated.

erikras:
@delijah, is this still needed?

delijah:
Yes it is. Sorry for delay. I will try to implement tests as soon as possible.

src/createFormValues.js
```diff
@@ -45,6 +25,25 @@ const createValues = ({ getIn }: Structure<*, *>): FormValuesInterface => (
...
+        } else {
+          const config: Object = resolvedFirstArg
+          valuesMap = Object.keys(config).map(k => ({
+            prop: k,
+            path: config[k]
+          }))
+        }
+        if (!valuesMap.length) {
+          throw new Error(
+            'formValues(): You must specify values to get as formValues(name1, name2, ...) or formValues({propName1: propPath1, ...})'
```
gustavohenke: Please add a function example in this error as well

delijah:
Reopened as #3432

PR #3338, opened 2017-08-18
Pass registeredFields to decorated form component.

### LINKED ISSUE #3337: registeredFields prop is not passed on to the decorated form component.

Bug report

The registeredFields property is not passed on to the form component.

The registeredFields property on the form component should contain the registered fields according to the [documentation](http://redux-form.com/7.0.3/docs/api/ReduxForm.md/#-code-registeredfields-array-code-)

### Sandbox Link

https://codesandbox.io/s/wk7m404kvk

### What's your environment?

redux-form v7.0.3

### Other information

Simply adding ```registeredFields: registeredFields,``` to reduxFormProps in createReduxForm.js solves the issue.

mtramm:
Fixes #3337

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -157,6 +157,7 @@ const describeReduxForm = (name, structure, combineReducers, expect) => {
         'invalid',
         'pristine',
         'pure',
+        'registeredFields',
         'reset',
         'submit',
         'submitFailed',
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -792,6 +792,7 @@ const createReduxForm = (structure: Structure<*, *>) => {
             initialValues,
             invalid,
             pristine,
+            registeredFields,
             reset,
             submitting,
             submitFailed,


erikras:
What is the need for this?

gustavohenke:
ping @mtramm

mtramm:
I'm currently working on an app in a domain with large business objects where only parts of the objects are editable by the user. What properties on the business object is editable varies depending on the value of other properties, so we have dynamic forms where fields are registered and unregistered depending on the value of other fields.

However, both from synchronous validation and from form submission we may (correctly) get errors with fields that are not editable by the user. However, field errors tied to a field that isn't currently registered in the form simply disappears. In order to be able to show these errors to the user we need to be able to determine what field errors won't be displayed on a field in the form, and to do that without having to duplicate the logic for registering/unregistering certain fields we need to be able to determine what fields are registered in any given form at any given time.

This pull request however, is just a simple fix that makes it a little easier to access the registered fields array. A 'getFormRegisteredFields' selector would be preferable, but I'm unsure what exactly the interface should look like.

erikras:
Okay. Can you update the docs, too?

mtramm:
This actually turns out to be ill conceived. As the fields are registered/unregistered when the form is being rendered (without triggering a rerender) the registeredFields prop will actually end up being out of date whenever the form data changes in a way that causes a field to be added to or removed from the form. The registeredFields prop will only accurately reflect the currently rendered fields whenever the form is rerendered with exactly the same fields as the last time it was rendered.

Issue #3350, opened 2017-08-23
Field level validation - get name of the currently validated field


Misiur:
This is a feature request
From quick glance at redux-form code, currently I see that the name of currently validated field name is discarded. I made the following change:
```diff
diff --git a/src/generateValidator.js b/src/generateValidator.js
index 8a7e503..35bd705 100644
--- a/src/generateValidator.js
+++ b/src/generateValidator.js
@@ -10,11 +10,12 @@ const getError = (
   value: any,
   values: Object,
   props: Object,
-  validators: any
+  validators: any,
+  name: string
 ) => {
   const array = toArray(validators)
   for (let i = 0; i < array.length; i++) {
-    const error = array[i](value, values, props)
+    const error = array[i](value, values, props, name)
     if (error) {
       return error
     }
@@ -28,7 +29,7 @@ const generateValidator = (
   let errors = {}
   Object.keys(validators).forEach(name => {
     const value = getIn(values, name)
-    const error = getError(value, values, props, validators[name])
+    const error = getError(value, values, props, validators[name], name)
     if (error) {
       errors = plain.setIn(errors, name, error)
     }
```

All mocha tests succeed, other tests are stuck on my computer. Shall I make a PR, or you are against changing API?

gustavohenke:
Done by #3364

erikras:
Published fix in [`v7.1.0`](https://github.com/erikras/redux-form/releases/tag/v7.1.0).

PR #3364, opened 2017-08-29
Add name parameter for field level validator

Misiur:
See https://github.com/erikras/redux-form/issues/3350


erikras:
Can you update [the docs](https://github.com/erikras/redux-form/blob/master/docs/api/Field.md), too?

Misiur:
There you go

erikras:

Issue #3381, opened 2017-09-05
5.3.2 Event data is object with multiple entries and has a key "value"


danikkks:
### Are you submitting a **bug report** or a **feature request**?

It`s a **bug report**

### What is the current behavior?

Widget return data to onChange funciton in format:

```
{
  includeUnknown: false,
  value: 123,
}
```

but getValue function in redux-form return only 123

### What is the expected behavior?

when value has more than 1 key return all object

### What's your environment?

"redux-form": "^5.3.2"

### Other information

https://www.evernote.com/l/AeJriOylJrhPF4K7a2wiolhL0jMAKbbkq3Y
https://www.evernote.com/l/AeJuYqM43DBJC5TDkpIQYVySr_0PTMcN1PE

gustavohenke:
Will close as the PR that fixed it was merged

PR #3382, opened 2017-09-05
fix: getValue not return object when one of keys is "value"

danikkks:
solution for https://github.com/erikras/redux-form/issues/3381


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "redux-form",
-  "version": "5.3.6",
+  "version": "5.3.8",
```
gustavohenke: This must be removed

.npmrc
```diff
@@ -0,0 +1 @@
+registry=http://10.150.10.130:4873/
```
gustavohenke: This must be removed

erikras:

PR #3399, opened 2017-09-10
Jest codemods part 1

skovhus:
This is the first part of migrating to Jest:
- remove commented out tests (Jest doesn't like these)
- fix tests that contains no assertions (Jest found these)

The rest of the migration to Jest should be done by [`jest-codemods`](https://github.com/skovhus/jest-codemods), I'm working on that.


skovhus:
Replaced by https://github.com/erikras/redux-form/pull/3400

PR #3409, opened 2017-09-12
FieldArray swap does not rerender

krtcom:
I Added test for FieldArray fields.swap method because it looked that it was not working. 
After running tests this one indeed failed so I updated shouldComponentUpdate method of ConnectedFieldArray so that it returns true if some elements in array were swapped.

Can someone confirm if the test is correct? Possibly change it if needed. I think some expect calls might be unnecessary.
Is the shouldComponentUpdate solution acceptable?

### DIFF
--- a/src/ConnectedFieldArray.js
+++ b/src/ConnectedFieldArray.js
@@ -37,8 +37,11 @@ const createConnectedFieldArray = (structure: Structure<*, *>) => {
       const nextValue = nextProps.value
 
       if (thisValue && nextValue) {
+          let nextValueItemsSame = nextValue.every(val => thisValue.includes(val));
+          let nextValueItemsOrderChanged = nextValue.some((val, index) => val !== thisValue[index]);
         if (
           thisValue.length !== nextValue.length ||
+          (nextValueItemsSame && nextValueItemsOrderChanged) ||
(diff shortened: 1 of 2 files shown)


erikras:
Fantastic sleuthing! 👍

erikras:

src/ConnectedFieldArray.js
```diff
@@ -37,8 +37,11 @@ const createConnectedFieldArray = (structure: Structure<*, *>) => {
       const nextValue = nextProps.value
 
       if (thisValue && nextValue) {
+          let nextValueItemsSame = nextValue.every(val => thisValue.includes(val));
+          let nextValueItemsOrderChanged = nextValue.some((val, index) => val !== thisValue[index]);
         if (
           thisValue.length !== nextValue.length ||
+          (nextValueItemsSame && nextValueItemsOrderChanged) ||
```
derekcannon: @erikras Is there a reason this is an `&&`? If the `nextValueItemsSame` is `false`, shouldn't we always rerender? In this case, it'll only rerender if the items have changed AND they've rearranged.

In our use-case, we're deleting an item from a field list array. We add `destroy: true` to the value object and want to rerender the list with the "destroyed" item hidden. The list does not rerender because of this logic.

PR #3427, opened 2017-09-19
Deep equal the value of radio input form fields to detect changes

delijah:
Intro:
To detect changes on form field values, redux form is using deepEqual. This means, for example, that we can use (complex) object values for text inputs (which is great) - or simple values.

Problem:
Unfortunately the usage of (complex) object values does not work for radio inputs currently.

Solution:
This PR enables radio inputs to handle (complex) object values as well.


delijah:
Got replaced by #3430

PR #3430, opened 2017-09-20
Use deep equal to compare radio input values

delijah:
Intro:
To detect changes on form field values, redux form is using deepEqual. This means, for example, that we can use (complex) object values for text inputs (which is great), but also simple values.

Problem:
Unfortunately the usage of (complex) object values does not work for radio inputs currently (radio input values are compared with ===).

Solution:
This PR enables radio inputs to handle (complex) object values as well by using deepEqual for radio input value comparison.


delijah:
Any news on this? Do you see a possibility to merge this into master soon?

delijah:
Amazing! 🎉  Thank you for merging 👍

erikras:

Issue #3434, opened 2017-09-21
Unnecessary render


devhak2:
### Are you submitting a **bug report** or a **feature request**?
In my opinion, it's a bug

### What is the current behavior?
In the code that i wrote a link below, when i write down a letter into the input box, component is re-rendered. Also, that's re-rendered if i remove the first letter.
What I mean is that when you only hit the first letter or erase the first letter, it happen.

### What is the expected behavior?
Regardless of whether what i write is the first character or not, it should be rendered or not rendered.

### Sandbox Link
https://codesandbox.io/s/8x5186lyzj

### What's your environment?
In chrome, you can see the issue at the above link

### Other information
Nope

erikras:
I disagree. `dirty`/`pristine` are very important props to notify the entire form of. e.g. for disabling the submit button when the form is pristine.

PR #3436, opened 2017-09-21
Just add 'dirty' prop into propsToNotUpdateFor.

devhak2:
This is for [issue #3434](https://github.com/erikras/redux-form/issues/3434) that i wrote.

Please check this.


devhak2:
I think adding dirty prop is not a big deal.

erikras:
https://github.com/erikras/redux-form/issues/3434#issuecomment-331131971

Issue #2600, opened 2017-02-16
[BUG] When passing input.onChange(false) component dont rerender


GautierT:
Hello,
Found a bug with `input.onChange(false)` : it doesn't rerender.

I create a Custom Field who create a group of buttons to mirror the behaviour of radio true/false but with buttons : 
```
<Field
    component={DefaultGroupedButton}
    name="user.statut"
    required
    label="Question ?"
    elements={[
        {
            label: 'Yes',
            value: true
        },
        {
            label: 'No',
            value: false
        }
    ]}/>
``` 

and in `DefaultGroupedButton` i map each elements with a button and it's value.
When i click on one of the button i call onChange with the value in element (can be `true` or `false`) : 

```
const onChange = (e, data) => {
    // if i click again in the same button i want to remove the value so i pass null
    if (input.value === data.value) {
        console.log('changing to null');
        input.onChange(null)
    }
    else {
        console.log('changing to ' , data.value);
        input.onChange(data.value);
    }
};
```

It's working great with `true`. 
**But with `false` the component don't rerender.** I see **it change in redux but their is no rerender** so i can't apply display condition in other element.

gustavohenke:
It seems like this was yet another issue closed by the [v7.0.0 release](https://github.com/erikras/redux-form/releases/tag/v7.0.0):

> `undefined` and `false` are now treated as NOT equal. This will affect your dirty/pristine state, as well as whether or not a field is rerendered when the value goes from `undefined` to `false` (previously it did not)

@GautierT, if this is still not the case, let us know.

jonathonwalz:
@gustavohenke both this and #2978 are *not* solved with the 7.0 release. The issue here is that the comparison is considering null/''/false to all be the same. So if the initial value is `null` (not `undefined`), setting the value to `false` isn't detected as a change.

I've worked around this by using a string constant for false in the form, but it was definitely unexpected behavior to me. (My use case was a similar to this, a yes/no/no-response style question)

shodanuk:
Confirmed NOT fixed in 7.0. Still seeing this with 7.04

viniciusdacal:
@gustavohenke @erikras I just opened a PR fixing that. Please, can you guys review it?

erikras:
Published fix in [`v7.1.0`](https://github.com/erikras/redux-form/releases/tag/v7.1.0).

PR #3462, opened 2017-09-28
Fix deepEqual comparison to tell the diff between null and false

viniciusdacal:
This PR intends to fix the issue reported at #2600 
The issue is closed but the bug remains.

Wrote a unit test that ensures that bug was present before the fix.


erikras:

Issue #3449, opened 2017-09-24
Feature Request: Use separate criteria for shouldValidate and shouldWarn


kennykaye:
**Current Behavior**:
When specifying `props` for a form, the optional `shouldValidate` function is provided which allows the user to determine whether or not validation is performed on the form. When the user provides a `validate` and `warn` function, the `shouldValidate` function is run to determine whether to perform both.

**Desired Behavior**:
`props` accepts `shouldValidate` and `shouldWarn` functions. The behavior should be as follows:

* `shouldValidate` is called to determine whether or not the `validate` should be run.
* `shouldWarn` is called to determine whether or not the `warn` function should be run.
* `shouldWarn` does not impact `validation` and `shouldValidate` does not impart `warn`.

**Risks**:
There may be implementations that depend on the current behavior of `shouldValidate`. Since this is the case, it is important backward-compatibility is maintained. To achieve this, `shouldValidate` should be deprecated in favor of two new functions, which implement the desired behavior. 

**Final Thoughts**
I am happy to submit a PR for this myself if your team sees value in this feature.

kennykaye:
I've added an initial implementation of this feature in https://github.com/erikras/redux-form/pull/3464

erikras:
Published fix in [`v7.1.0`](https://github.com/erikras/redux-form/releases/tag/v7.1.0).

PR #3464, opened 2017-10-01
Separate error and warnings in validation override

kennykaye:
This is an initial implementation for the feature requested in https://github.com/erikras/redux-form/issues/3449

### DIFF
--- a/src/__tests__/defaultShouldError.spec.js
+++ b/src/__tests__/defaultShouldError.spec.js
@@ -1,0 +1,97 @@
+import expect from 'expect'
+import plain from '../structure/plain'
+import immutable from '../structure/immutable'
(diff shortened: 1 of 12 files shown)


erikras:
This looks great!

I wonder how we should go about deprecating the old `shouldValidate()`? Maybe some sort of only-on-dev "this will go away in the next major version" message, like React itself uses?

kennykaye:
@erikras That's a great idea. 

Is there some sort of build flag or configuration that we can leverage for the warning? For the error message, I'm thinking:

> shouldValidate() is deprecated and will be removed in v8.0.0. Use shouldWarn() or shouldError() instead.

I will update the documentation for `shouldValidate` and add new documentation for `shouldError` and `shouldWarn`.

PR #3479, opened 2017-10-05
Changed immutable/deepEqual to use ===

### LINKED ISSUE #3357: structure/immutable/deepEqual should use === (or better?)
Bug Report

When using a custom field that coerces strings into numbers, immutable/deepEqual will return true when comparing string "1." vs numeric 1.  This is because it uses ==, which will coerce "1." to 1 before doing the comparison.  This causes problems with onChange handlers, where redux-form doesn't think anything changed.

immutable/deepEqual should at the minimum be using === to do comparisons.  Theoretically, it seems like immutable/deepEqual should be satisfied by merely doing a reference comparison between the two values.  That's what using immutablejs is supposed to buy us anyway.

### Sandbox Link
See: https://codesandbox.io/s/8x3kox4949
Enter "1." and then try to backspace

### What's your environment?
redux-form: 6.7.0
Chrome: 61.0.3163.59
OS: macOS 10.12.5
Node: v6.10.2

This problem has existed since commit bd4cd9e

### Other information
The plain version is doing the right thing.

erikras:
Fixes #3357.

Not the first time ignoring an eslint error created a bug. 🤦‍♂️

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -3571,6 +3571,47 @@ const describeReduxForm = (name, structure, combineReducers, setup) => {
       })
     })
 
+    it('should call form-level onChange when values change using ===', () => {
+      const store = makeStore({})
+      const renderFoo = jest.fn(props => <input {...props.input} />)
+      const onChange = jest.fn()
+
+      class Form extends Component {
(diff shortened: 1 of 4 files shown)


erikras:

PR #3480, opened 2017-10-05
React16: Prevented deepEqual checks on props.children in shouldComponentUpdate()

### LINKED ISSUE #3461: Performance issue (crash) in large apps when using Field with children in React v16
bug

Reproduction:
 - Use redux-form with React v16 in a large application (a page with a lot of different nested React components)
 - Add a field with children:  
```
<Field component={SomeCustomInput} ...>
   <SomeChild \>
   <SomeChild \>
   <SomeChild \>
</Field>
```
 - The page becomes so slow that the browser becomes unresponsive and crashes

What seems to happen is that `<ConnectedField />` will execute `shouldComponentUpdate`, which will call lodash `isEqual` on its props. The `children` prop in react v16 contains `FiberNode` objects, which seem to have references to a lot (if not all) other `FiberNode` on the page. Because of this, lodash `isEqual` will take a very very long time recursively traversing through these objects. 

### Sandbox Link
Unfortunately I haven't been able to reproduce this in stand-alone example. If I create a `<Field>` with children in one of the examples of this repo, I can see lodash `baseIsEqualDeep` being called with very long call stacks. However, it is not enough to visually slow down the page.

### What's your environment?
All browsers
React v16
redux-form v7.0.4

erikras:
Fixes #3461.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -63,14 +63,18 @@ const createConnectedField = (structure: Structure<*, *>) => {
     shouldComponentUpdate(nextProps: Props) {
       const nextPropsKeys = Object.keys(nextProps)
       const thisPropsKeys = Object.keys(this.props)
+      // if we have children, we MUST update in React 16
+      // https://twitter.com/erikras/status/915866544558788608
       return (
-        nextPropsKeys.length !== thisPropsKeys.length ||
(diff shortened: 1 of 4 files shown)


erikras:

Issue #3437, opened 2017-09-21
Field value after initialize works strange


meuwka:
### Are you submitting a **bug report** or a **feature request**?
Probably a bug

### Description
I have component with 3 states: 'yes', 'no', 'unknown'.
I use this way to change value to 'unknown' :
```
input.onChange('')
```
And it works good. 
![screenshot from 2017-09-21 14-52-44](https://user-images.githubusercontent.com/2752193/30694418-9553aee0-9edc-11e7-80c2-c85faa4a5d7a.png)

But after form `initialize` plus `input.onChange('')` value is set to `''`. Also `onChange` stop work for 'unknown' -> 'no' and vice versa

![screenshot from 2017-09-21 14-56-08](https://user-images.githubusercontent.com/2752193/30694523-09a526f2-9edd-11e7-9059-8493ba6a7932.png)

### What is the current behavior?
After initialize `onChange('')` set value `''`

### What is the expected behavior?
Delete value of field after `onChange('')` like it works without `initialize`.
I expect that there should be one behavior with initialization and without it

### What's your environment?
redux-form@7.0.4

### Other information
You can view code here https://github.com/meuwka/redux-form-issue-test
It works good on another project with redux-form@5.3.6

erikras:
Fixed by #3482.

erikras:
Published fix in [`v7.1.0`](https://github.com/erikras/redux-form/releases/tag/v7.1.0).

Issue #3162, opened 2017-07-08
Checkbox unchecked state has empty string value when initialValues is set


lightwave:
### Are you submitting a **bug report** or a **feature request**?

This is a bug report.

### What is the current behavior?

When initialValues is set for a form's checkbox field, the uncheck state will have an empty string value. You can see the example at https://codesandbox.io/s/mQovglORR

Uncheck the "employed" checkbox in the sandbox demo and you'd see the value become an empty string.

### What is the expected behavior?

The expected behavior is to have false as the value when the check box is uncheck. It works when no initialValues is set.

### Sandbox Link

https://codesandbox.io/s/mQovglORR

### What's your environment?

redux-form 6.7.0 and 6.8.0 are both broken.

klk1010:
I'm seeing the same thing when I copy over the [example from the docs](http://redux-form.com/6.8.0/examples/simple/) - defaults to v6.6.3. It seems to work as expected in v6.6.3, but not in v6.7.0 or v6.8.0.

kinyat:
I also noticed this issue.
The current work around is the use parse to do `parse={value => !!value}`.

davidgomes:
I can confirm to be able to reproduce this issue.

danielrob:
This is the direct result of #2857.

danielrob:
[![Edit Redux Form - Simple Example](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/kZvpkml0X) - 7.0.3 sandbox.

lightwave:
7.0.3 is still broken.

pmg103:
7.0.4 is still broken

erikras:
Fixed by #3482.

erikras:
Published fix in [`v7.1.0`](https://github.com/erikras/redux-form/releases/tag/v7.1.0).

Issue #2922, opened 2017-05-12
Checkbox loses attribute "value" 


Slavenin:
I create a checkbox with stateless function and set value. If checkbox changed - my value set to true or false. If i uncheck checkbox why checkbox exists in state? In simple html form if checkbox is not checked by form submit he not submited to server. 

In previous version value in state be true or false. Now - true or undefined. But not value that i set in start

Sandbox link https://jsfiddle.net/Slavenin/o31L5ns2/2/

danielmackey:
Also experiencing this too. This looks like regression in 6.7.0. 

If you inspect the Employed checkbox input between these two versions, you'll see the error too. Or just look at the values output between the two. 

http://redux-form.com/6.7.0/examples/simple/ 
Unchecked: `<input value />` and Checked: `<input value="true" />`

http://redux-form.com/6.6.3/examples/simple/
Unchecked: `<input value="false" />` and Checked: `<input value="true" />`

westhood:
Have a same issue in 6.7.0

Slavenin:
Why value may by only `true` ot `false`? In my case it can have any value - 1,2,3 or uuid! Why change attribute `value`? Checkbox have `checked` attribute.

evgeny-myasishchev:
Same here, unchecked checkbox yields empty value instead of false. Rolled back to 6.6.3 and it works there.

Slavenin:
I am not need `false`! I need my setted value 1,2,3 or any other!

mvirbicianskas:
having the same issue. Locking out to 6.6.3 fixes it. Would be great to see how we could work around this. Because now if checkbox is checked and then unchecked its payload is "" (empty string) and is removed from form's state :( 
@erikras any updates on that? or is our approach wrong altogether? :(
personally what I'm using is just fields with unique names which expects value true/false which in 6.7.0 now is true/""

stunaz:
Experimented the same issue - the onBlur method set it empty

mvirbicianskas:
also, i'm curious why the decision was made for checkbox field to be removed from form state altogether? It's a checkbox and it cannot be undefined, because this is a field with value, either checked or unchecked, IMO this cannot be checked or non-existant_field

Slavenin:
Can anybody tell me why the value is converted to a boolean type?
```html
<form>
    <input type="checkbox" name="list[]" value="1">
    <input type="checkbox" name="list[]" value="2">
    <input type="checkbox" name="list[]" value="3">
[...]

PR #3482, opened 2017-10-05
Fixed broken checkbox behavior

### LINKED ISSUE #2857: Checkbox value breaking pristine state

erikras:
Fixes #2857, #2922, #3162, #3437.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -578,6 +578,49 @@ const describeReduxForm = (name, structure, combineReducers, setup) => {
       expect(inputRender).toHaveBeenCalledTimes(1)
     })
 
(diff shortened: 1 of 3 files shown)


erikras:

PR #3486, opened 2017-10-06
Fixed plugin bug demonstrated in #2935

erikras:
Closes #2935.


erikras:

Issue #3493, opened 2017-10-09
Immutable 4 is breaking change - release React 16 version that still works with immutable 3


ntucker:
### Are you submitting a **bug report** or a **feature request**?

Following semver, changing immutable to 4 should probably be a major release, as 3 would no longer be supported. However, I think many would like React 16 compatibility before they upgrade to the not-yet-released immutable 4.

### What is the current behavior?

7.1 requires Immutable 4 and 3 does not work

### What is the expected behavior?
Version 7 should not require a breaking library bump of immutable.

erikras:
How might this library detect what version of Immutable you are using?

erikras:
What, exactly, is broken when using `redux-form@7.1.0` with `immutable@3`?

The immutable example seems to be working normally.

hsz:
@erikras I confirm this issue.

```
  TypeError: (0 , _immutable.isCollection) is not a function
  
  - index.js:42 getIn
    [my-project]/[redux-form]/lib/structure/immutable/index.js:42:40
  
  - createFormValueSelector.js:22 nonNullGetFormState
    [my-project]/[redux-form]/lib/createFormValueSelector.js:22:14
  
  - createFormValueSelector.js:31 
    [my-project]/[redux-form]/lib/createFormValueSelector.js:31:13
  
  - index.js?:1 Function.mapStateToProps [as mapToProps]
    webpack:///./src/components/MyComponent/index.js?:1:5775
  
  - wrapMapToProps.js:54 mapToPropsProxy
    [my-project]/[react-redux]/lib/connect/wrapMapToProps.js:54:92
  
  - wrapMapToProps.js:63 Function.detectFactoryAndVerify
    [my-project]/[react-redux]/lib/connect/wrapMapToProps.js:63:19
  
  - wrapMapToProps.js:54 mapToPropsProxy
    [my-project]/[react-redux]/lib/connect/wrapMapToProps.js:54:46
  
  - selectorFactory.js:37 handleFirstCall
    [my-project]/[react-redux]/lib/connect/selectorFactory.js:37:18
  
  - selectorFactory.js:85 pureFinalPropsSelector
    [my-project]/[react-redux]/lib/connect/selectorFactory.js:85:81
  
  - connectAdvanced.js:43 Object.runComponentSelector [as run]
    [my-project]/[react-redux]/lib/components/connectAdvanced.js:43:25
  
  - connectAdvanced.js:195 Connect.initSelector
[...]

PR #3460, opened 2017-09-28
Skip adding preventDefault to onChange events in React Native

jkomyno:
# Bug
When using onChange in React Native, the users are faced with a couple possible error messages, as seen in #3153 and #3238.
This is due to the fact that it doesn't seem to be possible to properly extends React Native's event ( https://github.com/facebook/react-native/issues/5507). My proposal is to skip adding .preventDefault to TextInput's onChange's event.

## Tests
No test was broken

## Eslint
Linting was respected

Closes #3153, closes #3238

PR #3496, opened 2017-10-09
Extends PR #3460, also locks immutable version

jkomyno:
The main purpose of this PR is to add the same logic that PR #3460 (which was merged) added to `onChange` also to `onBlur` and `onFocus`.

I've also edited a dev-dependecy line of package.json:

```diff
-"immutable": "^4.0.0-rc.3",
+"immutable": "4.0.0-rc.3",
``` 

This was necessary to prevent the test suite to fail miserably with 21 errors (all of them are `immutable`-related). This isn't a solution to #3493, but it's needed to mark the fact that versions `4.0.0-rc.>3` of `immutable` aren't supported as of now. I hope this doesn't add to much noise to the PR @erikras.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "flow-bin": "^0.56.0",
     "flux-standard-action": "^1.0.0",
     "get-typed": "^1.0.0-beta.2",
-    "immutable": "^4.0.0-rc.3",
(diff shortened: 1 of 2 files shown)


jkomyno:
Thanks for merging!

PR #3502, opened 2017-10-10
Changed meta prop type to any

dobryanskyy:
In case errors are passed as something different than strings(e.g.: `<FormattedMessage />` when using react-intl)


erikras:
Oops. My mistake.

erikras:

Issue #3513, opened 2017-10-12
Values not being passed to reduxForm decorated component


dobryanskyy:
### Are you submitting a **bug report** or a **feature request**?

bug report

### What is the current behavior?

values is not passed to decorated component

### What is the expected behavior?
According to https://redux-form.com/7.1.1/docs/api/reduxform.md/#instance-api values should be passed to decorated component

gustavohenke:
Hi @dobryanskyy, you understood the instance API incorrectly.
Please see this amazing explanation: https://github.com/erikras/redux-form/issues/2101#issuecomment-261253199

dobryanskyy:
Thanks for explanation.

PR #3514, opened 2017-10-12
Added values to props which are passed to wrapped component

dobryanskyy:
Fixing https://github.com/erikras/redux-form/issues/3513


gustavohenke:
See related comment in #3513

dobryanskyy:
Thanks for the explanation

PR #3535, opened 2017-10-18
Replaced includes with indexOf

### LINKED ISSUE #3533: ConnectedFieldArray.js uses "includes" which is unsupported in IE11

IE11 blows up when visiting a form that uses a FieldArray

IE11 shouldn't blow up

### What's your environment?

Form v7.1.1
IE11

### Other information

`TypeError: Object doesn't support property or method 'includes'`

andokai:
Resolves #3533


kraaas:
👍 (this bug gave us a lot of trouble)

erikras:

PR #3556, opened 2017-10-23
update hoist-non-react-statics version for v5

MhMadHamster:
Upgraded single package to the latest version
should not have any breaking changes


erikras:

Issue #3055, opened 2017-06-13
Feature proposal: `clear` action creator


timhwang21:
### Are you submitting a **bug report** or a **feature request**?

Feature request

A common pattern I'm running into for dependent fields (e.g. field B's value depends on field A having a value) is that I want to clear field B when field A changes.

So I do something like:
```javascript
// Assume field B, C, and D are conditional and A having a value
handleFieldAChange(event, value) {
  const { change } = this.props;

  if (!value) { // if field a was emptied
    change('b', null);
    change('c', null);
    change('d', null);
    untouch('b', 'c', 'd');
  }
}
```

`change` doesn't support passing in multiple fields currently. In my case, I know that the value for each changed field will be `null`, so an action creator `clear()` (or some better name) with the same signature as `untouch()` would be very useful:
```javascript
handleFieldAChange(event, value) {
  const { clear } = this.props;

  if (!value) {
    clear('b', 'c', 'd');
  }
// Behavior: args.forEach(arg => { blur(arg); change(arg, null); })
```

*Alternative / additional suggestion:* Change the argument order of `change` to `(form:string, value:any, ...fields:string)`, to bulk change values. The downside is that similar action creators will need to be changed to be consistent with the signature.

Is there a better existing way that I'm missing? If not, would there be interest in this functionality? If so I can take a crack at it over the weekend. Thanks!

danielrob:
Seems sensible to me, I'd say go for it!

danielrob:
Note as well the linked issue.. maybe there are other cases for having bulk setting of formValues.

timhwang21:
Hm, I just saw the issue, so maybe a `changeMany` or something would be more general-purpose. Thanks! (By the way, I'm not sure what the best way to proceed on this is -- I've filed another issue #3295)

danielrob:
`setFormValues` is another name suggestion ;).

danielrob:
Consider also #2224

timhwang21:
@danielrob So I'm running into an issue in the implementation of a bulk change action, and was hoping one of the maintainers could chime in... (cc @erikras)

The issue: the function signature for actions that change multiple fields is `...fields`, taking rest arguments instead of an array for convenience/ nicer syntax, I guess. This means that the list of fields must be the last arguments passed.

[...]

Issue #2224, opened 2016-12-06
undo/reset for a field


jole78:
Couldn't find it in the docs or other places.

It would be a nice enhancement to be able to reset/undo just one field and restore the value to the initial value it had when it was mounted.

danielrob:
(with signature `reset(...fields)`)

erikras:
Fix published in [`v7.2.0`](https://github.com/erikras/redux-form/releases/tag/v7.2.0).

PR #3576, opened 2017-10-31
Add cleanFields action

dvdmgl:
`clearFields(form:String, keepTouched: boolean, persistentSubmitErrors: boolean, ...fields:String)`

Cleans fields values for all the fields passed in.

> If the `keepTouched` parameter is `true`, the values of currently touched fields will be retained
> If the `persistentSubmitErrors` parameter is `true`, the values of currently submit errors fields will be retained

Closes: #3055, Closes: #2224

### DIFF
--- a/docs/api/ActionCreators.md
+++ b/docs/api/ActionCreators.md
@@ -13,7 +13,7 @@ form, and, in the case of field-specific actions such as `CHANGE` or `BLUR`, the
 
 ### `arrayMove(form:String, field:String, from:Number, to:Number)`
 
-> Moves an item from one index in the array to another. In effect, it performs a remove and an 
+> Moves an item from one index in the array to another. In effect, it performs a remove and an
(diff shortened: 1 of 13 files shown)


erikras:
What a perfect, perfect PR! Docs, flow, tests, everything. 👍 👍 👍 

Thanks!

erikras:

MrBlenny:
It looks like clearFields doesn't clear the store value or errors when destroyOnUnmount:false. Is this expected?

Ripppe:
Hi.

Is this feature suppose to work with field arrays (i.e. can you give the name you give to the FieldArray for this function)? If yes, I have a bug report on standby. If not, is there any feasible workaround to get FieldArray's value completely cleared from store rather than leaving it with empty array (the case when you use removeAll)? :)

BR, Ripppe

PR #3586, opened 2017-11-06
Update non dirty values also for non registered field values

ms88privat:
I struggled a lot after working with dynamic fields. I figured out, that the fields have to be registered in order to get the (initial) values updated if keepDirty.

The problem is, that I got fields which will register after the initialising happens. Just initialising again doesn't solve the problem, because the first initialise did mutate the initialValues, which will be  dirty in the second case and will not update because of this.

After this change, the logic for updating pristine fields will now always apply, no matter if the fields a registered or not. This may be breaking to some apps? But I think this is more expected then the other case?

Does this make any sense to you @erikras ?

Thanks

### DIFF
--- a/docs/api/ActionCreators.md
+++ b/docs/api/ActionCreators.md
@@ -81,7 +81,7 @@ insert, so the item already at the `to` position will be bumped to a higher inde
 
 > Marks the given field as `active` and `visited`.
 
-### `initialize(form:String, data:Object, [keepDirty:boolean], [options:{keepDirty:boolean, keepSubmitSucceeded:boolean}])`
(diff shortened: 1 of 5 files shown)


gustavohenke:
> This may be breaking to some apps? But I think this is more expected then the other case?

Yes, it definitely will be a breaking change. What would require a v8 release...

ms88privat:
@gustavohenke Thanks for the early feedback. What do you think about this change in general?

Maybe I will just do it as a optional configuration? So it will be non breaking?

erikras:
Non-breaking > Breaking, especially for such an edge case.

What on earth would you call the config flag? 🤔

ms88privat:
Really difficult to find a good name....
Please have a look at my changes @erikras 

Thank you :)

ms88privat:
Any news here :)?

Issue #3529, opened 2017-10-16
`array` convenience prop not getting added to `propNamespace`


jeremyosborne-xactly:
### Are you submitting a **bug report** or a **feature request**?
bug report

### What is the current behavior?
the `array` property that provides bound actions for working with FieldArrays does not get added to the`propNamespace` object if declared.

### What is the expected behavior?
If I use `propNamespace: 'reduxForm',` in my `reduxForm` decorator, I should be able to access `this.props.reduxForm.array` like i can with other correctly namespaced props, but instead I need to use `this.props.array` (e.g. `this.props.array.push({...some data});`).

### Sandbox Link
see: https://codesandbox.io/s/045x946yj0 in the SimpleForm.js file

### What's your environment?
Sandbox uses 7.0.4, app uses 6.7.0. Behavior seems same in sandbox version and my app.

gustavohenke:
Thanks for the report!

erikras:
This should be an easy PR for newbies.

vnctaing:
I've been trying to investigate and open my first PR, but I haven't been lucky so far. I've looked into [createReduxForm](https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L842-L846) file, it seems correctly passing the propNameSpace and I also looked into the passing [test](https://github.com/erikras/redux-form/blob/master/src/__tests__/reduxForm.spec.js#L426-L430) that looks also fine to me, does anyone has thought on where the problem could come from ?

gustavohenke:
@vnctaing that test is not checking whether `array` is there.
If you add a expectation for this, it will likely break the test.

PR #3614, opened 2017-11-15
Add `array` to propNamespace

vnctaing:
This should fix https://github.com/erikras/redux-form/issues/3529


gustavohenke:
Perfect - and you did it using TDD ❤️

Issue #3435, opened 2017-09-21
Form doesn't reinitialize when replaced


eliseumds:
### Are you submitting a **bug report** or a **feature request**?

Bug report. I'm seeing React lifecycle methods being called in a different order with React 16 which breaks the forms (leaving them destroyed). 

### Expected order

* componentWillUnmount
* componentWillMount
* componentDidMount

### Actual order

* componentWillMount
* componentWillUnmount
* componentDidMount

It looks like it's an [expected behaviour](https://twitter.com/dan_abramov/status/843822280132190208) but it leads to issues because the forms don't reinitialize.

kjanoudi:
This also is the case after a form submission using a pure component with `initialValues`. After the form is submitted, and the new data is passed to the wrapper component via props, the redux form will `DESTROY` after registering all the new fields. In react 15, it would `DESTROY`, `INITIALIZE`, then register all the fields.

### React 15

![](https://files.slack.com/files-pri/T04CXPZ4L-F7A6RBRJ8/pasted_image_at_2017_09_28_04_41_am.png)


### React 16

![](https://files.slack.com/files-pri/T04CXPZ4L-F79FDE6PJ/pasted_image_at_2017_09_28_04_40_am.png)

eliseumds:
@kjanoudi same thing here :( It's good to know that someone else is having the same problems. This was such a massive change from React that breaks so much stuff and not only with *redux-form*:

blord-fullscreen:
React 16 compatilbility is crucial. Would love to see this made a high priority.

kjanoudi:
Shocked this isn't getting more attention. Guess I'll be waiting to upgrade to React 16

erikras:
Is it obvious to anyone how to solve this? It isn't to me.

`redux-form` doesn't behave all that well under dev hot reloading in my experience. I usually turn `destroyOnUnmount: false` for the duration of the development of my form. Does that make any difference here?

kjanoudi:
@erikras I'm less concerned with the behavior or hot reloading as opposed to in production. I'm experiencing the scenario I outlined above on a production build of my app. After saving the form and remaining on the same screen, previously the form would not clear. Presently, it clears.

eliseumds:
@erikras sorry for the confusion. I edited the title and description. It happens every time the form instance gets replaced.

eliseumds:
[...]

PR #3615, opened 2017-11-15
move initialization to componentDidMount

brucewpaul:
React 16 ensures a certain order of lifcecycles when replacing a
component which causes redux-form to destroy the form, breaking
it. This change, when used with `enableReinitialize`, causes the
form to reinitialize correctly, preserving the form.

this pr address https://github.com/erikras/redux-form/issues/3435

this is probably going to fail a lot of tests, but want to start the 
conversation as to what is needed for redux-form to work with
React 16

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -526,6 +526,18 @@ const createReduxForm = (structure: Structure<*, *>) => {
           )
         }
 
+        componentDidMount() {
+          if (!isHotReloading()) {
+            this.initIfNeeded()
+            this.validateIfNeeded()
+            this.warnIfNeeded()
+          }
+          invariant(
+            this.props.shouldValidate,
+            'shouldValidate() is deprecated and will be removed in v8.0.0. Use shouldWarn() or shouldError() instead.'
+          )
+        }
+
         componentWillUnmount() {
           const { destroyOnUnmount, destroy } = this.props
           if (destroyOnUnmount && !isHotReloading()) {


brucewpaul:
cc @erikras @gustavohenke

brucewpaul:
actually, just copying the same functionality form `componentWIllMount` to `componentDidMount` passes all the test and is working in our app. Although this solution of course comes with the added perf costs which isn't ideal at all...

gustavohenke:
Well, actually, quite a few tests are breaking!
https://travis-ci.org/erikras/redux-form/jobs/302635057#L6766

brucewpaul:
:) @gustavohenke this pr doesn't include the most recent note I wrote. Locally, all the tests are passing. I will update this pr to be that version since all the tests pass

brucewpaul:
@gustavohenke updated and the tests just passed locally

stevenmusumeche:
is it likely that this PR will be merged? I have a workaround using setTimeout that I'd rather not implement if I don't have to.

brucewpaul:
@stevenmusumeche link to pr?

stevenmusumeche:
In my case, I am navigating AWAY from the form, and when I go back to it, it already has "submitSucceeded" set to true from the previous submission.  So, I was able to do a setTimeout and call `destroy` on the form after 1 second.  Dirty hack, but it works for my use case.

brucewpaul:
@stevenmusumeche does this solution work for you?

stevenmusumeche:
This PR? I have not run my code with it, sorry.

erikras:
Sorry guys. I was sick last week.

Meh, I don't care about "perf costs" when hot reloading. Looks good.

brucewpaul:
@erikras I was experiencing this in production, not just hot reloading, which is why I brought up the perf cost of this change.

stevenmusumeche:
Yes, same here, it's not just a hot reloading issue.

brucewpaul:
@erikras would you be able to do another release that includes this?

erikras:

PR #3619, opened 2017-11-17
Dont delete initial initialValues structure on withCleanUp

### LINKED ISSUE #2597: Impossible to clear last field in a form when initialValues is an empty object and the form was submitted

Given that the form is being passed an empty object to `initialValues` prop, and then the form has been submitted, after clearing last field of this form (for e.g. with a backspace button) it will still return last field's value in `onSubmit` handler.

An example is here:

[https://jsfiddle.net/piteer/o4wjnjyz/2/](https://jsfiddle.net/piteer/o4wjnjyz/2/)

You need to fill **one** field and then press submit, then clear the field and again press submit. This will return last submitted value.

Also worth mentioning: If you fill any other field and submit a form again, last previous field will again be empty as it should be.

Value send to a `onSubmit` handle should represent form state, not the last submitted state.

### Other informations

It can be workaround by checking if initialValues object is empty and passing null instead. Then the form will work as expected.

### LINKED ISSUE #3366: FieldArray removing property names if value is empty on blur (added after initialization)

Bug Report

When an object is pushed to a field array, it can have properties with values defined in advance. (In the example below, note 'bob' as a pre-defined `firstName` when adding a member. However, when a property's value is an empty string blurring the corresponding input field causes the the form values to drop the property name in question.

To reproduce:
- open link below
- open console
- add a member
- open fields array printed in console. has both firstName and lastName props in first object.
- blur lastName without changing the value
- open newly printed fields array in console, lastName property is missing.
This appears to happen in `@@redux-form/BLUR`.

That the key/value persist in the object.

### Sandbox Link

[A reproduction of the issue.](https://codesandbox.io/s/myjp96vk8p)

### What's your environment?

react v15.4.1
redux v3.6.0
redux-form v6.3.2

ms88privat:
possibly fixes #2597  #3366 

@erikras What's your opinion here?

I will create an additional test to catch this case for the future.

Edit: Let me try to restructure it. I don't like to have this logic inside a util function. 

Edit2: The second commit did restructure the implementation. `deleteInWithCleanUp` now takes an optional `shouldDelete` function, where you can specify when to delete and when not.

Edit3: Added a test to the `BLUR` reducer, if you like it, I can add it also to the `CHANGE` and `CLEAR_FIELDS` reducer.

### DIFF
--- a/src/__tests__/deleteInWithCleanUp.spec.js
+++ b/src/__tests__/deleteInWithCleanUp.spec.js
@@ -1,4 +1,4 @@
-import createDeleteInWithCleanUp from '../deleteInWithCleanUp'
+import createCreateDeleteInWithCleanUp from '../deleteInWithCleanUp'
 import plain from '../structure/plain'
 import plainExpectations from '../structure/plain/__tests__/expectations'
 import immutable from '../structure/immutable'
@@ -6,7 +6,7 @@ import immutableExpectations from '../structure/immutable/__tests__/expectations
 
 const describeDeleteInWithCleanUp = (name, structure, setup) => {
   const { fromJS } = structure
-  const deleteInWithCleanUp = createDeleteInWithCleanUp(structure)
+  const deleteInWithCleanUp = createCreateDeleteInWithCleanUp(structure)()
 
   describe(name, () => {
     beforeAll(() => {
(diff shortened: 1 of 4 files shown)


ms88privat:
@erikras Any opinion on that one?

PR #3624, opened 2017-11-20
Added optional type definitions for Change action type

### LINKED ISSUE #3620: `touch` and `persistentSubmitErrors` not marked as optional in action creators
Action creators need to indicate which arguments are optional.

For example

```javascript
const change: Change = (
  form: string,
  field: string,
  value: any,
  touch: boolean,
  persistentSubmitErrors: boolean
): ChangeAction => ({
  type: CHANGE,
  meta: { form, field, touch, persistentSubmitErrors },
  payload: value
})
```
should be
```javascript
const change: Change = (
  form: string,
  field: string,
  value: any,
  touch?: boolean,
  persistentSubmitErrors?: boolean
): ChangeAction => ({
  type: CHANGE,
  meta: { form, field, touch, persistentSubmitErrors },
  payload: value
})
```
at least according to the API docs.

Without this, I get flow errors dispatching `change` because I'm not using enough arguments.

pfftdammitchris:
Fixes #3620

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -235,8 +235,8 @@ const change: Change = (
   form: string,
   field: string,
   value: any,
-  touch: boolean,
-  persistentSubmitErrors: boolean
+  touch?: boolean,
+  persistentSubmitErrors?: boolean
 ): ChangeAction => ({
   type: CHANGE,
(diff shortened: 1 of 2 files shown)


pfftdammitchris:
Ready to merge

Issue #2827, opened 2017-04-19
Print warning instead of error when using async


ccjruemenapf:
Hi Folks,
i would be nice, if there will be a possibility to create warnings instead of errors when using the async validation. 
And if i am not wrong only asyncErrors exists and no asyncWarnings.

We want to build async uniqueness checks and show a warning not an error, if the endpoint sad - Nope.

Is there a chance or a plan to get a feature like this?

Greetings and thank you
Jan

griffinmichl:
Agreed on this enhancement! I want it in an app I'm working on where there's a URL input field. I'd like to be able to warn users that we couldn't resolve the URL they input.  
  
I'm gonna work on this if no one else is.

kentandersen:
It would be great if this could have the same api as [asyncValidate](https://redux-form.com/6.7.0/examples/asyncValidation/).
For example:
```
reduxForm({
  form: 'asyncWarn',
  validate,
  asyncValidate,
  asyncValidateBlurFields: ['username'], // renamed form asyncBlurFields to prevent confusion
  asyncWarn,
  asyncWarnBlurFields: ['username']
})(AsyncWarnForm)
```

`asyncBlurFields` should be an alias to `asyncValidateBlurFields` to prevent breaking backwards compatibility

SimonPNorra:
Hey @erikras and @griffinmichl,

do you have any idea for a workaround to support warnings in asyncValidations? We really need it in our current project and we are struggling finding a solution.

Thanks a lot.

Regards,
Simon

terolepisto:
Has this have any progress? Is there a way to "push" any warnings otherway, like we have ability to change form values with this.props.change("foo","bar") ?

Like this.props.warning("foo","Double check Your Foo!") ?

AndyCJ:
Has anyone found a workaround for this?

As @terolepisto says, if there's a way to manually add warnings to a form then external code could handle the async validation warning check and add to warning to the form when it returns.

terolepisto:
I've now used reducers to pass errors to the components which means I have
to pass them as props deep deep deep to the actual input fields and forms.

---
Tero Lepistö

pe 26. tammikuuta 2018 klo 13.35 Andy Jump <notifications@github.com>
kirjoitti:

AndyCJ:
Thank you!

I hadn't thought of side-stepping redux-forms entirely, but that makes a lot of sense.

That gives me a workaround to keep me going.

[...]

PR #3631, opened 2017-11-22
Export updateSyncWarnings action

dagstuan:
Since #2827 is not implemented I have written code around it by using the `updateSyncWarnings`-action from the library to manually update warnings after doing an async call. However, this action is not currently exported from the library, so instead of just doing `import { updateSyncWarnings } from 'redux-form'` I have to do `import actions from 'redux-form/lib/actions'`.


erikras:

PR #3667, opened 2017-12-06
Added check for submitting when checking for onChange and onBlur 

marcusnilssonjickl:
To avoid errors on name is undefined when submitting


gustavohenke:
> To avoid errors on name is undefined when submitting

How does this relates to this change?

Anyway, this PR needs tests and possibly some tests, although I'm not sure if this isn't a breaking change -- cc @erikras

erikras:
Closing due to inactivity.

PR #3697, opened 2017-12-17
Fix field-level validation on previously emptied FieldArray

adamcapone:
Empty `lastFieldValidatorKeys` when  no field level validators were found. This allows `validateIfNeeded` to detect when a previously emptied FieldArray has an item added back.


erikras:

Issue #3636, opened 2017-11-24
[BUG] Swallowed errors in onSubmitSuccess when onSubmit returns a Promise


ClementParis016:
### Are you submitting a **bug report** or a **feature request**?
Bug Report

### What is the current behavior?
When returning a Promise from `onSubmit` handler, errors happening in `onSubmitSuccess` get swallowed.

### What is the expected behavior?
Expect errors happening in `onSubmitSuccess` to be logged to the console.

### Sandbox Link
Here is a CodeSandbox to reproduce the error: https://codesandbox.io/s/18yk8rpn54

### What's your environment?
redux-form@7.1.1
react@16.0.0
redux@3.7.2
react-redux@5.0.6
Chromium@62

I ran into this issue while trying to redirect on successful form submission. My `onSubmit` handler successfully made a network request and returned a Promise (through redux-thunk), but then I had a `TypeError` in my `onSubmitHandler` that I couldn't even know about because it wasn't bubbling up to reach the console :disappointed:

erikras:
Published fix in [`v7.2.1`](https://github.com/erikras/redux-form/releases/tag/v7.2.1).

PR #3723, opened 2017-12-29
Log onSubmitSuccess thrown errors

h1b9b:
Following this [issue](https://github.com/erikras/redux-form/issues/3636), I added a `console.error` to get the expected behavior  
Is it ok ? I have the impression that I didn't understand well the bug, cause it's normal that the error get swallowed as it's thrown in a promise

### DIFF
--- a/src/handleSubmit.js
+++ b/src/handleSubmit.js
@@ -56,24 +56,30 @@ const handleSubmit = (
       }
       if (isPromise(result)) {
         startSubmit()
-        return result.then(
-          submitResult => {
-            stopSubmit()
-            setSubmitSucceeded()
-            if (onSubmitSuccess) {
-              onSubmitSuccess(submitResult, dispatch, props)
+        return result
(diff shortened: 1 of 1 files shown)


erikras:
Can you add a test that fails _before_ your change, but passes _after_ your change?

erikras:

src/handleSubmit.js
```diff
@@ -56,33 +56,38 @@ const handleSubmit = (
...
         startSubmit()
-        return result.then(
-          submitResult => {
-            stopSubmit()
-            setSubmitSucceeded()
-            if (onSubmitSuccess) {
-              onSubmitSuccess(submitResult, dispatch, props)
+        return result
+          .then(submitResult => {
+            try {
```
naw: What is the purpose of this `try` block if you're just rethrowing the exception? Seems extraneous, but maybe I'm missing something.

Issue #3658, opened 2017-12-04
Error when submit form with asyncValidate and asyncChangeFields


swarthy:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

When submit form with `asyncValidate` and `asyncChangeValues`  got error because of that line:
https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L95
`asyncValidate` called with no arguments, so here
https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L684-L689
after webpack build conditions are inlined:
```javascript
var fieldNeedsValidation =
  !submitting &&
  trigger === 'blur'
    ? !asyncBlurFields || ~asyncBlurFields.indexOf(name.replace(/\[[0-9]+\]/g, '[]'))
    : !asyncChangeFields || ~asyncChangeFields.indexOf(name.replace(/\[[0-9]+\]/g, '[]'));              
```
`trigger` is not blur (its undefined, like all other args), `asyncChangeFields` is array

### What is the expected behavior?
Do not throw error?

### Sandbox Link

Example: [JSFiddle](https://jsfiddle.net/036ur33k/10/)

### What's your environment?

React 16.2.0
Redux-Form 7.2.0
Ubuntu 17.10 Google Chrome 62.0.3202.94

### Other information

Error stacktrace
```
Uncaught TypeError: Cannot read property 'replace' of undefined
    at Form._this.asyncValidate (VM731 redux-form.js:7774)
    at handleSubmit (VM731 redux-form.js:9630)
    at Form._this.submit (VM731 redux-form.js:7816)
```

**UPD:** My bad, code is ok, looks like it's babel/webpack issue?
source:
```javascript
            const fieldNeedsValidation =
              !submitting &&
              (formHasAsyncFields ||
                (trigger === 'blur'
                  ? fieldNeedsValidationForBlur
                  : fieldNeedsValidationForChange))
```
after translation no parentheses (in JSFiddle too):
```javascript
var fieldNeedsValidation = !submitting && trigger === 'blur' ? !asyncBlurFields || ~asyncBlurFields.indexOf(name.replace(/\[[0-9]+\]/g, '[]')) : !asyncChangeFields || ~asyncChangeFields.indexOf(name.replace(/\[[0-9]+\]/g, '[]'));
```

andyb-:
I'm having the same issue and see the same error stacktrace.
```
[...]

PR #3733, opened 2018-01-06
fix submit crashing when asyncBlurFields or asyncChangeFields is given

imrekoszo:
Fixes https://github.com/erikras/redux-form/issues/3658

Also added some variations to an existing test to verify the change. The new tests will fail without the code change.


imrekoszo:
Damn, that codeclimate is tough to satisfy.

Issue #3738, opened 2018-01-11
v5 remove array field by predicate function


danikkks:
### Are you submitting a **bug report** or a **feature request**?

**feature request**

### What is the current behavior?

User can delete array value only by index

### What is the expected behavior?
User can delete array value by predicate function

### Other information

In some large forms with lot of array fields order of fields mapped to component is different from fields in store, and you can`t delete right field by index

gustavohenke:
Since your PR was merged, I'll close this.

PR #3739, opened 2018-01-11
V5 new feature: remove array field by predicate function

danikkks:
solution for this feature request - https://github.com/erikras/redux-form/issues/3738


danikkks:
@danikkks
fix: getValue not return object when one of keys is "value"
9ea4a7c
 @danikkks
add _initData prop to reducer
7896340
 @danikkks
remove .npmrc and add it to gitignore
d5370bf
 @danikkks
remove .npmrc and add it to gitignore
a0c083d
 @danikkks
fix version
db300aa
 @danikkks
Merge branch 'v5' of https://github.com/danikkks/redux-form into v5
a386334

this commits are already merged here - https://github.com/erikras/redux-form/pull/3382
I dont know how to remove them from this pull request

src/isFunction.js
```diff
@@ -0,0 +1,4 @@
+export default (value) => {
```
itommen: why don't you use the isFunction from loadash?

https://lodash.com/docs/4.17.4#isFunction

src/reducer.js
```diff
@@ -79,10 +80,20 @@ const behaviors = {
   [INITIALIZE](state, {data, fields, overwriteValues}) {
     return createInitialState(data, fields, state, overwriteValues);
   },
-  [REMOVE_ARRAY_VALUE](state, {path, index}) {
+  [REMOVE_ARRAY_VALUE](state, {path, indexOrPredicate}) {
```
itommen: suggestion: it would be better if the param 'indexOrPredicate' would split into 2 different params, the first 1 is 'index' and the second is 'predicate', and the logic is basically the same.

when you check isFunction the indexOrPredicate, you'll check if (predicate && isFuncion(predicate))

the reason for the split is to keep Single Responsibility, each param should have only 1 purpose, and the word 'or' applies that this param have 2 different purposes

it's just a suggestion :)

src/isFunction.js
danikkks: v5 of redux-form don't include lodash as dependency

danikkks:
> @itommen commented on this pull request.

> In src/reducer.js:

> @@ -79,10 +80,20 @@ const behaviors = {
   [INITIALIZE](state, {data, fields, overwriteValues}) {
     return createInitialState(data, fields, state, overwriteValues);
   },
> -  [REMOVE_ARRAY_VALUE](state, {path, index}) {
> +  [REMOVE_ARRAY_VALUE](state, {path, indexOrPredicate}) {

> suggestion: it would be better if the param 'indexOrPredicate' would split into 2 different params, the first 1 is 'index' and the second is 'predicate', and the logic is basically the same.
> 
> when you check isFunction the indexOrPredicate, you'll check if (predicate && isFuncion(predicate))
> 
> the reason for the split is to keep Single Responsibility, each param should have only 1 purpose, and the word 'or' applies that this param have 2 different purposes
> 
> it's just a suggestion :)
> 
> 

DONE!

erikras:

PR #3743, opened 2018-01-11
Update es6-error module

adybionka:
es6-error in version 4.1.0 has an error and due of this redux-form doesn't show errors. 

An instance of SubmissionError class doesn't recognize as its class. 

```
const error= new SubmissionError({...});

error instanceof SubmissionError // gives false but should true
```

No issue with module es6-error@4.1.1. 

Unfortunately even there is never version of es6-error, yarn on my production server still installs broken version which is 4.1.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
   "homepage": "https://redux-form.com/",
   "dependencies": {
     "deep-equal": "^1.0.1",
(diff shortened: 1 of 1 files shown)


mrova:
👍

p-grzelczak:
Encountered described issue aswell, this PR solves it 👍

Issue #3303, opened 2017-08-09
Errors are being swallowed again, due to #3227


miracle2k:
### What is the current behavior?

If my form's onSubmit handler returns a promise, and the onSubmit handler is passed as a form option, than the promise returned is rejected, that rejection is swallowed and I never see the error.

### What is the expected behavior?

The error thrown is passed up to the browser console.

### Sandbox Link

https://codesandbox.io/s/5w7Gmk6Z

Note how pressing Submit logs no error.

### What's your environment?

Using 7.0.2.

### Other information

Possibly related to https://github.com/erikras/redux-form/issues/3179.

I tried to debug this and I believe the problem is `listenToSubmit`: https://github.com/erikras/redux-form/blob/171ed689cf998774c39bcd435d73dd8228cb7e02/src/createReduxForm.js#L657

It seems this was  introduced recently in https://github.com/erikras/redux-form/pull/3227, because before that submitFailed was throwing the error.

Roman-Maksimov:
The problem with the #3227 (issue #1297) was about the `handleSubmit` method mustn't throw any errors as it doesn't handle the errors anyhow. As a result the `uncaught error` appears and falls down the rest of the JS.
Don't know why do you want to have the `uncaught error` in your case (seems you used to integrate the bug into your logic).
If you need to catch the error, there is a `onSubmitFailed` method:
```javascript
export default reduxForm({
  form: 'submitValidation', // a unique identifier for this form
  onSubmit: async () => {
      throw "test";
  },
  onSubmitFail: (errors, dispatch, submitError, props) => {
    console.log(submitError);
  }
})(SubmitValidationForm);
```

miracle2k:
I don't want the error. It's a bug in my onSubmit logic (say I call a relay mutation but the import is wrong and the function undefined).As it stands, I don't have any indication that something is wrong, and I don't see an error it a stack trace, so I don't know where the problem is precisely. In my code I actually call handleSubmit myself and the fact that I can get error values back just like a successful return value is an added complication, but not the main issue.I think onSubmitFail might be a workaround, but I don't know that we shouldn't view it as a bug that such errors are swallowed out of the box. On Aug 9, 2017 9:41 PM, Roman Maksimov <notifications@github.com> wrote:The problem with the #3227 (issue #1297) was about the handleSubmit method mustn't throw any errors as it doesn't handle the errors anyhow. As a result the uncaught error appears and falls down the rest of the JS.
[...]

PR #3744, opened 2018-01-12
Ensure errors thrown from submit handler are not silenced

10xjs:
Fixes https://github.com/erikras/redux-form/issues/3303

In the very likely case that an error is encountered during a submit callback we need to be able to reliably detect any failures.

This restores the correct behavior stated in the [documentation](https://redux-form.com/7.2.0/docs/api/reduxform.md/#-code-submit-promise-code-):

> Submits the form. [You'd never have guessed that, right?] Returns a promise that will be resolved when the form is submitted successfully, or rejected if the submission fails.

This will revert the change made in https://github.com/erikras/redux-form/pull/3227, but throwing the error is a _far better_ compromise that benefits a _much larger set_ of redux-form users. It's possible to silence errors in your own code, but not possible to recover discarded errors from inside an isolated context in a lib.

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -718,7 +718,7 @@ const createReduxForm = (structure: Structure<*, *>) => {
 
         submitFailed = (error: any): void => {
           delete this.submitPromise
-          return error
+          throw error
         }
 
         listenToSubmit = (promise: any) => {


erikras:

dave-irvine:
Hi @10xjs, @erikas,

While I 100% agree with this pull, errors should not be swallowed, I'm wondering how we are supposed to consume the error thrown here?

At the moment the error I throw in asyncValidate is not handled anywhere and ends up in the console. I'd love to catch the error, but I can't find anywhere to catch what this change throws.

Any tips?

10xjs:
@dave-irvine I haven't used `asyncValidate` myself, but I believe that you are able to catch its errors with the `onSubmitFail` prop. (https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L107)

dave-irvine:
Hi @10xjs, unfortunately no matter what you do in onSubmitFail, https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L109 rejects with the errors, which are then subsequently thrown by your change. :(

10xjs:
I wonder if we need a new `onAsyncValidationFail` prop?

PR #3755, opened 2018-01-15
Fix async/submit errors getter for nested fields

xcrzx:
asyncErrors & submitErrors are flat key-value objects: `{ 'field.a.b.c': 'error' }`, so we should escape field name in order to get data from it.

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -306,12 +306,12 @@ const createConnectedField = (structure: Structure<*, *>) => {
       const syncWarning = getSyncWarning(getIn(formState, 'syncWarnings'), name)
       const pristine = deepEqual(value, initial)
       return {
-        asyncError: getIn(formState, `asyncErrors.${name}`),
+        asyncError: getIn(formState, `asyncErrors['${name}']`),
         asyncValidating: getIn(formState, 'asyncValidating') === name,
         dirty: !pristine,
         pristine,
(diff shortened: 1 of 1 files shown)


erikras:
Doesn't this work, too, and is better?
```jsx
asyncError: getIn(formState.asyncErrors, name),
```

xcrzx:
Unfortunately not.

```
const formState = {
    asyncErrors: {
        'a.b.c': 'Error message',
    }
};

const name = 'a.b.c';

getIn(formState, `asyncErrors['${name}']`); // 'Error message'
getIn(formState.asyncErrors, name); // undefined
```

Following code `getIn(formState.asyncErrors, 'a.b.c')` is trying to access `formState.asyncErrors.a.b.c` while actually we need `formState.asyncErrors['a.b.c']`

erikras:

sergeyfedotov:
It's a breaking change. Before v7.2.1 only nested error object could be used.

dagstuan:
This should have been released as `8.0.0` as this is a breaking change.

erikras:
Apologies for this accidental breaking change... it should be rolled back (unbroken).

erikras:
Reverted this change in [`v7.2.2`](https://github.com/erikras/redux-form/releases/tag/v7.2.2), as it broke several things. Let's discuss how to accomplish what you need to.

PR #3756, opened 2018-01-15
fix validation after duplicate field unmount

### LINKED ISSUE #3688: Validation breaks after unmounting of a duplicated field
This is a bug.

When two fields for the same form value are rendered, then one of these fields is unmounted, validation stops to work.
For the provided sandbox:
1. check that the validation is working for the field
2. add another field (+ button) - validation is still working
3. remove a field (- button) - validation is broken

In a more real-world scenario it occured to me when a part of the form was being re-rendered in a different branch of the DOM tree - 'new' fields get registered before old ones get unregistered, causing the error to occur.

Validation should work if there's a field rendered that requires it

### Sandbox Link
https://www.webpackbin.com/bins/-L0ES7_aTU4kA-O9gsgc

### What's your environment?
react 16,
redux-form 7.2.0

### Other information
Probably need to check the count of the fields before removing the validators?
https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L585

jedwards1211:
fix #3688 
fixed by keeping a reference count of names registered with `ReduxForm.register`/`unregister`, just like the reducer keeps a reference count of fields.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -5336,6 +5336,71 @@ const describeReduxForm = (name, structure, combineReducers, setup) => {
       expect(propsAtNthRender(formRender, 1).valid).toBe(false)
       expect(propsAtNthRender(formRender, 2).valid).toBe(true)
     })
+    
+    it('should preserve validators when a duplicate field is unmounted', async () => {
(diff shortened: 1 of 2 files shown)


erikras:

jedwards1211:
Muchas gracias!

PR #3758, opened 2018-01-16
Fields re-render every time due to regression introduced with flow typings.

dagstuan:
This PR fixes a regression introduced in #3138 where the `shouldComponentUpdate` method of `createField.js` no longer sent `nextState` as an argument to `shallowCompare` when checking if a component should update. This regression effectively made every field re-render every time since the default state of a react component is `null` and an unspecified argument is `undefined`, making `shallowCompare` do a `null === undefined` check to determine if the component should update.

### DIFF
--- a/src/util/shallowCompare.js
+++ b/src/util/shallowCompare.js
@@ -24,8 +24,11 @@ const shallowCompare = (
   instance: { props: any, state?: any },
   nextProps: Object,
   nextState?: Object
-): boolean =>
-  !isEqualWith(instance.props, nextProps, customizer) ||
-  !isEqualWith(instance.state, nextState, customizer)
+): boolean => {
+  const propsEqual = isEqualWith(instance.props, nextProps, customizer)
+  const stateEqual = isEqualWith(instance.state, nextState, customizer)
+
+  return !propsEqual || !stateEqual
+}
 
 export default shallowCompare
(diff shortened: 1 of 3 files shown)


src/util/shallowCompare.js
```diff
@@ -24,8 +24,11 @@ const shallowCompare = (
...
   nextProps: Object,
   nextState?: Object
-): boolean =>
-  !isEqualWith(instance.props, nextProps, customizer) ||
-  !isEqualWith(instance.state, nextState, customizer)
+): boolean => {
+  const propsEqual = isEqualWith(instance.props, nextProps, customizer)
+  const stateEqual = isEqualWith(instance.state, nextState, customizer)
+
+  return !propsEqual || !stateEqual
```
gustavohenke: I like how readable this function is now 💯

dagstuan:
@erikras is there any way to retry travis-ci? I dont really understand why it failed, since there seems to be no logs :)

dagstuan:
@gustavohenke I attempted for a while to write a test for it. But for some reason the actual component specified to the `<Field />` does not re-render, probably due to smart react-diffing or something. However, the `render()`-function of the `<Field />` re-renders every time. 

The test I wrote to check if is basically equal to https://github.com/erikras/redux-form/blob/878176b0e960d58a3277a2bd83703e8da3e82eb6/src/__tests__/Field.spec.js#L874 however I could not make the `input`-function render twice.

dagstuan:
@gustavohenke I figured out a way :) Had to make a sub-class of the Field-class to check it though, so the test is kind of hacky. If you know of any other way of testing this please let me know :)

erikras:

PR #3773, opened 2018-01-22
Add info about asyncChangeFields to docs

b0gok:
After this PR https://github.com/erikras/redux-form/pull/3531 there is new option in redux-form.
I added info about it to the docs.

### Default value for shouldAsyncValidate

Interesting thing about default values for `shouldAsyncValidate` option:

- In the PR there were two cases: `blur` and `change`
https://github.com/ericbiewener/redux-form/blob/270ea7b1114ff57cbb74e3bdf0faac4ccc575b3f/src/defaultShouldAsyncValidate.js#L22
- In master there is only `blur` case
https://github.com/ericbiewener/redux-form/blob/master/src/defaultShouldAsyncValidate.js#L21

### Suggestions
Maybe it will be more clear to change key from `blurredField` to `changedField` there
https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L718

### DIFF
--- a/docs/api/ReduxForm.md
+++ b/docs/api/ReduxForm.md
@@ -42,6 +42,15 @@ at "design time" or passed in as props to your component at runtime.**
 > [Asynchronous Blur Validation Example](https://redux-form.com/7.2.1/examples/asyncValidation/)
 > for more details.
 
+#### -`asyncChangeFields : Array<String>` [optional]
+
+> field names for which `onChange` should trigger a call to the `asyncValidate`
+> function.
+
+> See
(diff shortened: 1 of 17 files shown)


erikras:
Looks good.

erikras:
Voila! https://redux-form.com/7.2.2/examples/asyncchangevalidation/

b0gok:
@erikras what do you think about my suggestion?
> Maybe it will be more clear to change key from blurredField to changedField there
https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L718

PR #3787, opened 2018-01-28
Fix error in multiple async validation

anpieber:
The Problem is that a successful validation triggered directly after an error does not get cleaned up correctly. The reason is that the asyncErrors field never gets cleared. This wasn't a problem in 6.x, but could easily be reproduced in 7.x. Simply go to https://redux-form.com/7.2.1/examples/asyncvalidation/ enter a password, go to the user field, enter john. No execute the following actions faster than one second:

1. press tab (going to password, triggering blur and validation)
1. press shift+tab (going back to user)
1. enter any letter (which will be a correct result in this example)
1. press tab again (triggering another blur and another validation which should be successful now)

The result should be that the error goes away once the second validation had been triggered (1-2 seconds later), BUT the error does not go away - since the asyncErrors still persist in the redux store.

Using my patch fixes the problem. TBH I dont know if I broke any rules or miss any tests the way I submit the report and I would gladly provide any additional information/data/changes you require.

With one field, such as in the example, it's not really a big deal, but with multiple fields validated at once it's quite easy to get into this situation.

### DIFF
--- a/src/createReducer.js
+++ b/src/createReducer.js
@@ -418,6 +418,7 @@ function createReducer<M, L>(structure: Structure<M, L>) {
         }
       } else {
         result = deleteIn(result, 'error')
+        result = deleteIn(result, 'asyncErrors')
       }
       return result
     },


anpieber:
This problem still persists in 7.2.3 and is as easily to reproduce

erikras:

eliseumds:
Thank you very much for this! We were having this issue but I couldn't understand why it was happening and I spent quite a while thinking that it was because my validation functions were debounced. Cheers!

Issue #3793, opened 2018-01-29
[BUG] formValues does not update on props change


manubb:
### What is the current behavior?
When a component is decorated with `formValues` using a function argument, connected values are not updated when props of the component change.
For example:
```js
formValues({ person } => ({
  name: `${person}.name`,
}))(Component)
```
When `person` changes, the component is still connected with the name of the person it has when mounted.

### What is the expected behavior?
I am expecting formValues to react to props change.

### Other information
The cause of the problem is straightforward when reading the [sources](https://github.com/erikras/redux-form/blob/v7.2.1/src/createFormValues.js): 
 * `valuesMap` is only computed in `constructor` and is not computed again when props change. 
 *  `this.Component` is not updated either and the `render` method uses the initially computed `this.Component`.

I can PR a patch if you feel this is worth to fix.

gustavohenke:
Great, it seems like you already have the fix! Can you send the fix in a PR?

manubb:
The fix still needs a little work. I will PR this evening or next. Thanks for fast reply.

erikras:
Published fix in [`v7.2.2`](https://github.com/erikras/redux-form/releases/tag/v7.2.2).

PR #3794, opened 2018-01-29
Fix for formValues not reacting to props change

manubb:
Fix proposal for #3793

### DIFF
--- a/src/createFormValues.js
+++ b/src/createFormValues.js
@@ -1,11 +1,12 @@
 // @flow
+import { isEmpty, isEqual, mapValues } from 'lodash'
 import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import prefixName from './util/prefixName'
 import type { ComponentType } from 'react'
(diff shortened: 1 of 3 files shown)


src/createFormValues.js
```diff
@@ -25,57 +27,65 @@ const createValues = ({ getIn }: Structure<*, *>): FormValuesInterface => (
...
           typeof firstArg === 'function' ? firstArg(props) : firstArg
         if (typeof resolvedFirstArg === 'string') {
-          valuesMap = [
-            resolvedFirstArg,
-            ...rest
-          ].map((k: string): PropPath => ({
-            prop: k,
-            path: k
-          }))
+          valuesMap = {}
```
itommen: you better use reduce function over init empty object and then go over the array with foreach to init the values

[resolvedFirstArg, ...rest].reduce((result, current) => { ...result, [current]: current }, {});

manubb: Fixed

erikras:

PR #3797, opened 2018-01-30
Fix for double INITIALIZE happening on first mount

### LINKED ISSUE #3690: INITIALIZE action is called twice when initialValues is presented
**Bug** introduced by https://github.com/erikras/redux-form/pull/3615

`INITIALIZE` action is called twice when `initialValues` prop is used. This behaviour can cause problems with reinitializing props for the second time after values could be already changed (in `componentWillMount` of children of the form).

Call `INITIALIZE` once. 

Do not call for the second time in `componentDidMount` or handle properly `initialized` in `this.initIfNeeded`, which is `false` after the first initialization in `componentWillMount`.

### What's your environment?
Redux Form 7.2.0 

### Other information
Bug was introduced in https://github.com/erikras/redux-form/releases/tag/v7.2.0 in bug fix called "Fixed hot reloading problem with React 16" https://github.com/erikras/redux-form/pull/3615

### LINKED ISSUE #3706: Values via props.change() in componentDidMount() will be overwritten
I assume this is a bug, since it worked in previous versions of redux-form.
Edit: Can confirm it worked in 7.1.2

We have a form with a few fields which may have a default value. Those are set via props.change(),
 see the shortened code below.
```js
componentWillMount() {
  this.preSelectCountry = locale === 'de';
}

componentDidMount() {
    if (this.preSelectCountry) {
      this.props.change('country_code', Object.keys(this.CountryList)[0]);
    }
}
```

`@@redux-form/INITIALIZE` is dispatched as first action, followed a couple of `@@redux-form/REGISTER_FIELD`
`@@redux-form/CHANGE` is dispatched with correct values. `country_code` is a registered field the action results in the correct change of the `values` state for `country_code`.
Then again `@@redux-form/INITIALIZE` is dispatched resulting in overwriting the previously set values. Then fields are registered and unregistered again.

To this time there is not a single action which is not dispatched by redux-form.
Some information to set the initial value of those fields are only available as props, I am unsure about how to use`reduxForm({ ...initialValues: { ... }})` to do this. This would work, since we use this method to set a initial value to a field. 

Form Validation will somehow consider those set values, they are even displayed correctly. But are missing when submitting.

Note: There is not much more logic in this app besides sending the form to server and displaying a result page. 

I don't expect the form the initialize twice, resulting in overwriting previously set data.

### Sandbox Link
I will setup a demo for this as soon as I have time for it.

Edit: I condensed the application we have: https://codesandbox.io/s/konx6r9zpv

### What's your environment?
redux-form: 7.2.0
the app is tested with: Chrome, Firefox, Safari, IE10/11 (😢) on osx/window

bmv437:
This is a fix for an issue where the INITIALIZE action is fired twice, clearing change actions that happen when a form mounts.

See #3690 for details of the bug.

This fixes #3706 and fixes #3690.

I attempted to remove those formatting changes, but they got reversed by the precommit hook I believe.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -735,6 +735,71 @@ const describeReduxForm = (name, structure, combineReducers, setup) => {
       expect(propsAtNthRender(inputRender, 0).input.value).toBe('bar')
     })
 
+    it('should support calling change in Form componentDidMount', () => {
+      const store = makeStore({})
+      const inputRender = jest.fn(props => <input {...props.input} />)
+      const formRender = jest.fn()
+      const initialValues = {
+        deep: {
+          foo: 'bar'
+        }
+      }
+      class Form extends Component {
+        componentDidMount() {
+          this.props.change('deep.foo', 'baz')
+        }
+        render() {
+          formRender(this.props)
+          return (
+            <form>
+              <Field name="deep.foo" component={inputRender} type="text" />
+            </form>
+          )
+        }
+      }
+      const Decorated = reduxForm({ form: 'testForm' })(Form)
+      TestUtils.renderIntoDocument(
+        <Provider store={store}>
+          <Decorated initialValues={initialValues} />
+        </Provider>
+      )
+      expect(store.getState()).toEqualMap({
+        form: {
+          testForm: {
+            initial: initialValues,
+            values: {
+              deep: {
+                foo: 'baz'
+              }
+            },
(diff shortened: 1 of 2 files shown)


erikras:
Dammit. I accidentally reverted this. Can you submit a new copy of this PR from your fork?

erikras:
Nevermind. [I did it myself](https://github.com/erikras/redux-form/pull/3805).

erikras:

PR #3802, opened 2018-01-31
Fix flat structure for asyncErrors and submitErrors for FieldArray and Fields.

### LINKED ISSUE #3779: On SubmissionError section nested fields aren't highlighted as error
With the current 7.2.1 version SubmissionErrors that happen after server responds with invalid set of action -> field, the invalid fields aren't updated and highlighted as invalid!
*In case if i move the fields out of the section (remove the section) the errors that came from server validation are highlighted correctly. (https://codesandbox.io/s/2zzry94pp)*

I expect that after handleSubmit is getting a `throw SubmissionError({ section1: { field1: 'error happend'}})` error the corresponding field is highlighted as invalid!

### Sandbox Link (Fields wrapped within section are not highlighted)
https://codesandbox.io/s/l9k27jky1m

If you enter a username (which is required) and press on the submit button i expect that 
`{
      applicant: { username: 'User does not exist 3333' }
    }` is getting highlighted with `User does not exist 3333`

dagstuan:
This PR fixes the rest of #3755 that was introduced in `7.2.1` where nested async errors and submit errors are flat. The other change should have been released as `8.0.0` since this is a breaking change in how `asyncErrors` and `submitErrors` are handled by redux-form.

This PR should also fix #3779 

Note that this change requires errors for parent FieldArrays to be structured as `foo._error` when returned from the server. The tests should be clear as to what will happen.

### DIFF
--- a/src/ConnectedFieldArray.js
+++ b/src/ConnectedFieldArray.js
@@ -149,11 +149,11 @@ const createConnectedFieldArray = (structure: Structure<*, *>) => {
       const syncWarning = getSyncWarning(getIn(formState, 'syncWarnings'), name)
       const pristine = deepEqual(value, initial)
       return {
-        asyncError: getIn(formState, `asyncErrors.${name}._error`),
+        asyncError: getIn(formState, `asyncErrors['${name}._error']`),
         dirty: !pristine,
         pristine,
         state: getIn(formState, `fields.${name}`),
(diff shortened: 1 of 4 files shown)


marlonmleite:
Could they generate a release? This is affecting me.

gustavohenke:
@marlonmleite this is released as of v7.2.2.

PR #3692, opened 2017-12-14
fix erikras/redux-form#3489

Defenderbass:

PR #3803, opened 2018-02-01
Immutable setIn is broken with more than 32 elements in an array.

### LINKED ISSUE #3791: [Bug]Changes to items in FieldArray cause whole list of items to update
Bug

Changing a single item in a FieldArray causes whole list to be resorted. Inspecting actions and state diff using Redux dev tools shows an action dispatched where payload updates the correct item (via index key), but the difference in state includes changes to all items in the list.

Changing a single item only updates that item's state in the redux store.

### Sandbox Link
Attempted to create sandbox with scenario, but there are issues with the current sandbox setup with 7.2.1
https://codesandbox.io/s/811pjkq2y9

### What's your environment?
"redux-form": "7.2.1"
"redux": "^3.5.2"
"immutable": "^3.8.1"
"react-redux": "^5.0.1",
Chrome  63.0.3239.132 
Node 8.9.1

### Other information
Note this does not happen in version 7.0.4/7.1.2 and when number of items provided to FieldArray is less than 32.

dagstuan:
This PR fixes an issue introduced in #3692 where the `mergeLists` method in `setIn.js` for `immutable.js` was updated due to `mergeDeepWith` being removed as a `List` function in `immutable.js@4.0.0-rc.5`. Unfortunately this change introduced and error in which arrays with more than 32 elements were behaving unexpectedly since the new way of doing it didnt use ordered maps, instead just calls `toMap()`, which causes the merging to lose order with more than 32 elements due to optimizations in `immutable.js`. After `4.0.0-rc.5`, the `mergeDeepWith`-method is exported as a global export instead of being attatched to `Map`, and `List`. This PR uses the global export version `4.x` and uses the old way of merging for users with `3.x`. I also added a test with long arrays for immutable `setIn`.

This PR should fix #3791

@erikras @gustavohenke can this be merged fast? Its pretty important since all long FieldArrays are currently broken.

### DIFF
--- a/src/structure/immutable/__tests__/setIn.spec.js
+++ b/src/structure/immutable/__tests__/setIn.spec.js
@@ -312,4 +312,25 @@ describe('structure.immutable.setIn', () => {
     let b0c = b0.get('c')
     expect(b0c).toEqual('123')
   })
+  it('should handle arrays with length 64', () => {
+    let initial = fromJS({
+      a: {
+        b: []
+      }
+    })
+
+    let result = setIn(initial, 'a.b[64].d', '12')
+
+    let a = result.get('a')
+    expect(a).toBeTruthy()
+
+    let b = a.get('b')
+    expect(b).toBeTruthy()
+
+    let b64 = b.get(64)
+    expect(b64).toBeTruthy()
+
+    let b64d = b64.get('d')
+    expect(b64d).toEqual('12')
+  })
 })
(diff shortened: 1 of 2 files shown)


dagstuan:
@erikras is it possible to allow the testfile for `setIn` to be longer than 250 lines so codeclimate will accept it? Or should I move the test to a separate file?

erikras:
Code Climate is supposed to be ignoring `*.spec.js`, but their API changed.

cristian-sima:
you are a the God. I just encored this problem and I was writing an issue :D waiting for solution

erikras:

PR #3806, opened 2018-02-01
Fix for formValues not reacting to props change (AGAIN)

erikras:
Duplicate of #3794, which accidentally got reverted.


erikras:

PR #3807, opened 2018-02-01
Reverted breaking change in #3755

erikras:
Oops. #3755 should never have been merged.


erikras:

Issue #3810, opened 2018-02-02
Redux form doesn't handle recursive structures as props to Field component due to deepEqual


itsuryev:
Hello,

I've encountered an issue with deepEqual inside of ConnectedField shouldComponentUpdate call.
Code in question:
```js
        ...
        nextPropsKeys.some(prop => {
          return (
            !~propsToNotUpdateFor.indexOf(prop) &&
            !deepEqual(this.props[prop], nextProps[prop])
          )
        })
```
In my case one of the props is immutable recursive structure which makes this check run in circles.
To handle this on form level we have `immutableProps`:
```
immutableProps : Array<String> [optional]
Prop names that only require strict-equals, not deep equals, to determine shouldComponentUpdate. Useful for performance and compatibility with 3rd party immutable libraries. Defaults to [].
```
But somehow they are not applied on Field level.

Thank you!

itsuryev:
Hello!

Can we please start the ball rolling on this one, at least the discussion.

In the linked pull request (#3849 ) I've added `immutableProps` option to Field component, test that everything fails without it and updated documentation to list new option.
What else should be done?

Thank you!

I humbly summon you, mighty @erikras

PR #3812, opened 2018-02-03
Apply reduxForm immutableProps to Fields as well

itsuryev:
Suggestion for #3810


itsuryev:
Closing this pull request in favor of different solution: `immutableProps` option directly on Field level. It would remove inconsistency when Field behave differently under two forms (one have defined imuutableProps for field, other - didn't).
Will open different pull request once new solution is ready.

PR #3817, opened 2018-02-05
Feat: add keepValues to initialize (closes #3649)

### LINKED ISSUE #3649: Feature Request: initialize new fields and preserve older values
Feature Request

When I dispatch the ActionCreator `initialize('myForm', { myField: 'Init Value' })`. It will initialize all values as expected. But will also wipe all previous `initial` values and all `values` from the store within the form.

The expected behavior is fine, as it is useful in many use cases. The only thing I wish is a ActionCreator, which allows to initialize another field in a already registered form, e.g. `initializeField(form, field, value, ...)`. This will just initialize one field at a time and keep other `initial` values.

Another thing what I can imagine is another option in the ActionCreator `initialize` to preserve older already initialized values, e.g.: `keepValues` or `preserveValues`. 

### What's your environment?
Redux-Form: `v7.2.0`

JPeer264:
I added the feature of #3649. Now I added just an option `keepValues`, but kept the initial values with the same flag. Should another flag `keepInitial` be set for this?

### DIFF
--- a/docs/api/ActionCreators.md
+++ b/docs/api/ActionCreators.md
@@ -88,7 +88,7 @@ actions such as `CHANGE` or `BLUR`, the specific field.
 
 > Marks the given field as `active` and `visited`.
 
-### `initialize(form:String, data:Object, [keepDirty:boolean], [options:{keepDirty:boolean, keepSubmitSucceeded:boolean, updateUnregisteredFields:boolean}])`
+### `initialize(form:String, data:Object, [keepDirty:boolean], [options:{keepDirty:boolean, keepSubmitSucceeded:boolean, updateUnregisteredFields:boolean, keepValues:boolean}])`
 
 > Sets the initial values in the form with which future data values will be
 > compared to calculate `dirty` and `pristine`. The `data` parameter may contain
@@ -106,6 +106,9 @@ actions such as `CHANGE` or `BLUR`, the specific field.
 > initialValue if still pristine instead of only registered fields. Highly
 > recommended, defaults to false because of non-breaking backwards
 > compatibility.
+
+> If the `keepValues` parameter is `true`, it will keep the old values and
+> initial values.
 
 ### `registerField(form:String, name:String, type:String)`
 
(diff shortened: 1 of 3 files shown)


swils:
This is a fantastic feature, and by coincidence exactly what I need right now :-) Anyone an idea on how long would it take for this be part of an official release?

JPeer264:
@swils happy to hear. As you also have a use case for this, would you more prefer to keep the initial values and values just with one option, or separately? Maybe like `keepValues` and `keepInitial`?

Also I guess this might take some time, regarding to other PRs. My temporarily solution (as I really need this right now) would be to fork this repo, merge this branch into master and directly add my repo as dependency in my node modules 😅

swils:
Not sure if I understand what you mean by the difference between `keepValues` and `keepInitial`? :smile:

JPeer264:
Like my PR is about keeping the `values` and the `initial`s. So in the following example I have an already registered field `aField`.

```js
formReducer: {
  myForm: {
	values: {
	  aField: 'some value',
    },
    initial: {
      aField: 'some initial value',
    },
  },
},
```

**Current PR implementation**
When I now initialize another field with the option `{ keepValues: true }` it will keep both previous `values` and `initial`s:

```js
formReducer: {
  myForm: {
	values: {
	  aField: 'some value', // <- kept
      aNewField: 'new value',
    },
    initial: {
      aField: 'some initial value', // <- kept
      aNewField: 'new value',
    },
  },
},
```

**Seperated options**
So what if I don't want to preserve the old `values` and just keep the `initial`s. I need separated options: `{ keepValues: true, keepInitial: true }`: 

Here with the option `{ keepInitial: true }`
```js
formReducer: {
  myForm: {
	values: {
	  // any old value is lost here
      aNewField: 'new value',
    },
    initial: {
      aField: 'some initial value',
      aNewField: 'new value',
    },
  },
},
```

Here with the option `{ keepValues: true }`
```js
formReducer: {
  myForm: {
	values: {
      aField: 'some value',
      aNewField: 'new value',
    },
    initial: {
	  // any old value is lost here
      aNewField: 'new value',
    },
  },
},
```

erikras:

swils:
Hmmm @JPeer264 no real opinion on this, I guess more options are better in this case?

JPeer264:
@swils ya it got merged now, so it is one option (`keepValues`) now.

PR #3825, opened 2018-02-08
Revert "fix: Log onSubmitSuccess thrown errors (#3723)"

### LINKED ISSUE #3801: [BUG] onSubmitFail not called anymore on submitHandler promise rejection in v7.2.1
Hi,

bug report

after moving from **v7.2.0** to **v7.2.1** the registered `onSubmitFail` handler is not called anymore if `submitHandler` rejects the promise.

`onSubmitFail` should be called if an an async submitHandler rejects the promise.

> **onSubmitFail**
> A callback function that will be called when a submission fails **for whatever reason.** It will be called with the following parameters:
> https://redux-form.com/7.2.1/docs/api/reduxform.md/#-code-onsubmitfail-function-code-optional-

### What's your environment?

Chrome

### LINKED ISSUE #3821: redux-form should not write errors to the console explicitly.
Formerly https://github.com/erikras/redux-form/pull/3723 was merged to address https://github.com/erikras/redux-form/issues/3636

The original issue (https://github.com/erikras/redux-form/issues/3636) was that exceptions thrown in an `onSubmitSucceeded` handler were being swallowed. The fix in https://github.com/erikras/redux-form/pull/3723 handled this by explicitly logging an error to the console (i.e. with `console.error`)

I'd like to suggest that this isn't the correct fix for the problem.

When an error is thrown within a promise, the normal and expected behavior is for the promise to reject. This was occurring, but due to a separate issue (https://github.com/erikras/redux-form/issues/3303), the promise was later being marked as resolved.

The _correct_ fix to the original problem is to let the error in an `onSubmitSucceeded` handler bubble up as an unhandled rejection. This has already been solved in https://github.com/erikras/redux-form/pull/3744, although it wasn't solved at the time that https://github.com/erikras/redux-form/issues/3636 was opened.

In other words, the fix in https://github.com/erikras/redux-form/pull/3744 makes the fix in https://github.com/erikras/redux-form/pull/3723 unnecessary, and also the fix (explicit use of `console.error`) in https://github.com/erikras/redux-form/pull/3744 is undesirable. We should not consider exceptions raised in a promise to be errors that should be written to the console -- instead, just let them bubble up as rejected promises.

I'm opening this issue to request that you consider reverting https://github.com/erikras/redux-form/pull/3723

Thanks!

naw:
This reverts commit 1a0660c3477024c094f1906ddab0984a386a45f0 from #3723 

The rationale for this revert is explained in #3821 and #3801 

Closes #3821 
Closes #3801

### DIFF
--- a/src/handleSubmit.js
+++ b/src/handleSubmit.js
@@ -56,30 +56,24 @@ const handleSubmit = (
       }
       if (isPromise(result)) {
         startSubmit()
-        return result
-          .then(submitResult => {
-            try {
-              stopSubmit()
-              setSubmitSucceeded()
-              if (onSubmitSuccess) {
-                onSubmitSuccess(submitResult, dispatch, props)
-              }
-              return submitResult
-            } catch (err) {
-              throw err
+        return result.then(
+          submitResult => {
+            stopSubmit()
+            setSubmitSucceeded()
+            if (onSubmitSuccess) {
+              onSubmitSuccess(submitResult, dispatch, props)
             }
-          })
-          .catch(submitError => {
+            return submitResult
+          },
+          submitError => {
             const error =
               submitError instanceof SubmissionError
                 ? submitError.errors
                 : undefined
             stopSubmit(error)
             setSubmitFailed(...fields)
-            if (error && onSubmitFail) {
+            if (onSubmitFail) {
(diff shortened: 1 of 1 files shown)


erikras:
Makes sense.

erikras:

PR #3849, opened 2018-02-20
Add immutableProps option to Field component

### LINKED ISSUE #3810: Redux form doesn't handle recursive structures as props to Field component due to deepEqual

itsuryev:
Fixes #3810

### DIFF
--- a/docs/api/Field.md
+++ b/docs/api/Field.md
@@ -134,6 +134,12 @@ Allows you to to provide a field-level warning rule. The function is given the f
 If `true`, the rendered component will be available with the `getRenderedComponent()` method.
 Defaults to `false`. **Cannot be used if your component is a stateless function component.**
 
+#### `immutableProps : Array<String>` [optional]
+
+> Prop names that only require strict-equals, not deep equals, to determine
+> `shouldComponentUpdate`. Useful for performance and compatibility with 3rd
+> party immutable libraries. Defaults to `[]`.
+
 ## Usage
 
 The `component` prop will be passed to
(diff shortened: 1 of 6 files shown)


erikras:
Can you add a test to get code coverage back up to 100%?

itsuryev:
Updated the code, now test coverage is back to 100%. 

I might be missing something but I believe that drop in coverage was in error - any assignments in `shouldComponentUpdate` marked `onChange` for ReactNative  in `handleChange` as uncovered.

So following decreases coverage:
```js
shouldComponentUpdate(nextProps: Props) {
  const nextPropsKeys = Object.keys(nextProps)
  const thisPropsKeys = Object.keys(this.props)
  const {immutableProps = []} = nextProps

  // or even
  const hello = "world"
```

And inlining immutableProps usage works fine:
```js
if (~(nextProps.immutableProps || []).indexOf(prop)) {
```

PR #3856, opened 2018-02-22
Provide `name` to `Field` events

### LINKED ISSUE #2849: Get name of changed field in onChange
I cannot think of any way to get the name of the field being changed when onChange is fired. Is it impossible? or is there any way I can achieve the same result (get the name of the field being changed) with other methods?

kevva:
Decided not add it to the `event` because that wouldn't work in React Native.

Fixes #2849.


src/ConnectedField.types.js.flow
```diff
@@ -18,11 +18,11 @@ export type Props = {
...
   asyncError: any,
   asyncValidating: boolean,
-  onBlur: { (event: Event, newValue: ?any, previousValue: ?any): void },
-  onChange: { (event: Event, newValue: ?any, previousValue: ?any): void },
-  onDrop: { (event: Event, newValue: ?any, previousValue: ?any): void },
-  onDragStart: { (event: Event): void },
-  onFocus: { (event: Event): void },
+  onBlur: { (event: Event, newValue: ?any, previousValue: ?any, name: string): void },
+  onChange: { (event: Event, newValue: ?any, previousValue: ?any, name: string): void },
+  onDrop: { (event: Event, newValue: ?any, previousValue: ?any, name: string): void },
```
gustavohenke: Does this work well in Flow? A required argument after an optional one 🤔 

I have zero experience with Flow, only TypeScript in typed-JS world and it definitely wouldn't work 😛

src/ConnectedField.js
```diff
@@ -208,11 +213,11 @@ const createConnectedField = (structure: Structure<*, *>) => {
     }
 
     handleDragStart = (event: any) => {
-      const { onDragStart, value } = this.props
+      const { name, onDragStart, value } = this.props
```
gustavohenke: Will this name include full name, including `FormSection`s?
That's probably open to discussion! cc @erikras

kevva: Yup, it's generated and passed to the field [here](https://github.com/erikras/redux-form/blob/master/src/createField.js#L119).

PR #3874, opened 2018-02-27
Add clearFields and clearSubmitErrors to propTypes

mk0x9:
Added propTypes for `clearFields` & `clearSubmitErrors` and test case comparing declared prop types in `src/propTypes.js` and passed to decorated component


erikras:

PR #3881, opened 2018-03-01
Use shouldError/shouldWarn by default unless only shouldValidate is passed in.

FMartinez4:
When providing a custom `shouldError` prop, it needs to be passed into both `shouldError` and the deprecated `shouldValidate`.

```javascript
const customShouldError = () => { ... };
const myForm = reduxForm({
  shouldError: customShouldError,
  shouldValidate: customShouldError,
});
``` 

If only `shouldError` is passed in the `shouldValidate` may override the expected behavior.  For example the `shouldError` prop returns `false` but the default `shouldValidate` returns `true`, the code will still run the validation.

This PR will allow for either `shouldError` or `shouldValidate` to be used.


erikras:

PR #3885, opened 2018-03-02
Update age test within examples

restlessdesign:
The current form of these tests is ageist and not something that I feel ought to be joked about. It would be a shame if anyone using this library came across this example and felt excluded or discriminated against as a result! 😞

This change switches the test to be a more common example: ensuring that an individual filling out a form meets a particular legal age requirement.


oshanley:
👏 👏 👏

erikras:
Thanks for this. 👍 

Is making fun of AOL users still acceptable? 😄

PR #3907, opened 2018-03-13
Add .nvmrc

ryami333:
Adds an `.nvmrc` file which specifies a version of node (and therefore also npm) that the developer and CI environment is expected to build against, so that all of those environments are consistent. 

The version 8.10 is not arbitrary - I deliberately chose the major version '8' because it is an LTS release, of which `8.10` is the latest. It's also somewhat important to use npm ~5.6  because most of the major issues with `package-lock.json` like [this one](https://github.com/npm/npm/issues/17749) were resolved with that release, which does not come bundled with node until `8.7`. 

~[Travis CI respects](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Specifying-Node.js-versions-using-.nvmrc) `.nvmrc` if you don't explicitly specify `node_js` keys, so I removed those keys and the subsequent `npm install` script because they're now redundant as there's a source-of-truth for node/npm versioning.~

Also updated the version of npm used in Travis to 5.6 for the same reasons.

See [here](https://nodejs.org/en/download/releases/) for information about node releases.


ryami333:
Sure - I'll revert but I'll update the `npm install` line to use 5.6 instead of 5.3, for the reasons mentioned about package-lock.json. That OK?

Issue #3500, opened 2017-10-10
Field level validation must revalidate on component re-register.


theKashey:
### Are you submitting a **bug report** or a **feature request**?
It is a bug
### What is the current behavior?
If you will change validation rule (or any other prop) Field will re-register itself.
### What is the expected behavior?
It will re-register itself and continue behave as it might behave.
### What's your environment?
7.0.4
### Other information

The bug is quite simple - 
I have form with 2 fields: The A and The B.
Validation Rule is simple: B must be equal A. And this is field level validation.

Then I'v change A it will call validation of B and display "SYNC_ERROR". Just after it I will provide a new validation function for B, and Field will de-register itself, wiping all the errors, and register back.

But this point it might call validation again, but it will not.
In other words - then I am providing a new field validation function to a Field - it must be called.

HarrisGri:
I experience the same issue.

I have a field that is connected to a Country selection field. Which means, every time somebody changes a country the other field, that is connected to the Country selection will rerender and update his validators. Because depending on the selection the other field has to validate the postal_code depending on the country regulation.

**What currently happen's**
After the user selects an another payment country the field that is listening to that change is going to be rerendered and the _**validate**_ will be updated with the new _validate functions_ (redux-form will perform a @@redux-form/UNREGISTER_FIELD then @@redux-form/REGISTER_FIELD).

What happens after Unregister/Register is that all syncErrors are gone and the fields are marked as valid.

danielrob:
Sandbox https://codesandbox.io/s/20635p081y

atsu85:
Shouldn't this be closed, now that https://github.com/erikras/redux-form/pull/3909 is merged?

danielrob:
@atsu85 as soon as 7.4.0 is released.

Issue #3566, opened 2017-10-25
Re-mounting a Field component erases the field-level validation function


sandkampderek:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

When a Field with a field-level validation function attached to it is unmounted and immediately re-mounted, that validation function will never be run again.

In the example below, when the key changes, the Field will be unmounted and re-mounted, and the validation function will never be run again:

          <Field
            key={`${this.state.firstNameFieldKey}`}
            component={renderField}
            name="firstName"
            validate={this.validateFirstNameField}
          />

Perhaps a more common method of unmounting and remounting fields is conditional rendering of fields with the same name. In the example below, when this.state.shouldRenderTextArea changes, the validate function will never be run again:

        {
          this.state.shouldRenderTextArea ?
            <Field 
              name="firstName"
              component={renderTextArea}
              validate={this.validateFirstNameField}
            />
            :
            <Field
              name="firstName"
              component={renderTextInput}
              validate={this.validateFirstNameField}
            />
        }

### What is the expected behavior?
After the field is re-mounted, its validation function should run.

### Sandbox Link

https://codesandbox.io/s/m7l1lyzyoy

### What's your environment?

Reproduced in redux-form 7.1.1

### Other information

The problem appears to be that Field registers the validate function in componentWillMount, and erases the validate function in componentWillUnmount. However, React 16 will always run componentWillMount of the new component BEFORE componentWillUnmount of the old component. That means that we register the new validate function, and immediately delete it.

Would there be any problem with having Field register itself it componentDidMount instead?

marcneander:
Status on this?

marcneander:
To add some context: I get the same problem on all 7.x.x versions. Works as expected on 6.8.0

React: 15.6.2

Luchanso:
Yes, I have the same problem. It was because componentWillUnmount call after componentWillMount.

-> componentWillMount First
--- Start Remounting
-> componentWillMount Second
-> componentWillUnmount First

I think redux-form must call register function when componentDidMount. I can try make PR for this?

related links: 
* [How to work validate, register, unregister](https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L563-L593)
* [When call register and unregister](https://github.com/erikras/redux-form/blob/master/src/createField.js#L40-L73)

Luchanso:
[...]

PR #3909, opened 2018-03-13
Fix field not validating after re-mount (#3566, #3500)

viljark:
Because we do
```
return (
    !structure.deepEqual(values, nextProps && nextProps.values) ||
    !structure.deepEqual(lastFieldValidatorKeys, fieldValidatorKeys)
  )
```
in the `defaultShouldValidate`, we have to remove the field keys from the `lastFieldValidatorKeys` when Field is destroyed, otherwise on re-mount the key still exists and validation is skipped.

Should fix:
https://github.com/erikras/redux-form/issues/3500
https://github.com/erikras/redux-form/issues/3566


asaarnak:
Hei @erikras @gustavohenke!
Could take a look at this PR please?
Quite important for us.

Issue #3901, opened 2018-03-09
isAsyncValidating selector


cbarringer:
### Are you submitting a **bug report** or a **feature request**?

feature request

### What is the current behavior?

There is no selector for isAsyncValidating. This value is passed down as a prop, but having it for use in, for example, re-select would be convenient.

### What is the expected behavior?
`import { isAsyncValidating } from 'redux-form'` should provide a selector factory that takes the form name as the sole argument and returns a selector which takes the redux state and returns whether or not asynchronous validation is in progress.

cbarringer:
I have created pull request #3910 to add this enhancement

PR #3910, opened 2018-03-13
Add isAsyncValidating selector #3901

cbarringer:
Added isAsyncValidating selector to determine if the form is asynchronously validating


cbarringer:
I'm a little confused by the codeclimate error. `.codeclimate.yml` seems to indicate that `.spec.js` files should be ignored...

PR #3913, opened 2018-03-14
Add immutableProps to HOC propTypes, do not pass it to wrapped component

lgodziejewski:
`immutableProps` is described in the `ReduxForm.md` file, however it is not included in the `Config` type definition and is passed down to the wrapped component.

I think this is not intended, so I've excluded it from passing down. This is a minor change, so I haven't opened any issue

This is my first public PR ever, so please provide feedback if I have omitted some step (btw it seems the CONTRIBUTING.md file is outdated, as e.g. there is no `format` script in the `package.json` file.)

Maybe I should also add a test verifying if proper props are passed down to the `WrappedComponent`?


erikras:
Perfect. Great work! Thanks!

PR #3956, opened 2018-04-02
Use componentDidMount instead of componentWillMount to register Field (#3566)

asaarnak:
Use componentDidMount instead of deprecated componentWillMount to register Field, Fields, FieldArray.
Should fix: https://github.com/erikras/redux-form/issues/3566


asaarnak:
@erikras 
Could you take a look at this PR please?
This fixes a bug https://github.com/erikras/redux-form/issues/3566 for us.

https://reactjs.org/docs/react-component.html#unsafe_componentwillmount
"Avoid introducing any side-effects or subscriptions in this method. For those use cases, use componentDidMount() instead."

asaarnak:
Seems that @erikras is busy @gustavohenke could anyone take a look at this PR?

asaarnak:
@gustavohenke @erikras 
I added a test for https://github.com/erikras/redux-form/issues/3566#issuecomment-377955540
The test fails on master branch but succeeds on this PR.

eliseumds:
This will break SSR, no?

asaarnak:
@eliseumds
What SSR means?
Should we write some new test?

erikras:
SSR = Server Side Rendering

`componentWillMount` happens on the server during SSR, but `componentDidMount` does not.

asaarnak:
https://github.com/reactjs/reactjs.org/issues/727#issuecomment-376857584
From this tweet it seems that we should use `constructor` for SSR.
But we could also keep `UNSAFE_componentWillMount` in short term, but it would not be safe for async rendering.
In longer term they plan to introduce new api called `suspense` for data fetching.

asaarnak:
https://hackernoon.com/problematic-react-lifecycle-methods-are-going-away-in-react-17-4216acc7d58b
`Async rendering will cause componentWillMount to trigger multiple rendering of your component tree. This makes it unsafe.`

asaarnak:
@erikras I tried using `constructor` instead of `componentDidMount` to register Field but then the test fails because component is reregistered before previous component is unregistered:
```jest
Expected value to be:
  "Max allowed weight is 50"
Received:
  "Max allowed weight is 20"
```
Any ideas how to fix this test?

asaarnak:
@erikras Interesting that these 2 new tests i added in this PR fail on my localhost...

asaarnak:
@erikras any news on this PR ?

RichardJuchelka:
Hi @erikras 
Any progress on this one?

erikras:
I just kicked the tires on this via server-side rendering with Next.js and it seems to work just fine.

asaarnak:
You requested PR change to use `constructor` instead of `componentDidMount` to support SSR for field registration.
So tests created by this PR fail after this change.
Moving field register into `constructor` we are at the same state as before with deprecated `componenWillMount`, but no more deprecated method and issue still exists: https://github.com/erikras/redux-form/pull/3956/commits/82781c08e00b1e3bfc462353dad9e2d9490a0cee

asaarnak:
Currently we used patch-package to patch our live version with `componentDidMount` version and so far no issues.

src/__tests__/Field.spec.js
```diff
@@ -2210,6 +2210,88 @@ const describeField = (name, structure, combineReducers, setup) => {
...
+          }
+        }
+        
+        render() {
+          // This will trigger unregister and register Field
+          // and these must happen in correct order for Field validation to work
+          const someCrazyBusinessLogic = this.state.weightLimit > 30
+          return (
+            <div>
+              {someCrazyBusinessLogic &&
```
asaarnak: 1. Here we register field named 'weight'.

src/__tests__/Field.spec.js
```diff
@@ -2210,6 +2210,88 @@ const describeField = (name, structure, combineReducers, setup) => {
...
+              }
+              {!someCrazyBusinessLogic &&
+                <Field
+                  name={weightFieldName}
+                  component={renderWeight}
+                  validate={this.state.validate}
+                />
+              }
+              <button onClick={() => this.setState({ 
+                weightLimit: nextWeightLimit,
```
asaarnak: 2. Here we update state and set new weight limit.
Which will update the validate function as well.

src/__tests__/Field.spec.js
```diff
@@ -2210,6 +2210,88 @@ const describeField = (name, structure, combineReducers, setup) => {
...
+            <div>
+              {someCrazyBusinessLogic &&
+                <Field
+                  name={weightFieldName}
+                  component={renderWeight}
+                  validate={this.state.validate}
+                />
+              }
+              {!someCrazyBusinessLogic &&
+                <Field
```
asaarnak: @erikras See comments 1. 2. and 3. on test.
3. Based on state(internal state or new props) Field is recreated into other place in code.
And calling componentDidUnmount for 1.(prev field) after this field(new field) constructor or componentWillMount.
When using componentDidMount for this(new field) it's called after 1.(prev field) componentDidUnmount.
Which fixes this issue when Field is moved around with crazy business logic and that recreates Field validate function in wrong order.

asaarnak:
@erikras This test should succeed to fix https://github.com/erikras/redux-form/issues/3566#issuecomment-377955540
Currently fails in jenkins, displays old validation message when new validate prop is changed:
bug-validate-prop-changed-but-validation-message-is-still-old

asaarnak:
Updated to latest master and reverted this commit to fix tests: https://github.com/erikras/redux-form/pull/3956/commits/82781c08e00b1e3bfc462353dad9e2d9490a0cee

PR #3960, opened 2018-04-04
Fix onSubmit documentation for reduxForm configuration

charlax:
`onSubmit` is actually only passed the values.


gustavohenke:
This is incorrect. See https://codesandbox.io/s/yjl3q0oo1v

PR #4010, opened 2018-04-27
Use a cached object for syncErrors and syncWarnings

sf-old:
It appears that https://github.com/erikras/redux-form/pull/3201 introduced a regression where a prop that was cached to the same value when falling back (https://github.com/erikras/redux-form/pull/3072) now creates a new object each render. This can cause a rendering performance penalty hit due to preventing shallow prop comparisons.

This PR fixes the regression by always using the `empty` property from the plain structure, rather than from the structure being passed as an argument to `createReduxForm` (which I believe was the intended fix in #3201).


erikras:

PR #4014, opened 2018-05-02
Fix clear errors in parent of array after child change

### LINKED ISSUE #3580: SubmissionError with FieldArray remove only after add new field
I have a problem with SubmissionError with FieldArray. I add two the same values to two Fields in FieldArray and send them to API. Then API send me response: "Values not unique" and I throw SubmissionError and set values._error.. So I change one of value, or remove one of value and click Send one more time. But error doesn't disappear after this changes and I can't send request to API.
Error disappears only after add new value to FieldArray.

tomasztomys:
Fixes #3580


tomasztomys:
@erikras

erikras:

tomasztomys:
@erikras are you sure? You revert this changes

Issue #3729, opened 2018-01-04
Bug: submit -> submitSuccess -> reinitialize -> form.pristine == true => *edit_form* -> form.pristine == TRUE


rasdaniil:
A Possible Bug with global form pristine field

### What is the current behavior?
Given reduxForm with
1) enableReinitialize = true
2) keepDirtyOnReinitialize = false (default)
2) onSubmitSuccess(res, dispatch, props) {
  dispatch(initialize(props.form.name, props.form.values)
}

After the form was successfully submitted, it reinitializes and pristine becomes true again (very good).
However, after I make any alteration to the form pristine keeps staying true.

Here is what goes on in the debugger:

```
// createReduxForm.js: 929
const shouldUpdateInitialValues = enableReinitialize (true) && initialized(true) && deepEqual(initialValues, stateInitial) (deepEqual returns false)
// true
```
A combination of `shouldUpdateInitialValues == true` and `keepDirtyOnReinitialize == false` always produces `pristine = true` a few lines below. 

If I set `keepDirtyOnReinitialize` to true, the pristine prop gets evaluated correctly.

I wasn't able to grasp from the documentation how `keepDirtyOnReinitialize` affects global pristine, because it only refers to fields' pristine.

Thanks for the awesome piece of work.

### What is the expected behavior?

I expect pristine to become false after form data is changed.

dmitrykrylov:
I have the same issue. It's a piece of my code: 
```
Form = reduxForm({
    onSubmitSuccess: (result, dispatch, props) => {
        dispatch(initialize(
            props.form,
            props.values,
            { keepSubmitSucceeded: true }
        ));
    }
})(Form);
```

And after `SET_SUBMIT_SUCCEEDED => INITIALIZE  => CHANGE` I have `pristine === true`

iamdanthedev:
@dmitrykrylov 
Hello! I don't  remember really, I suppose I would have posted it here. Could you try setting `keepDirtyOnReinitialize` to true? Maybe that was the thing.. Thanks

dmitrykrylov:
@iamdanthedev I actually want to have pristine === true after initialize and before the first change.

Here is what I want to get:
1. Initialize form, show no messages, pristine === true
2. Make a change, show "you have unsaved changes" message, pristine === false 
3. Submit form, show "submit succeeded" message, pristine === true
4. Make a change, show "you have unsaved changes" message, pristine === false

iamdanthedev:
[...]

PR #4020, opened 2018-05-08
Fix form reinitialize problem: using previous initialValues

zmefz:
# Problem:
On reinitialize form it calls `onChange` method with previous initial values.

# Workaround:
[Use `keepDirtyOnReinitialize={true}`](https://github.com/erikras/redux-form/issues/3729)

# Code problem explanation:
When initial values changed and reinitialize enabled:
```
// enableReinitialize(true) && initialized(true) && (!deepEqual(initialValues, stateInitial)) (true)
const shouldUpdateInitialValues = enableReinitialize && initialized && !deepEqual(initialValues, stateInitial)
// shouldUpdateInitialValues(true)

// initialValues (new initial values), stateInitial (previous initial values)
let initial = initialValues || stateInitial || empty
// initial (new initial values)

// if we should update initial values
if (shouldUpdateInitialValues) {
        // we set old initial values as current values
	initial = stateInitial || empty
}

// here at initial we have previous value instead of current.
```

# Solution
`if (**!**shouldUpdateInitialValues) {`
It means if we shouldn't update initial values we will set at `initial` old initial values or `empty`

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -986,7 +986,7 @@ const createReduxForm = (structure: Structure<*, *>) => {
 
           let initial = initialValues || stateInitial || empty
 
-          if (shouldUpdateInitialValues) {
+          if (!shouldUpdateInitialValues) {
             initial = stateInitial || empty
           }
 


erikras:

Issue #2056, opened 2016-11-01
Async validation strange behaviour


afitiskin:
I had a very strange bug on my project when I used async validation for fields. During form filling everything works without any issues, but when I submit the form I got `Uncaught (in promise)` error. I checked every line of my code and found that if I resolve async validating promise with some data `redux-form` treat it as an error and fail form submission.

As I see, the problem is here:
```javascript
const handleErrors = rejected => errors => {
  if (errors && Object.keys(errors).length) {
    stop(errors)
    return errors
  } else if (rejected) {
    stop()
    throw new Error('Asynchronous validation promise was rejected without errors.')
  }
  stop()
  return Promise.resolve()
}
```
First `if` statements will not be not passed even if `rejected` is `false`. I don't know if it is a feature of a bug?

I forked the repo and added [example of the problem](https://github.com/afitiskin/redux-form/blob/broken-async-validation-example/examples/brokenAsyncValidation/src/asyncValidate.js#L10).
Form works until you submit it with valid values, you will get error in the console.

gustavohenke:
Closing due to inactivity.

In case you still have problems, please open a formal bug report/feature request/docs improvement request.
Otherwise, please ask for help at StackOverflow.

erikras:
Fix released in [`v7.4.0`](https://github.com/erikras/redux-form/releases/tag/v7.4.0).

PR #4028, opened 2018-05-16
Async validation improvements

### LINKED ISSUE #3895: Async validation fails if promise resolved with some data
We have a form with async validation (same as in [Async Blur Validation Example](https://redux-form.com/7.3.0/examples/asyncvalidation/)) 

We have a following validation function:
```javascript
const asyncValidate = (values/*, dispatch */) => {
  return sleep(1000)
    .then(() => {
        if ([ 'john', 'paul', 'george', 'ringo' ].includes(values.username)) {
          reject({ username: 'That username is taken' })
        } else {
          resolve('You shall not pass!')
        }
      })
    })
}
```
In case of success validation, promise is resolved with some data (in our case just string `You shall not pass!`).

Async validation works just fine, but submitting of the form is impossible. 
Error in console after submit: `(index):1 Uncaught (in promise) You shall not pass!` 

Async validation returns promise, if promise is resolved (no matter with or without data provided) – async validation should pass and form submission should work as usual.

### Sandbox Link
Live example: https://redux-form-broken-validation.herokuapp.com/
Live example source code: https://github.com/afitiskin/redux-form/tree/broken-async-validation-example/examples/brokenAsyncValidation

### What's your environment?
Latest versions of packages and browsers

It is a duplicate of #2056 with formal bug report.

afitiskin:
Fixes #3895 (#2056)

### DIFF
--- a/src/asyncValidation.js
+++ b/src/asyncValidation.js
@@ -19,14 +19,16 @@ const asyncValidation = (
     )
   }
   const handleErrors = rejected => errors => {
-    if (errors && Object.keys(errors).length) {
-      stop(errors)
-      return errors
-    } else if (rejected) {
-      stop()
-      throw new Error(
-        'Asynchronous validation promise was rejected without errors.'
-      )
+    if (rejected) {
+      if (errors && Object.keys(errors).length) {
+        stop(errors)
+        return errors
+      } else {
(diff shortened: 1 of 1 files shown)


erikras:

PR #4040, opened 2018-05-23
Export action `updateSyncErrors`.

dwick:
Same issue as https://github.com/erikras/redux-form/pull/3631, but a different action.


erikras:

PR #3995, opened 2018-04-21
Added proptype Proptypes.node to FormSection's component propTypes

anu1097:
I have changed the propTypes of components props in FormSection so that the following bug can be resolved.
>FormSection propType doesn't allow React 16 Fragment #3859 

I want to contribute further. Looking forward to your comments and feedback. Do tell me if I have missed anything or code of conduct I needed to follow. Thank you.

PR #4041, opened 2018-05-23
Accept PropTypes.node for component props

cherewaty:
## Purpose
I've created a higher-order component that makes it easy for custom form element components to consume `redux-form`'s `input` and `meta` props. I realized that for a couple of the form elements, I need to use [`React.forwardRef()`](https://reactjs.org/docs/forwarding-refs.html) in the HOC.

The ref-forwarding works as expected, but I get a couple prop type errors:
```
ERROR: 'Warning: Failed prop type: Invalid prop `component` supplied to `Field`.
ERROR: 'Warning: Failed prop type: Invalid prop `component` supplied to `ConnectedField`.
```

`React.forwardRef()` returns an object, not a function. `Field`'s `component` prop type is either `function` or `string`.

## Approach
I borrowed the approach from https://github.com/erikras/redux-form/pull/3995. Now the `component` prop can be a `node` in `ConnectedField`, `ConnectedFieldArray`, `ConnectedFields`, `createField`, `createFieldArray`, and `createFields`.

## Alternate options
- `PropTypes.element`?
- `React.isValidElementType()`?

## Extra info
Looks like `prettier` reformatted a few unrelated things.

### Related
- [`typeof React.forwardRef(...)` is not a 'function', breaking third-party libs](https://github.com/facebook/react/issues/12453)
- [Add ReactIs.isValidElementType()](https://github.com/facebook/react/pull/12483)
- [New PropType to check if a prop is a valid React Component](https://github.com/facebook/react/issues/9186)

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -296,14 +296,20 @@ const createConnectedField = (structure: Structure<*, *>) => {
   }
 
   ConnectedField.propTypes = {
-    component: PropTypes.oneOfType([PropTypes.func, PropTypes.string])
-      .isRequired,
+    component: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.string,
+      PropTypes.node
+    ]).isRequired,
     props: PropTypes.object
   }
 
   const connector = connect(
     (state, ownProps) => {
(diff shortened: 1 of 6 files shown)


erikras:

PR #4051, opened 2018-06-01
Upgraded a ton of dependencies (e.g. webpack 4)

erikras:


erikras:
Ugh. Code Climate exclude pattern isn't working. 😢

Issue #2870, opened 2017-04-30
arraySplice not exposed on FieldArray fields prop


jacobk:
All of the other array action creators are exposed as bound functions on the `fields` prop passed to `FieldArray`.

Looking at https://github.com/erikras/redux-form/blob/master/src/createFieldArrayProps.js#L4 it seems like it might be deliberate since the eslint error is suppressed. If that is the case I think it should be documented why.

I need it for the purpose of replacing an item in the field array that is currently edited outside of the rendered form (an `update(index, value)` would be a nice option if it's a common enough use case).

DongCai008:
it's true. really want to splice can be supported in FieldArray. and the change is really tiny.

diff --git a/src/createFieldArrayProps.js b/src/createFieldArrayProps.js
index 481edec..3a015c1 100644
--- a/src/createFieldArrayProps.js
+++ b/src/createFieldArrayProps.js
@@ -33,6 +33,7 @@ const createFieldArrayProps = (getIn, name, form, sectionPrefix, getValue,
         arrayShift()
         return getIn(value, 0)
       },
+      splice: arraySplice,
       swap: arraySwap,
       unshift: arrayUnshift
     },

bmv437:
You should submit a PR 😄

danielrob:
@erikras had a look at the blame history, but you did all the action creators at once. Could you enlighten as to why arraySplice was marked as advanced usage and not handed to FieldArray?

gustavohenke:
Because this issue saw almost no activity for a few months, I'm closing it.

brneto:
This fields prop is missing in the documentation, isn’t it?

Issue #3244, opened 2017-07-25
ArraySplice: accept multiple replacement values


danielrob:
Presently `ArraySplice` only accepts one replacement value, unlike it's native equivalent `Array.splice`

Example: 
```
arraySplice('form', 'field', 0, field.length, "bob", "alice", "mary") should add items "bob", "alice", and "mary" into the array, however only "bob" is added.
```

This issue has previously been tracked (#203), but I'm raising this as a clear problem statement, as this seems a sensible enhancement, for someone, at some time, to take on.

This enhancement is to align the action with `Array.splice` by allowing multiple replacement values to be added to the array in the spliced location.

danielrob:
Related: #2870 would like arraySplice exposed as a fields prop.

PR #4065, opened 2018-06-11
Add missing splice() to FieldArray's fields prop

soulmerge:
The FieldArray's fields prop was missing the splice() function. This was once discussed in #2870.


danielrob:
Me too. For bonus points: #3244 :).

Issue #4072, opened 2018-06-13
Fields are not getting registered in React 15


zwug:
### Are you submitting a **bug report** or a **feature request**?
Bug report

### What is the current behavior?

With react 15 there are no actions `@@redux-form/REGISTER_FIELD`. That leads to errors with some form properties like [valid](https://redux-form.com/7.4.0/docs/api/props.md/#-code-valid-boolean-code-) because [validation checks](https://github.com/erikras/redux-form/blob/master/src/selectors/isValid.js#L37) only registered fields.

### What is the expected behavior?

I've made some research and found that the reason of not dispatching necessary actions is this [refactoring](https://github.com/erikras/redux-form/pull/4051/files#diff-ca0d4d55e50da37455f88ffb85da436bR40) of lifecycle methods. Seems like all the old React lifecycle methods are not calling methods they should.
Since redux-form still supports react 15 there should be a workaround to use `componentWillMount` to register fields.

### Sandbox Link
https://codesandbox.io/s/j4lx76v98v
No `@@redux-form/REGISTER_FIELD` actions here

### What's your environment?

redux-form: 7.4.0
react: 15.6.2

vladhd93:
also i have a lot of bugs with 7.4.0, in general form init and validators not workingat all (react 15.6.2 and redux-form 7.4.0)

JoshMock:
I believe this is a duplicate issue to #4069, or at least has the same root cause.

danielrob:
Agreed, this is almost certainly a duplicate of #4069. Thank you for the well reported issue.

erikras:
Published fix in [`v4.7.1`](https://github.com/erikras/redux-form/releases/tag/v7.4.1). Please confirm.

erikras:
Published _ACTUAL_ fix in [`v7.4.2`](https://github.com/erikras/redux-form/releases/tag/v7.4.2).

Issue #4070, opened 2018-06-12
[v7.4.0] Field-level validation not working


patrikmolsson:
### Are you submitting a **bug report** or a **feature request**?

Bug report

### What is the current behavior?

I forked the sandbox presented here: https://redux-form.com/7.4.0/examples/fieldlevelvalidation/ and updated the redux-form version, please see https://codesandbox.io/s/vmvkqnox2y.

In the new update field-level validation functions are not triggered.

### What is the expected behavior?

The validation functions are supposed to trigger.

### Sandbox Link

https://codesandbox.io/s/vmvkqnox2y

### What's your environment?

Chrome 65.0.3325.181, 
redux-forms 7.4.0

shamanking:
I have some issue!

h0wXD:
Me too!

gustavohenke:
Okay folks, no need to comment like this, please just react with 👍 to @patrikmolsson's opening post and it's enough for us to know how many of you have been affected.

danielrob:
Please see #4069, the root cause is the same which is the lifecycle methods were updated inappropriately to their UNSAFE counterparts. Will leave this open for now for this specific symptom to be tested after the fix.

erikras:
Published fix in [`v4.7.1`](https://github.com/erikras/redux-form/releases/tag/v7.4.1). Please confirm.

erikras:
NOT fixed. Looking into it....

erikras:
Published _ACTUAL_ fix in [`v7.4.2`](https://github.com/erikras/redux-form/releases/tag/v7.4.2).

PR #4078, opened 2018-06-15
Hotfix: Revert all UNSAFE_ lifecycle methods

### LINKED ISSUE #4069: [v7.4.0] initialValues not working
bug report

initialValues prop is not working on initial mount after update to 7.4.0

### Sandbox Link
I've modified the **Initialize from state** demo just to show the issue. I've just change:
```
 state => ({
    initialValues: data
 }),
```
to get values directly from the variable instead after clicking the button.

https://codesandbox.io/s/ox3p66023q

As you can see the fields are empty. If you change the redux-form version to 7.3.0 it works.

### What's your environment?
redux-form 7.4.0

danielrob:
The addition of the `UNSAFE_` lifecycle methods broke React 15 compatibility. Those methods are not intended for libraries still supporting React 15. 

The only way to remove the React 16.4+ `StrictMode` warnings about `cWM`, `cWRP`, and `cWU` in a React 15 compatible way is to migrate these methods to the new lifecycle methods and use `react-lifecycles-compat` for backwards compatibility. The `UNSAFE_` methods were to support application authors with a known target version of React (16.3+) from having to migrate every component before upgrading. Unfortunately we can't do that trick. 

#### Additional Testing
Since the automated tests are all run against React 16, I've confirmed this locally by building my fork, and copying the `lib` folder to `node_modules/redux-form` of a download of [this sandbox](https://codesandbox.io/s/r0o9q33lmp) which is against 15.6, and confirming that the initialValues (as per #4069) are now showing. That's the only test I've run.  

#### Should fix
#4069, #4070, #4072

### DIFF
--- a/src/Form.js
+++ b/src/Form.js
@@ -19,7 +19,7 @@ class Form extends Component<Props> {
     }
   }
 
-  UNSAFE_componentWillMount() {
+  componentWillMount() {
     this.context._reduxForm.registerInnerOnSubmit(this.props.onSubmit)
   }
 
--- a/src/createField.js
+++ b/src/createField.js
@@ -37,7 +37,7 @@ const createField = (structure: Structure<*, *>) => {
       return shallowCompare(this, nextProps, nextState)
     }
 
-    UNSAFE_componentWillMount() {
+    componentWillMount() {
       this.context._reduxForm.register(
         this.name,
         'Field',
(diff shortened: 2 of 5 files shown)


erikras:
This is because I'm an idiot and didn't use the polyfill properly. I made the same mistake in another library.

erikras:
I'll fix this immediately.

erikras:
Published fix in [`v4.7.1`](https://github.com/erikras/redux-form/releases/tag/v7.4.1).

danielrob:
Sorry, I should have looked more deeply into the polyfill, they didn't mention on the front page it also supported the `UNSAFE_` methods... I should have dug deeper! Thanks Erik!

danielrob:
Wait... it doesn't...

erikras:
Misunderstanding on my part, apparently. 🤦‍♂️

erikras:
Published in [`v7.4.2`](https://github.com/erikras/redux-form/releases/tag/v7.4.2).

Issue #3637, opened 2017-11-25
FieldArray Field dynamic change function does not validate field


rntorm:
I have a case where I'm trying to empty selected value in dropdown when another dropdown is changed. I manage to do it with change function.

```
if(order.thickness) {
    const indexOf = thicknessOptions.indexOf(order.thickness.value);
    if(indexOf === -1) {
        change(`orders[${index}].thickness`, null);
    }
}
```

The problem is that the changed field does not show error after that and allows submit, allthough field has required validation. 
Any ideas how to trigger validation on the changed field? I did not find anything that works. Help is much appreciated!

Issue #3288, opened 2017-08-04
Cannot use a higher-order function in Field Validation


duro:
### Are you submitting a **bug report** or a **feature request**?
Bug

### What is the current behavior?
When using field level validators, I am unable to use a higher order function to produce the validation function

This works:
```
const required = value => (value ? undefined : 'Required')

<Field
  label="Email"
  name="email"
  component={InputField}
  validate={required}
  placeholder="email"
  type="text"
/>
```

This DOES NOT work:
```
const required = (msg = 'Default Message') => value => (value ? undefined : msg)

<Field
  label="Email"
  name="email"
  component={InputField}
  validate={required('Message Override')}
  placeholder="email"
  type="text"
/>
```

### What is the expected behavior?
I would expect to pass into the field validator a returned function that matches the validator interface even if it was produced by a higher order function

### Sandbox Link
https://codesandbox.io/s/G6ARklAMr

### What's your environment?
Redux Form: 7.0.3
OS: macOS
Browser: Chrome 60

duro:
Just reverted back to 6.8.0 and this works as expected.

danielrob:
This is expected behaviour see #3271.

duro:
@danielrob Cool. I followed through on the threads and updated docs, but I'll be honest. This is still a very confusing change to me. I guess I need to switch to using a bound method instead of a lambda created at render time, but I'm still not really sure.

danielrob:
I fully admit my documentation update for this was minimal... and that while this change makes the library more powerful, it also makes the consumption more advanced. 

Your help would be appreciated to alleviate this... perhaps you could think about where in the documentation / how this could be explained to make it clear? 

The key point is, that the function should not change each render, unless 

Consider: 
```
const z = x => y => console.log(y)
z(1) === z(1) // false (e.g. LHS is render 1, RHS is render 2)
```
However: 
```
import { memoize } from 'lodash'
var z = y => x => (console.log(x))
var memoized = _.memoize(z);
memoized(1) === memoized(1) // true  (e.g. LHS is render 1, RHS is render 2). 
memoized(1) === memoized(2) // false, and yes, we actually want the field to re-register if this changes, because the validation function has actually changed it's behaviour. 
```

Or, as you say, instead of memoization pre-preparing the validate function as an instance variable based on props in the class is another option.

gabrielhpugliese:
Could you please enlighten me here? What's the cause for this not to work? I didn't get it yet.

[...]

PR #4081, opened 2018-06-15
FieldArray should be re-registered (like Field is) when validate or warn props are changed (#3637)

viljark:
At the moment the validate/warn functions are not re-registered when changing the references of the functions. 
This PR enables us to create new validation function when some validation conditions change (for example validation fn depends on other input's value).
Fixes https://github.com/erikras/redux-form/issues/3637


viljark:
Not sure if the codeclimate result has anything to do with the changed code.

danielrob:
FYI https://github.com/erikras/redux-form/issues/3288#issuecomment-320729482 - there might be a better way to do it if you're intrigued.

Issue #4094, opened 2018-06-22
redux-form/immutable handleSubmit function breaks react-native apps


dapids:
### What is the current behavior?
- I have a `redux-form/immutable` form with **async validation** in a react-native app
- I submit the form
- The `handleSubmit` function tries to merge `syncErrors` and `asyncErrors`. Unfortunately the former is a plain object and the latter is an immutable Map.
- The app crashes

### What is the expected behavior?
- The `handleSubmit` function should properly handle the merge

### What's your environment?
- redux-form version: `7.3.0`
- react-native version: `0.55.4`
- OS affected: `iOS`, `android`

### Other information
Here is the stacktrace.
![image](https://user-images.githubusercontent.com/1473240/41775644-0b9d3b64-7625-11e8-9361-5f9966b8a74a.png)

dapids:
Probably related to https://github.com/erikras/redux-form/issues/3532.

erikras:
Published fix in [`v8.0.0`](https://github.com/erikras/redux-form/releases/tag/v8.0.0).

PR #4095, opened 2018-06-22
Fix handleSubmit with redux-form/immutable

dapids:
Fixes https://github.com/erikras/redux-form/issues/4094.


filipposarzana:
Hi @erikras did you have a chance to look into this?
We have the same problem and this is a great fix 👍

dapids:
@jaridmargolin and @erikras, I've create a new PR that uses duck typing instead of depending on Immutable.

However I'm open to other solutions if you don't like this implementation. :relaxed:

erikras:

PR #4097, opened 2018-06-22
Fix FormProps#asyncValidating type

mike1808:
Change type of asyncValidating from `boolean` to `boolean | string`


erikras:

PR #4103, opened 2018-06-24
Validate component props with isValidElementType()

cherewaty:
## Purpose
Continuation of https://github.com/erikras/redux-form/pull/4041

I had added `PropTypes.node` to the allowed prop types for `ConnectedField`, `ConnectedFieldArray`, `ConnectedFields`, `createField`, `createFieldArray`, and `createFields`, but that didn't solve the original problem: the value returned by `React.forwardRef()` can be passed to `React.createElement()`, but it's just an object.

I get this output:
```
ERROR: 'Warning: Failed prop type: Invalid prop `component` supplied to `Field`.
ERROR: 'Warning: Failed prop type: Invalid prop `component` supplied to `ConnectedField`.
```

## Approach
Use `isValidElementType()` from `react-is` to check the `component` prop. `isValidElementType()` was created to deal with the much more diverse set of things React can now treat as renderable (`Fragment`, `forwardRef`, etc).

I wish there was a pre-defined prop type for this use case (https://github.com/facebook/react/issues/9186), but for now a custom validator that calls `isValidElementType()` seems like the best practice.

## Alternative
I also tried just adding `PropTypes.element`, but the value returned by `React.forwardRef()` still failed that check. It only validates against `PropTypes.object`, but opening up the `component` prop's validation that wide seemed to defeat the purpose of type checking.

## Related
- [`typeof React.forwardRef(...)` is not a 'function', breaking third-party libs](https://github.com/facebook/react/issues/12453)
- [Add ReactIs.isValidElementType()](https://github.com/facebook/react/pull/12483)
- [New PropType to check if a prop is a valid React Component](https://github.com/facebook/react/issues/9186)


erikras:

PR #4117, opened 2018-07-04
[Feature] QueryField component

EmaSuriano:
### Objective

Allows retrieving the information of a field by its name.

### Motivation

I know that there is a selector in `redux-form` called `formValueSelector` which can retrieve the value of a field, but I always find annoying to connect a component/screen in order to access the value of a field. This is an example of what I'm saying (let's assume that I've already called `reduxForm()` in a component from above): 

``` javascript
import React from 'react'
import { Field, formValueSelector } from 'redux-form'
import { connect } from 'react-redux'

const MyComponent = ({ firstNameValue }) => {
  return (
    <div>
      <Field name="firstName" component="input" />
      <p>{firstNameValue}</p>
    </div>
  )
}

const mapStateToProps = state => ({
  firstNameValue: formValueSelector('MY_FORM')(state, 'firstName')
})

export default connect(mapStateToProps)(MyComponent)
```

Now, let's see the same result but in this case using `QueryField`

``` javascript
import React from 'react'
import { Field, QueryField } from 'redux-form'

const MyComponent = () => {
  return (
    <div>
      <Field name="firstName" component="input" />
      <QueryField name="firstName">{({ value }) => <p>{value}</p>}</QueryField>
    </div>
  )
}

export default MyComponent
```

### How it works?

The QueryField would connect to the redux store for us, and call our renderProp function which could be passed inside `children` or `render` prop.

``` javascript 
import React from 'react'
import { QueryField } from 'redux-form'

const WithChildren = () => (
  <QueryField name="firstName">{({ value }) => <p>{value}</p>}</QueryField>
)

const WithRender = () => (
  <QueryField name="firstName" render={({ value }) => <p>{value}</p>} />
)
```

Apart from the value of the Field, we can take all the relevant information of the field like `submitting`, `syncError`, `asyncError` and many more (check for ConnectedQueryField to see all the values the renderProp will receive), this an advantage over the `formValueSelector` because we can access to more information of the field.

### Last Words

This is my first PR to `redux-form` project so maybe there might some things I skipped because I don't know the codebase completely, for that reason please feel free to request for modifications that I didn't see or a change in the API of the component.


danielrob:
@EmaSuriano thank you for contributing to redux-form! I <3 this idea, but upon first lookover, I wonder if we can simplify it implementation wise. Just to give some context, it has been previously passed around that the `<Field />` component itself can be used as a selector, i.e., there's no rule saying that it _has_ to render a form input. However, what I really like here is that you're making this usage an explicit pattern, and introducing a render-props api to this. Regarding implementation, (and as I like to point out I mainly maintain issues around here) is there any reason not to re-use `Field` in this manner to implement this - could just save a little footprint wise..: 

```javascript
const RenderQueryField = ({ 
  render, 
  children, 
  input: { onChange, onBlur, onFocus, ...input }, 
  meta 
}) => {
  invariant(render || children, 'render or child prop is required')
  return (render || children)({ input, meta })
}

export default = ({ name, ...props }) => (
  <Field
     component={RenderQueryField}
     name={name}
     ...props // e.g.`type="checkbox"` modifiers
  />
)
```

EmaSuriano:
@danielrob Thanks for the feedback. Using the `Field` component as a selector was my first approach! But I realized that it wasn't possible to do that, because of an error that crashes my application. The error was "Cannot read property 'getHostNode' of null".

I've just made a simple example with your code to see if I'm able to reproduce it again and it seems that it's working. I really don't know why I wasn't able to get it on the first try ...

Taking into consideration that, I think it would be helpful to write an example of using the Field as a selector inside the redux-form examples (there is one example of getting the value of a field but using the `formValueSelector`). I think the best approach is to close this PR and start a new one with the example. What do you think?

EmaSuriano:
I've been testing using the Field component as a selector, and there are some cases when it gets a little bit annoying. For example, let's assume I want to take 2 values from a form, join them and display a result. I would end up with something like this.

``` javascript
      <Field
        name="firstName"
        component={({ input: { value: firstName } }) => (
          <Field
            name="lastName"
            component={({ input: { value: lastName } }) => (
              <p>
                {firstName} {lastName}
              </p>
            )}
          />
        )}
      />
```

I think in this case QueryField could be useful if I provide a change in the API, resulting in:
``` javascript
<QueryField names={["firstName", "lastName"]}>
        {({ firstName, lastName }) => (
          <p>
            {firstName} {lastName}
          </p>
        )}
      </QueryField>
```

danielrob:
@EmaSuriano couldn't you have an accompanying `QueryFields` component that wraps the `Fields` component in the same manner? I really appreciate this discussion, and to be clear, I really support having an explicit `QueryField(s)` component as you suggest rather than just adding the example to the docs (although a good option). One thing I would say though is that I'd like the render function to return the meta as well as the value, the resulting api should still be palatable.

EmaSuriano:
@danielrob I really like your approach of re-using the `Field` and `Fields` component, but I added a little improvement: I re-use `ConnectedField` and `ConnectedFields`. The reason is that Field(s) implement the logic of registering and unregistering the Field inside the lifeCycle (CDM, CDU), which in my opinion don't have anything to do with QueryField and also add unnecesary traffic to the Redux store (because it's already registered).

So basically QueryField(s) are a wrapper of ConnectedField(s) 👍 

The final API of both components are:
### QueryField
``` javascript
<QueryField name="firstName">
  {({ value }) => {
    console.log('meta of firstName', meta);
    return <p>{value}</p>;
  }}
</QueryField>;
```
### QueryFields
``` javascript
<QueryFields names={['firstName', 'lastName']}>
  {({ firstName, lastName }) => {
    console.log("fields' meta", firstName.meta, lastName.meta);
    return <p>{`${firstName.value} ${lastName.value}`}</p>;
  }}
</QueryFields>;
```

I would like to know your thoughts of the changes! And one more thing, I'm thinking of performing a new validation inside the `constructor` of each component that checks if the provided name exists in the form. Right now if the field doesn't exists the returned object is an empty object, which is a bit odd.

danielrob:
I think this looks really great. Regarding `removeFieldHandler`, we should probably actually keep `[input.]checked` and `[input.]name` first one for obvious reasons, and second I can imagine use cases (e.g. looping with object.entries). Going further, since the whole intention is that this wouldn't render a field, I'm not sure there's actually a reason to separate meta and input anymore. How about we spread so that the props are: 

```
// Field
{ value, name, checked, {...meta} } 

// Fields
{ 
  fieldName:  { value, name, checked, {...meta} }
}
```

So, if you could: 
- [ ] spread `{...meta}` and include `value`, `name`, `checked` 
- [ ] add the tests you mentioned, 
- [ ] pluralize `removeFieldHandler` => `removeFieldHandlers`
- [ ] add docs (if you need help with that, I can)

Then you'd likely have final review/signoff from me and then over to Erik the next time he does his maintenance run. Hopefully he's happy / likes it too :). 

Thanks so much for this, I think this will make a huge difference as 'how do I access X outside of the Field' is a frequently asked question.

EmaSuriano:
Hey, @danielrob did you have some spare time to check this PR? I can help with the documentation if you want, I don't know if you already wrote something 😆

eliseumds:
What about naming it *FieldValue* instead? Or something even more generic because we might want to select the value of an entire section (profileData, paymentData, addressData, for ex).

It'd be good if users didn't need to pass the section name. We could make it explicit via a prop.

thomasbertet:
@danielrob & @EmaSuriano : What can we do about this PR to make it progress ? 
I think of many use cases where this is usefull, and just would like to know if you are still planning on adding this feature. Thanks for your work !

PR #4118, opened 2018-07-04
Feature/prefixname empty name

SBRK:
I found myself wanting a Field which, inside a `FormSection` would use the name passed to the `FormSection`. For that, I tried passing `name=''`, to no avail.

This is useful if, like in my case, you want a text field to edit your whole section as JSON, while also having other fields editing different fields of the section:

 ```
<FormSection name='foo'>
  <JSONField name=''/> // Will edit foo as json
  <InputField name='title'/> // Will edit foo.title
  [...]
</FormSection>
```


danielrob:
I'd worry about the `0` case. Not that numerical names particularly work anyway, but I think some people have used them to e.g. access array indexes. Might be good to explicitly check `name === ''` just in case.

SBRK:
@danielrob @gustavohenke I added the tests and handled the case name = 0

erikras:
I wonder if this should be documented somehow? Maybe a paragraph under `FormSection`?

erikras:
Sorry, but you'll need to refactor this with the new context API.

PR #4125, opened 2018-07-09
Remove dead code: isChecked file

danielrob:
A bit of a random one, but noticed investigating issues that this file has no references, and all the tests pass without it. Not sure if there was a reason to leave it lying around.


erikras:
BUT HOW DO WE KNOW IF SOMETHING IS CHECKED?!??!?!

🤣🤣🤣🤣

RIP `isChecked`, we hardly used ye.

Issue #4126, opened 2018-07-09
CreateReducer.js uses "startsWith" which is unsupported in IE11


soavie:
### Are you submitting a **bug report** or a **feature request**?

bug report

### What is the current behavior?

Unable to input text into Redux Form Field in IE11. Results in "TypeError: Object doesn't support property or method 'startsWith'" 

### What is the expected behavior?
It should be possible to input text without throwing any errors in IE11.

### What's your environment?

redux-form 7.4.2
IE11

### Other information

Line causing the error:
https://github.com/erikras/redux-form/blob/v7.4.2/src/createReducer.js#L47
```
  if (path.startsWith('values')) {
```
Commit that introduced the error:
https://github.com/erikras/redux-form/commit/a97606fee0ed365903fff80cc93956fe2313a3ce

`String.prototype.startsWith()` browser support:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

We've been able to work around this by using [core-js](https://github.com/zloirock/core-js) to polyfill the feature.

gustavohenke:
Thanks for the heads up. That seems to have been introduced in v7.4.0, and should be an easy fix, if you would like to go for it 🙂

chrisjlee:
Any preference how this is done? 

is this polyfill sufficient?

```
if (!String.prototype.startsWith) {
	String.prototype.startsWith = function(search, pos) {
		return this.substr(!pos || pos < 0 ? 0 : +pos, search.length) === search;
	};
}
```

gustavohenke:
Adding polyfills to Redux Form wouldn't be a good idea, but if you mean using its implementation, then yeah, that's an option.

An even better one would be to use a regular expression, like `/^values/.test(path)`, as then others don't need to guess where the magic number of the `.substr()` call is coming from.

chrisjlee:
Yes implementation-wise. Thank you @gustavohenke !

gustavohenke:
PR #4128 with the fix has been merged. Watch out for new releases!

ldthorne:
Hey @gustavohenke @erikras , do you know when there will be a new release with this included? I'd like to use this fix as I'm hitting the same issue, but haven't been able to because there's no release for it yet. Thanks!

razagill:
@erikras @gustavohenke any update on this?

DonWasyl:
I also waiting for this fix :(

gustavohenke:
[...]

PR #4128, opened 2018-07-11
Replaced startsWith with regex to fix IE11

soavie:
@gustavohenke Do you mean something like this?


soavie:
Sorry. The above is a proposed fix for #4126. Still figuring this stuff out.

erikras:

PR #4130, opened 2018-07-13
Fix getFormError Flow typedef.

davidgomes:
The getFormError selector takes in a form name but it was not accounted for in the Flow typedefs.

Also, the GetFormErrorInterface is not a polymorphic type.

See docs for this selector here:
https://redux-form.com/7.4.2/docs/api/selectors.md/


davidgomes:
Superseded by #4189.

PR #4133, opened 2018-07-16
Add name parameter to field level validation

abrad45:
The Field Level Validation example documents three parameters (`value`, `allValues`, and `props`) but the function accepts `name` as its fourth parameter. This appears to have been added in #3364; while the signature as updated in `docs/api/Field.md`, it was not updated in the examples.


erikras:

PR #4138, opened 2018-07-18
Upgrades performance of FieldArray with immutable data

lynkx:
@erikras I'm opening a pull request for the possible performance upgrade for FieldArray as we discussed. 
~~You'll notice that the FieldArray swap method test is failing, and I'd like to hear your opinion on how can we fix that.~~
There are ~1600 elements in an array.
This is a screenshot from of the Chrome profiler when the fixed is applied. 
image

(Note that this is a development build - so, yeah, redux-logger and why-did-you-update are there)

And this is the screenshot from the current implementation:
image
(A production - optimized build).

In both cases I was doing: Focus, Keystroke, Blur


lynkx:
@erikras can you take a look at this PR?

erikras:

PR #4140, opened 2018-07-23
Remove * imports to avoid warning in React 15

### LINKED ISSUE #4083: PropTypes and createReactClass warning
Bug

```
Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs
lowPriorityWarning.js:38:7
Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
```

No warnings 

### What's your environment?

redux-form 7.4.2
react 15

### Other information

It's because of [this import](https://github.com/erikras/redux-form/blob/ca02205ea339398d157c933bbc7355ce0ab23a1d/src/FormName.js#L3) please add some linting rules to prevent that because this problem keeps happening over and over.

renatoagds:
This PR removes all `*` react imports, since React 15 is supported, when we import everything from React, we enter in a deprecated warning.

Fix #4083

### DIFF
--- a/src/FieldArrayProps.types.js.flow
+++ b/src/FieldArrayProps.types.js.flow
@@ -1,9 +1,9 @@
 // @flow
-import * as React from 'react'
+import type {ElementType, ComponentType} from 'react'
 
 export type Props = {
   name: string,
-  component: React.ComponentType<*> | React.ElementType,
+  component: ComponentType<*> | tType,
   props?: Object,
   rerenderOnEveryChange?: boolean,
   validate?: (value: any, allValues: Object, props: Object) => ?any,
--- a/src/FieldArrayProps.types.js.flow
+++ b/src/FieldArrayProps.types.js.flow
@@ -3,7 +3,7 @@ import type {ElementType, ComponentType} from 'react'
 
 export type Props = {
   name: string,
-  component: ComponentType<*> | tType,
+  component: ComponentType<*> | ElementType,
   props?: Object,
   rerenderOnEveryChange?: boolean,
   validate?: (value: any, allValues: Object, props: Object) => ?any,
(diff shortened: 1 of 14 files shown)


src/FieldArrayProps.types.js.flow
```diff
@@ -1,9 +1,9 @@
 // @flow
-import * as React from 'react'
+import type {ElementType, ComponentType} from 'react'
 
 export type Props = {
   name: string,
-  component: React.ComponentType<*> | React.ElementType,
+  component: ComponentType<*> | tType,
```
gustavohenke: This is breaking Travis, see https://travis-ci.org/erikras/redux-form/jobs/407284062#L6161

renatoagds:
@gustavohenke typo fixed 👍

renatoagds:
@gustavohenke any news on this?

erikras:
I never liked this syntax, but it was needed for Flow. It is no longer needed?

renatoagds:
@erikras I didn't understand what you said.

erikras:
> Note: We import React as a namespace here with `import * as React from 'react'` instead of as a default with `import React from 'react'`. When importing React as an ES module you may use either style, but importing as a namespace gives you access to React’s utility types.
> -- [Flow Docs](https://flow.org/en/docs/react/components/#toc-class-components)

renatoagds:
@erikras oh! now I understood what you said. you could import the named types too, that's what I did.

https://flow.org/en/docs/react/types/

> You can also use named type imports in either an ES module environment or a CommonJS environment:
`import type {Node} from 'react';`

renatoagds:
@erikras can we go ahead with this one?

renatoagds:
@erikras any news? 🤔

PR #4166, opened 2018-08-09
Adding label as prop

nickerzb:
Adding label as prop for Field.  This is shown as being available in [this documentation](https://redux-form.com/7.4.2/examples/syncvalidation/).


gustavohenke:
As you noticed, the typings are done over DefinitelyTyped.
None of the Redux Form maintainers own it, and I have no idea how they are structured. I just intervened to prevent something that is wrong getting merged. 

If there's a lack of support for custom props, you should still send PRs to DefinitelyTyped.

PR #4174, opened 2018-08-15
feat(change): function supported for change-action

caosm:
function supported for `CHANGE`

e.g: `change(fieldName, (fieldValue, allValue) => {})`

```
this.props.change("data", preData => ({
  ...preData,
  name: "hello world"
}));
```

 @erikras PTAL


erikras:
Sure, why not? 😄

erikras:
In strict Redux rules, your actions are supposed to be strict JS objects, so this is a bit of a cheat, but....

erikras:

PR #4185, opened 2018-08-22
Add example for initializing type date

Dubes:
I was trying to help with this stackoverflow [question](https://stackoverflow.com/questions/51970636/how-to-display-initial-value-on-redux-form-date-input-field) where the user was struggling to initialize a `date` type Field and I thought adding it to the example might be helpful for others in the future. 

Here is it in action in a forked codesandbox: https://codesandbox.io/s/32n8221r21


Dubes:
I don't think I can/need to do anything about the codeclimate issues, but please let me know if I should take any action.

PR #4189, opened 2018-08-26
Fix getFormError Flow typedef.

davidgomes:
The getFormError selector takes in a form name but it was not accounted for in the Flow typedefs.

Also, the GetFormErrorInterface is not a polymorphic type.

See docs for this selector here:
https://redux-form.com/7.4.2/docs/api/selectors.md/

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -139,8 +139,9 @@ declare export function formValues(
 ): FormValuesInterface
 
 declare export function getFormError(
+  form: string,
   getFormState: ?GetFormState
-): GetFormErrorInterface<*>
+): GetFormErrorInterface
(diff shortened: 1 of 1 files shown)


davidgomes:
@erikras I re-created this PR from #4130. Would really appreciate a review since we are now using a fork of `redux-form` just to use this PR.

davidgomes:
@erikras @gustavohenke any chance we can push this to master?

mike1808:
@davidgomes we have this exact issue. Our temporary solution is explicit typecasting through `any`.

```js
formError: ((getFormError: any): (form: string) => (state: {}) => any)(form)(state),
```

PR #4248, opened 2018-10-24
Material-ui example with the 3.2.2 version

### LINKED ISSUE #3156: Material UI example doesn't work, please make it clear there is a redux-form-material-ui repo available
Both I guess

The Material UI example in the doc doesn't work, and would lead to unnecessary boilerplate code anyway since you guys created `redux-form-material-ui`.
Please point to https://github.com/erikras/redux-form-material-ui#usage instead

The example should work, *and* we should point to the helper package in any case.

### Sandbox Link
http://redux-form.com/6.8.0/examples/material-ui/
Click on the select input to chose a color, it won't work.

### What's your environment?
Whatever is used in the doc, 6.8.0 I'd assume.

### Other information
I was almost able to create my own SelectField wrapper for my project but I still had a super weird issue with the `onChange` prop, and it's only after a couple hours that I noticed there was an existing `redux-form-material-ui` repo. 
With `redux-form-material-ui` it took me 5 minutes to set it up, I wish I saw that right away, I wouldn't even have tried to develop my own.

Thanks, and thanks for developing these material-ui helpers, it was  a great idea.

caiobessa:
This pull request resolve #3156

The material's documentation example uses an old version of material-ui.

I updated the version "material-ui" to the version 3.2.2

### DIFF
--- a/docs/MigrationGuide.md
+++ b/docs/MigrationGuide.md
@@ -7,9 +7,9 @@ design change.
 
 In `v5`, only the outer form component was connected to the Redux state, and the
 props for each field were passed in via the form component. The problem with
-this is that the _entire_ form component had to rerender _on every single
+this is that the _entire_ form component had to re-render _on every single
 keypress_ that changed a form value. This was fine for small login forms, but
-lead to extremely slow performance on larger forms with dozens or hundreds of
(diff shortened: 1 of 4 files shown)


caiobessa:
@gustavohenke  Everything is ok? Do I need to improve another thing?

PR #4254, opened 2018-10-30
Pull down sections of lodash

TerribleDev:
Howdy ya'll

I've been looking into my bundle size and I noticed I have more lodash than I thought I should. 

I noticed redux-form `imports 'lodash'` and not the specific sub files containing the actual functions used. I think this becomes a problem when all the code is bundled up, and it certainly doesn't hurt to only import the specific bits instead of the root file. I largely left the tests untouched figuring it didn't matter so much for the tests.

I've never contributed to redux-form so plz go easy on me! I'm not sure how to test this to make sure I didn't break something.


TerribleDev:
This is an example of 1 dependency.

Before:

image


After:

image

erikras:
It is my understanding that the [`lodash` babel plugin](https://github.com/erikras/redux-form/blob/master/.babelrc#L16) does this automatically (one of it's primary functions). Have you confirmed a reduction in bundle size after this change?

TerribleDev:
Hi yes, sorry for not being responsive. You are right lodash was a different dependency we had

PR #4309, opened 2018-12-16
A way to cancel dispatching events for RN.

ErrorPro:
Right now there's no way to cancel dispatching an event for React-Native.
Assign defaultPrevented to a returned value of the onChange callback for React-Native to prevent dispatching onChange/onFocus/onBlur for some cases.


erikras:

Issue #4310, opened 2018-12-17
Reducer plugins should be able to receive actions from other forms


griffinmichl:
# Feature Request

### What is the current behavior?
https://github.com/erikras/redux-form/issues/1127 introduced a behavior where redux-form plugins only receive actions that belong to that form. I would say this is pretty counterintuitive and prevents some behaviors that would be nice to have, particularly in scenarios where you're using `destroyOnUnmount: false`.

### What is the expected behavior?
I would expect plugins to receive all actions. It sort of goes against the redux paradigm that I would not receive certain actions in the reducer based on some part of the action (in this case `action.meta.form`). In the plugin, I could check whether a given redux-form action is coming from my form or not.

This would be a breaking change so we'd probably have to introduce it as an option.

### Other information
Here is the line where the existing behavior happens: https://github.com/erikras/redux-form/blob/master/src/createReducer.js#L806

erikras:
Published fix in [`v8.1.0`](https://github.com/erikras/redux-form/releases/tag/v8.1.0).

PR #4311, opened 2018-12-17
Add config for plugins receving actions from all forms

griffinmichl:
One possible solution to: https://github.com/erikras/redux-form/issues/4310

If this seems like something we'd want to do, I can polish up this PR more.


erikras:

PR #4314, opened 2018-12-18
Convert immutable List to array

### LINKED ISSUE #4313: @@redux-form/TOUCH does not touch fields after update to v8.0.4 using immutable

bug report

### [Updated]
This issue only occurs in conjunction with usage of immutable js.
https://codesandbox.io/s/jlpklkx675

In v8.0.4 the submit dispatches the following:
```js
{
  type: '@@redux-form/TOUCH',
  meta: {
    form: 'syncValidation',
    fields: [
    ]
  }
}
```

Note that the fields array is empty so no fields get touched when the form is submitted. This causes the form not to show validation errors on fields which were not touched manually by the user if the field logic includes something like `{meta.touched && meta.error}`

https://codesandbox.io/s/pQj03w7Y6

In v7.0.4 the submit dispatches the following:
```js
{
  type: '@@redux-form/TOUCH',
  meta: {
    form: 'syncValidation',
    fields: [
      'username',
      'email',
      'age'
    ]
  }
}
```

### Sandbox Link

https://codesandbox.io/s/jlpklkx675

### What's your environment?

Redux form version: 8.0.4
OS/browser: Mac, Chrome

### Other information

I traced the issue to a change in the babel configuration for the project. `handleSubmit` uses the spread operator to pass an array of fields as parameters to the `touch` action creator. https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L38 
```js
  touch(...fields) // mark all fields as touched
```

and touch is defined as
https://github.com/erikras/redux-form/blob/master/src/actions.js#L385
```js
const touch: Touch = (form: string, ...fields: string[]): TouchAction => ({
  type: TOUCH,
  meta: { form, fields }
})
```

In v7.0.4 https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L38 is transpiled as:
https://unpkg.com/redux-form@7.0.4/lib/handleSubmit.js
```js
function _toConsumableArray(arr) {
    if (Array.isArray(arr)) {
        for (var i = 0, arr2 = Array(arr.length); i < arr.length; i++) {
            arr2[i] = arr[i];
        }
        return arr2;
    } else {
        return Array.from(arr);
    }
}

  touch.apply(undefined, _toConsumableArray(fields)); // mark all fields as touched
```
And in v8.0.4 https://github.com/erikras/redux-form/blob/master/src/handleSubmit.js#L38 is transpiled as:
https://unpkg.com/redux-form@8.0.4/lib/handleSubmit.js
```js
  touch.apply(void 0, fields); // mark all fields as touched
```

In v7.0.4 _toConsumableArray was passed an immutable `List` of field values which failed `Array.isArray(arr)` and returned an array via `Array.from(arr)`.

The transpiled version of `touch` is not compatible with an immutable list:
```js
var touch = function touch(form) {
  for (var _len6 = arguments.length, fields = new Array(_len6 > 1 ? _len6 - 1 : 0), _key6 = 1; _key6 < _len6; _key6++) {
    fields[_key6 - 1] = arguments[_key6];
  }

  return {
    type: _actionTypes.TOUCH,
    meta: {
      form: form,
      fields: fields
    }
  };
};
```

esetnik:
Fixes #4313

### DIFF
--- a/src/handleSubmit.js
+++ b/src/handleSubmit.js
@@ -35,7 +35,7 @@ const handleSubmit = (
     persistentSubmitErrors
   } = props
 
-  touch(...fields) // mark all fields as touched
+  touch(...Array.from(fields)) // mark all fields as touched
 
   if (valid || persistentSubmitErrors) {
     const doSubmit = () => {


esetnik:
Note: I think we're going to have the same problem with setSubmitFailed. See https://github.com/erikras/redux-form/blob/a59fb7c870523f68c8806bf401beef3e52d7a5a6/src/handleSubmit.js#L50

erikras:

PR #4320, opened 2018-12-20
fix hoc displayName issue

rkurbatov:
Our tests rely on component names and with special hoc additon that was broken. Nested components were named as Component. That fix restores original naming.

### DIFF
--- a/src/ReduxFormContext.js
+++ b/src/ReduxFormContext.js
@@ -16,7 +16,10 @@ export const withReduxForm = Component => {
       })
     }
   }
-  return React.forwardRef((props, ref) =>
+
+  const ref = React.forwardRef((props, ref) =>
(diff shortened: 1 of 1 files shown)


rkurbatov:
It's fine to have breaking changes but it's not fine to have `Connect(Component)` instead of `Connect(ReduxForm)` or `ForwardRef` instead of expected `Field` - that's pretty... ehm... non-informative.

There are two common approaches for any wrapper component naming:
1. Pass the name of the wrapped component transparently (as here)
2. Pass the name of the wrapped component with wrapper name surrounded (like `withReduxForm(WrappedComponnet` or `withForwardRef(Field)`.

I'm ok with both approaches even if the second one requires rewriting the tests. But as to mee it needs to be handled somehow.

erikras:
This seems fine to me.

erikras:

Issue #4323, opened 2018-12-21
Wrong imports/exports in es build


vtaits:
### Are you submitting a **bug report** or a **feature request**?

bug report

### What is the current behavior?

Commonjs in es-build, e.g. `es/actionTypes.js`

```
"use strict";

exports.__esModule = true;
exports.UPDATE_SYNC_WARNINGS = exports.UPDATE_SYNC_ERRORS = exports.UNTOUCH = exports.UNREGISTER_FIELD = exports.TOUCH = exports.SUBMIT = exports.STOP_SUBMIT = exports.STOP_ASYNC_VALIDATION = exports.START_SUBMIT = exports.START_ASYNC_VALIDATION = exports.SET_SUBMIT_SUCCEEDED = exports.SET_SUBMIT_FAILED = exports.RESET_SECTION = exports.RESET = exports.REGISTER_FIELD = exports.INITIALIZE = exports.FOCUS = exports.DESTROY = exports.CLEAR_ASYNC_ERROR = exports.CLEAR_SUBMIT_ERRORS = exports.CLEAR_SUBMIT = exports.CLEAR_FIELDS = exports.CHANGE = exports.BLUR = exports.AUTOFILL = exports.ARRAY_SWAP = exports.ARRAY_UNSHIFT = exports.ARRAY_SPLICE = exports.ARRAY_SHIFT = exports.ARRAY_REMOVE_ALL = exports.ARRAY_REMOVE = exports.ARRAY_PUSH = exports.ARRAY_POP = exports.ARRAY_MOVE = exports.ARRAY_INSERT = exports.prefix = void 0;
var prefix = '@@redux-form/';
exports.prefix = prefix;
var ARRAY_INSERT = prefix + "ARRAY_INSERT";
exports.ARRAY_INSERT = ARRAY_INSERT;
var ARRAY_MOVE = prefix + "ARRAY_MOVE";
exports.ARRAY_MOVE = ARRAY_MOVE;
var ARRAY_POP = prefix + "ARRAY_POP";
exports.ARRAY_POP = ARRAY_POP;
var ARRAY_PUSH = prefix + "ARRAY_PUSH";
exports.ARRAY_PUSH = ARRAY_PUSH;
var ARRAY_REMOVE = prefix + "ARRAY_REMOVE";
exports.ARRAY_REMOVE = ARRAY_REMOVE;
var ARRAY_REMOVE_ALL = prefix + "ARRAY_REMOVE_ALL";
[...]

PR #4330, opened 2018-12-23
Fix imports and exports in es-build (#4323)

vtaits:
#4323 Fixed `.babelrc.js`. Prevent imports and exports processing if current `BABEL_ENV` is `es`.


erikras:

PR #4337, opened 2018-12-26
disable loose mode for @babel/plugin-transform-spread

esetnik:
This is a more general solution to #4313 because spread operators are used with immutable in several places in the project. Instead of fixing them all manually we can disable `loose` mode in `@babel/plugin-transform-spread`.

```
In loose mode, all iterables are assumed to be arrays.
```


esetnik:
@erikras can we get a review of this? It fixes a number of issues.

erikras:

erikras:
Reverted in [`v8.2.1`](https://github.com/erikras/redux-form/releases/tag/v8.2.1).

PR #4352, opened 2019-01-12
FormProps export

### LINKED ISSUE #3823: [Flow] Can't "import { type FormProps } from 'redux-form';"
This appears to be because `Form.js.flow` is not published. In `index.js.flow` there are these relevant lines:
```jsx
import type { Props as _FormProps } from './Form'
...
export type FormProps = _FormProps
```

However `./Form` will not resolve to a `.flow` file, because one does not exist in `node_modules/redux_form/es` or `node_modules/redux-form/lib`.

This means that these docs are actually all incorrect: https://redux-form.com/7.3.0/docs/flow.md/. You can however workaround this by using this instead:
```
import { type FormProps } from 'redux-form/es/types.js.flow';
```

asazernik:
FormProps should have referred to props passed by reduxForm() to wrapped components, but was instead referring to the props expected by the trivial <Form> component. This will cause new typechecking errors for people relying on the old behavior, but I suspect that is an uncommon use case.

Fixes #3823

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -14,12 +14,10 @@ import type {
 } from './FormSection'
 import type {
   Props as FieldInputProps,
-  FieldProps as _FieldProps
 } from './FieldProps.types'
 import type { Props as FieldsInputProps } from './FieldsProps.types'
 import type {
   Props as FieldArrayInputProps,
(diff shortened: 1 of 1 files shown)


erikras:

PR #4353, opened 2019-01-12
handleSubmit type

### LINKED ISSUE #4351: handleSubmit type declaration is incorrect (too strict)

`handleSubmit` can be used in two ways: `<SomeElement onClick={handleSubmit} />` or `<SomeElement onClick={handleSubmit(innerCallback)} />`. The current type `(eventOrSubmit: any) => void | Promise<*>`, despite having made the parameter `any` to handle both cases, has the proper return type only for the former case. In the latter case, calling `handleSubmit` returns a function rather than `void | Promise`.

### Sandbox Link

Example is contrived to have my submit handler passed to a type-checked child component.

https://codesandbox.io/s/j4vjj6n2nv

You'll have to download it and run `npm install; npx flow` to see the errors, because Flow doesn't run in codesandbox.

### What's your environment?
Redux-form 7.1.2 and 8.1.0.

asazernik:
Fixes #4351

### DIFF
--- a/src/handleSubmit.js
+++ b/src/handleSubmit.js
@@ -1,14 +1,8 @@
 // @flow
 import isPromise from 'is-promise'
-import type { Dispatch } from 'redux'
+import type { SubmitFunction } from './types'
 import type { Props } from './createReduxForm'
 import SubmissionError from './SubmissionError'
-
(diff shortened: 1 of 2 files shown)


erikras:

PR #4361, opened 2019-01-17
Field 'type' prop documentation

asazernik:
There's some behavior here that is essential knowledge for users implementing custom components that are like special input types like 'radio' or 'select-multiple'.


docs/api/Field.md
```diff
@@ -126,12 +126,27 @@ localized date formats into `Date`s.
 `parse` is called with the field `value` and `name` as arguments and should return the new
 parsed value to be stored in the Redux store.
 
+#### `type : string` [optional]
+
+Used to determine the exact shape of the `input` prop passed through to
+the component provided to `component` prop. For example, if `type === 'checkbox'`, the `checked` field will be passed through. Use this if either:
+
+1. `component` is a `string` value (i.e. a native HTML element)
+2. `component` requires the `input` prop to follow the special semantics required for `'radio' | 'checkbox' | 'select-multiple' | 'file'`.
```
gustavohenke: Don't you think it would be worth listing what those "special semantics" mean?

asazernik: The actual differences are quite large, but I could give examples; how does this sound?

1. foo
2. `component` requires the `input prop to follow the special semantics required for `'radio' | 'checkbox' | 'select-multiple' | 'file'`, such as using `[]` instead of null for an empty `'select-multiple'` value or passing the boolean `checked` prop for `'radio' | 'checkbox'`.

asazernik: Actually, on reading through the code, it's not that bad - I'll put together a full write-up when I'm off work.

asazernik: Aaaaand off work ended up being several weeks. Is now much more detailed and describes all the special cases.

erikras:

Issue #4237, opened 2018-10-18
TOUCH triggers wizard forms to display errors when navigating back and forth


mantastrunce:
### Are you submitting a **bug report** or a **feature request**?
Bug report

### What is the current behavior?
Steps to reproduce:
1. Go to https://codesandbox.io/s/5436nzxrmn
2. Enter your first name and last name.
3. Click next.
4. Click previous.
5. Click next.
Results: Errors are triggered and shown on the second wizard page due to a TOUCH event.

### What is the expected behavior?
Expected Results: Errors are not triggered and shown on the second wizard page because the user has not taken any action.

### Sandbox Link
https://codesandbox.io/s/5436nzxrmn

### What's your environment?
7.4.2
https://codesandbox.io/s/5436nzxrmn
7.0.4 does not have this issue
https://codesandbox.io/s/0Qzz3843

showme79:
As I see the problem is originated that TOUCH is called for registerdFields, but "count" property - which is a counter for registered/unregistered field - is not checked.
See createReduxFrom.js/submit => createReduxFrom.js/getFieldList methods.

irvingv8:
This issue seems to be fixed as 8.2, I guess this issue could be close. @erikras is it possible to get this in 7.4.X ?

PR #4382, opened 2019-02-17
Fixed issue #4237

showme79:
TOUCH triggers wizard forms to display errors when navigating back and forth


showme79:
Please check the original issue #4237 
My change is done in master (8.1.0) but it should work on 7.4.2 as well.

PR #4383, opened 2019-02-17
Don't list and touch unregistered fields on submit

showme79:
I have added a new excludeUnregistered option to getFieldsList() method which enables / disables the excluding of fields registered with zero reference.

I have also added to use this option in the submit(...) call.

### DIFF
--- a/src/createReduxForm.js
+++ b/src/createReduxForm.js
@@ -639,7 +639,7 @@ const createReduxForm = (structure: Structure<*, *>) => {
             }
             if (options.excludeUnregistered) {
               keySeq = keySeq.filter(
-                name => getIn(registeredFields, `${name}.count`) !== 0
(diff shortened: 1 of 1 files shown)


showme79:
This is my second try, tested and working fine on my project (I made a small mistake in my first pull request, so I closed it).

showme79:
I think the test case 'should mark all fields as touched on submit' in reduxForm.spec.js should be updated with the following change:
 - test case should named _'should mark all **registered** fields as touched on submit'_
 - the test should be enhanced with an unmounted field (a similar solution what is in _'should not destroy on unmount if told not to'_ test case) and a check which tests if only the registered (mounted) fields are touched.

Do you agree with it or shall I create a completely new test case for it?

erikras:

PR #4398, opened 2019-03-12
Fix async validation in case when promise is rejected with some data

lama-pacos:
Fixed bug where async validation with a resolved Promise was halting form submission https://github.com/erikras/redux-form/issues/2056

-demo


renatoagds:
https://github.com/erikras/redux-form/issues/2056 was closed in [`v7.4.0`](https://github.com/erikras/redux-form/releases/tag/v7.4.0).

PR #4404, opened 2019-03-22
Improve detection of whether mapToProps functions depends on props.

krvajal:
When using this version with the latest version of `reselect`, the props are not been passed to the  `mapStateToProps` selector generated by `createStructuredSelector` that the `form` HOC from redux-form takes for its second argument. The reason for that is that it does not accurately report its *arity*
as seen here
https://github.com/reduxjs/reselect/blob/master/src/index.js#L68.
This case is handled in the `react-redux` library  here
https://github.com/reduxjs/react-redux/blob/72ed6db1947913bdd4900b324243c43a490b4c5a/src/connect/wrapMapToProps.js#L22
I just copied it over and made sure we treat it the same way.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,0 +1,9 @@
+// A length of one signals that mapToProps does not depend on props from the parent component.
+// A length of zero is assumed to mean mapToProps is getting args via arguments or ...args and
+// therefore not reporting its length accurately..
+export function getDependsOnOwnProps(mapToProps) {
+  return mapToProps.dependsOnOwnProps !== null &&
+    mapToProps.dependsOnOwnProps !== undefined
+    ? Boolean(mapToProps.dependsOnOwnProps)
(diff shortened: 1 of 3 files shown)


src/helpers.js
```diff
@@ -0,0 +1,9 @@
+// A length of one signals that mapToProps does not depend on props from the parent component.
+// A length of zero is assumed to mean mapToProps is getting args via arguments or ...args and
+// therefore not reporting its length accurately..
+export function getDependsOnOwnProps(mapToProps) {
+  return mapToProps.dependsOnOwnProps !== null &&
+    mapToProps.dependsOnOwnProps !== undefined
```
krvajal: Not sure this part is needed but it seems it is provided by some libraries. So I kept it here. But we can also simplify this and just use
```
return mapToProps.length !== 1;

```

erikras:

PR #4431, opened 2019-04-10
Upgrade dev dependency to react-redux v7

erikras:
This breaks a _lot_ of the tests, and I'm not certain why. Many of the `redux-form` tests focus on how many times stuff gets rerendered, and something about that has changed with `react-redux@7`.

🤔

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,17 +5,17 @@
   "requires": true,
   "dependencies": {
     "@babel/cli": {
-      "version": "7.2.3",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.2.3.tgz",
-      "integrity": "sha512-bfna97nmJV6nDJhXNPeEfxyMjWnt6+IjUAaDPiYRTBlm8L41n8nvw6UAqUCbvpFfU246gHPxW7sfWwqtF4FcYA==",
+      "version": "7.4.3",
+      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.4.3.tgz",
+      "integrity": "sha512-cbC5H9iTDV9H7sMxK5rUm18UbdVPNTPqgdzmQAkOUP3YLysgDWLZaysVAfylK49rgTlzL01a6tXyq9rCb3yLhQ==",
       "dev": true,
       "requires": {
-        "chokidar": "^2.0.3",
+        "chokidar": "^2.0.4",
         "commander": "^2.8.1",
         "convert-source-map": "^1.1.0",
         "fs-readdir-recursive": "^1.1.0",
         "glob": "^7.0.0",
-        "lodash": "^4.17.10",
+        "lodash": "^4.17.11",
         "mkdirp": "^0.5.1",
         "output-file-sync": "^2.0.0",
         "slash": "^2.0.0",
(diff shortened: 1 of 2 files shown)


markerikson:
That's... odd.

I would have expected the number of re-renders to be pretty consistent between v6 and v7.  The number of times `mapState` is called would likely have changed, but actual re-renders shouldn't (much).

belliAndrea:
Shouldn't react be updated as well? I see that v7 requires > 16.8.4.
In my project i've put redux-form 8.1.0, react-redux v7 and 16.8.4 and seems to run fine.

vincentjames501:
Thanks for your continued work on this @erikras , many people still love and rely on Redux-Form!

josepot:
Hi!

I have done some investigation and I think that I know what one of the issues is (quite possibly the main issue). I _think_ that I have some ideas on how to fix it, but it's getting late and I want to go to bed. So, I will keep looking at this in another moment.

In the meanwhile, I will explain here what I have found, in case someone else wants to have a look at it before I get back to it:

The issue that I have found is that when a Field registers itself to the reduxForm a dispatch takes place (`this.props.registerField(name, type)`). However, when that dispatch happens, there are no listeners registered into the store yet :open_mouth:

markerikson:
When exactly in the component lifecycle is that action being dispatched?

One thing that did change in v7 is that we're subscribing in a `useEffect()`, which will run just a bit async after the component is ready.  In v5, we subscribed in `componentDidMount` (which is equivalent to `useLayoutEffect()`).

josepot:
Oh, ok, thanks @markerikson !

So, let describe with more detail what I'm seeing:

- Let's debug [one of the "simple" tests that are failing](https://github.com/erikras/redux-form/blob/master/src/__tests__/reduxForm.spec.js#L340-L342):

```bash
node --inspect-brk node_modules/.bin/jest --runInBand -t 'should be `false` when `errors` has a `string` property'
```

- In the previous commit, where all the tests are passing,[ the following line](https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L1028) gets evaluated **4 times** for that test. Each time with a different state.

- In the latest commit that line only gets evaluated once, with the empty state.

- I can see that there is a dispatch that gets triggered and that dispatch has no listeners... However, later on, [this line of react-redux tries to take care of that](https://github.com/reduxjs/react-redux/blob/master/src/components/Provider.js#L30). That's cool :smile:

- Unfortunately, if I place a debugger [here](https://github.com/erikras/redux-form/blob/master/src/createReduxForm.js#L993). I don't see that selector being evaluated after that. :thinking: :confused:

josepot:
Dah! Bingo! I found the issue! It's a problem with react-redux... I will create a fix and send a PR! (It seems that I won't get any sleep tonight :smile: )

josepot:
More interesting stuff, [check this out.](https://codesandbox.io/s/7y5xw6pxx1).

That sandbox does the exact same thing that [this test does](https://github.com/erikras/redux-form/blob/master/src/__tests__/reduxForm.spec.js#L340-L342). However, in the sandbox App the output is the correct one.

That makes me think that the problem probably has to do with the way that the tests are being setup. Either that, or perhaps there is an issue with `react-dom/test-utils`.

mikkom:
I think we should add calls to `TestUtils.act` in order to flush the effects in the tested components. 

See for example this commit in a hook version of react-redux-promise-listener: https://github.com/mikkom/react-redux-promise-listener-hook/commit/eebee1eafff748d2c8809926a67780124ae704fa

erikras:
Thanks for your effort on this, guys.

PR #4438, opened 2019-04-15
FieldProps.types: Relax component type, document problem with strictness

### LINKED ISSUE #4434: FieldProps Flow Error redux-form@8.2.0
bug report

```react
<Field
	autoCapitalize="none"
	component={FormInput}
	keyboardType="email-address"
	label="Username"
	name="username"
/>
```
and FormInput's props are

```flow
type Props = {
	label: string,
} & FieldProps;
```

image

I expect the above to not return a Flow error.

### Sandbox Link
https://codesandbox.io/s/gJjDoKYWl seems to demonstrate an example of using a Field to pass-through properties.

### What's your environment?
redux-form: 8.2.0
react-native:  0.95.4
react-redux: 7.0.1
flow: 0.96.0
node: 11.10.1

### Other information
This flow error doesn't occur in 8.1.0 but does occur in 8.2.0.

I think it's because it doesn't like that I'm trying to pass properties through to my underlying component. Is it related to the changes in [this commit](https://github.com/erikras/redux-form/commit/a6b626dd795b898683a10a9245bc7af86cfcd385)?

asazernik:
Components are like functions - contravariant in their arguments! Which
means we'd need to have a generic Props<PassthroughProps>, and somehow
would need to deal with the two prop-passthrough styles (though the
`props` prop and passed straight to the Field).

And who knows how well that would even type-infer.

Should fix #4434

### DIFF
--- a/src/FieldProps.types.js.flow
+++ b/src/FieldProps.types.js.flow
@@ -12,7 +12,11 @@ type FieldTypeProps = {
 
 export type Props = {
   name: string,
-  component: ComponentType<FieldProps> | string,
+
+  // this is hard (impossible?) to type strictly - should be ComponentType<FieldProps & Passthrough>
+  // where Passthrough is `Props.props & ${any other props passed to the Field component}`
+  component: ComponentType<*> | string,
+
   format?: ?(value: any, name: string) => ?string,
   normalize?: (
     value: any,


erikras:
Published in [`v8.2.2`](https://github.com/erikras/redux-form/releases/tag/v8.2.2).

PR #4442, opened 2019-04-18
adds CONTRIBUTING.md

tnrich:
Can someone please help me fill out the CONTRIBUTING.md so it is actually useful. I'm just getting things started here. 

@erikras @asazernik @showme79 @YunzheZJU @sanex3339 

(you all contributed recently so that's why you're mentioned. Thanks!)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,10 @@
-# Instructions for new contributors 
+# Instructions for new contributors
 
- - Make a fork of https://github.com/erikras/redux-form
- - ? Install deps ?
- - Change the code to implement your feature/fix
- - ? Rebuild the app ?
- - pull request your branch into https://github.com/erikras/redux-form
- 
+- Make a fork of https://github.com/erikras/redux-form and a branch off that fork
+- Install deps: `npm install`
+- Change the code to implement your feature/fix
+- When you're ready to commit run flow typechecker: `npm run test:flow`
+- And the full test suite: `npm run test`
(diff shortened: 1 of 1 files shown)


tnrich:
Also if someone could explain why the travis tests are all failing..? I am not sure I understand. Probably related to me not knowing how to best contribute. Thanks!

CONTRIBUTING.md
```diff
@@ -0,0 +1,8 @@
+# Instructions for new contributors 
+
+ - Make a fork of https://github.com/erikras/redux-form
+ - ? Install deps ?
```
asazernik: `npm install`

CONTRIBUTING.md
```diff
@@ -0,0 +1,8 @@
...
+ - Change the code to implement your feature/fix
+ - ? Rebuild the app ?
```
asazernik: pre-commit hooks should do a sanity-check build; other good steps to take before making a PR are to run a typechecker (`npm run test:flow` - quick but not thorough) and then the full test suite (`npm run test`). Another good thing to check is `npm run test:cov` - this will check if you've added any new code that is not exercised by the test suite.

tnrich: I've improved the docs. Thanks for the feed back @asazernik !

CONTRIBUTING.md
tnrich: Added that!

tnrich:
Okay let me know if this is good to merge in @asazernik ! Thanks!

PR #4443, opened 2019-04-18
fix #3946

### LINKED ISSUE #3946: Calling Field's input.onChange(undefined) does not set field value to undefined.
Hey there, 
I'm trying to reset a form field to be `undefined` but it doesn't seem to work. Here is a codesandbox showing it not working: 

https://codesandbox.io/s/5w72x5rpmx

It does work passing `null`, but I am confused why `undefined` doesn't work. I couldn't find any mention of why this would be the case in the docs but maybe I missed something. 

Besides this behavior being confusing, I would especially love it to be set to `undefined` (and not `null` or `false`) in my case because I often set certain default values for my form values using the 
```
const {input: {value = "hey"}} = this.props
```
Which only works if the value gets reset to `undefined`.

Thanks!

tnrich:
connects https://github.com/erikras/redux-form/issues/3946

### DIFF
--- a/src/createReducer.js
+++ b/src/createReducer.js
@@ -303,7 +303,7 @@ function createReducer<M, L>(structure: Structure<M, L>) {
     ) {
       let result = state
       const initial = getIn(result, `initial.${field}`)
-      if (initial === undefined && payload === '') {
+      if (payload === '') {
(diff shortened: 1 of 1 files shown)


renatoagds:
Already fixed in #4507

PR #4446, opened 2019-04-23
feat: export ReduxFormContext

### LINKED ISSUE #3852: Consider _reduxForm context part of public, semantic-versioned API

Feature request

Sometimes I find it way more convenient to use `_reduxForm` context directly than to use the various utility HOCs provided by redux form.  As an example, below is a container for a group of form fields I made to be embedded within other forms.  When the user types in a different tag, it fetches the metadata for that tag from graphql and then updates the other corresponding fields with the metadata accordingly.

The only problem with this is I assume it's not safe from minor or patch releases changing the structure of context.  But I think this is convenient enough that I would like to request that any such breaking changes to context only be included in a major release.

Note especially that I'm using the `sectionPrefix` field from context, which there doesn't really seem to be a way to access directly.

To do this without using context, I could pass in the form's `change` method and a prefix for field changes manually, which would not be so slick.  Perhaps I could nest `MetadataItemFieldsContainer` and its `Fields` inside of a `Field` with the same name as my section prefix, but I doubt that would completely work (and I don't think `onChange` can be used to dispatch subfield changes anyway).

```js
// @flow

import * as React from 'react'
import PropTypes from 'prop-types'
import {compose} from 'redux'
import {formValues} from 'redux-form'
import gql from 'graphql-tag'
import {graphql} from 'react-apollo'
import type {MetadataItem} from '../types/MetadataItem'
import MetadataItemFields from './MetadataItemFields'

export type Props = {
  tag: ?string,
  data: {
    metadataItem?: MetadataItem,
    loading: boolean,
  },
}

const metadataItemQuery = gql(`query MetadataItem($tag: String!) {
  metadataItem: MetadataItem(tag: $tag) {
    _id
    tag
    name
    dataType
    ... on NumericMetadataItem {
      min
      max
      units
      storagePrecision
      displayPrecision
    }
    ... on DigitalMetadataItem {
      isDigital
    }
  }
}`)

type Context = {
  _reduxForm: {
    sectionPrefix?: ?string,
    change: (field: string, newValue: any) => any,
    dispatch: (action: any) => any,
  },
}

class MetadataItemFieldsContainer extends React.Component<Props> {
  static contextTypes = {
    _reduxForm: PropTypes.shape({
      sectionPrefix: PropTypes.string,
      change: PropTypes.func.isRequired,
      dispatch: PropTypes.func.isRequired,
    }),
  }

  _updateFields = (metadataItem: MetadataItem, context: Context = this.context) => {
    const {_reduxForm: {sectionPrefix, change, dispatch}} = context
    for (let key in metadataItem) {
      const field = sectionPrefix ? `${sectionPrefix}.${key}` : key
      dispatch(change(field, metadataItem[key]))
    }
  }

  componentDidMount() {
    const {data: {metadataItem}} = this.props
    if (metadataItem) this._updateFields(metadataItem)
  }

  componentWillReceiveProps(nextProps: Props, nextContext: Context) {
    const prevMetadataItem = this.props.data.metadataItem
    const nextMetadataItem = nextProps.data.metadataItem
    if (nextMetadataItem && nextMetadataItem !== prevMetadataItem) {
      this._updateFields(nextMetadataItem, nextContext)
    }
  }

  render(): ?React.Node {
    const {
      tag, data, // eslint-disable-line no-unused-vars
      ...props
    } = this.props
    return <MetadataItemFields {...props} />
  }
}

export default compose(
  formValues('tag'),
  graphql(metadataItemQuery, {
    options: ({tag}: Props) => ({
      variables: {tag},
      errorPolicy: 'all',
    }),
  })
)(MetadataItemFieldsContainer)
```

jedwards1211:
Pretty please 🍒 

fix #3852

`react-router` allows components to access the ancestor `<Route>` props from context by using `withRouter`;  Likewise `redux-form` should allow components to access the ancestor form props from context.

I have a component that needs to know know its enclosing `sectionPrefix` and dispatch `change` multiple fields under that `sectionPrefix`, and the public API (`formValueSelector`, custom `Field` components, action creators and selectors, etc) just doesn't cut it for this use case.

### DIFF
--- a/src/immutable.js
+++ b/src/immutable.js
@@ -1,7 +1,7 @@
 // @flow
 import actions from './actions'
 import * as _actionTypes from './actionTypes'
-export { ReduxFormContext, withReduxForm } from './ReduxFormContext'
+export { ReduxFormContext } from './ReduxFormContext'
 export {
   default as defaultShouldAsyncValidate
 } from './defaultShouldAsyncValidate'
--- a/src/immutable.js
+++ b/src/immutable.js
@@ -1,6 +1,7 @@
 // @flow
 import actions from './actions'
 import * as _actionTypes from './actionTypes'
+export { ReduxFormContext, withReduxForm } from './ReduxFormContext'
 export {
   default as defaultShouldAsyncValidate
 } from './defaultShouldAsyncValidate'
--- a/src/immutable.js.flow
+++ b/src/immutable.js.flow
@@ -1,6 +1,7 @@
 // @flow
+import * as React from 'react'
 import type { ComponentType } from 'react'
-import type { Action, GetFormState } from './types'
+import type { Action, GetFormState, Context, ReactContext } from './types'
 import type { Params as DefaultShouldAsyncValidateParams } from './defaultShouldAsyncValidate'
 import type { Params as DefaultShouldValidateParams } from './defaultShouldValidate'
 import type { Params as DefaultShouldErrorParams } from './defaultShouldError'
@@ -76,6 +77,10 @@ import type {
 import type { Actions } from './actions.types'
 
 type StructureMap = Object
+
+declare export var ReduxFormContext: React.Context<Context>
+declare export var withReduxForm: <P>(React.ComponentType<P>) =>
+  React.ComponentType<P & ReactContext>
 
 export type FormProps = _FormProps
 export type FormNameProps = _FormNameProps
--- a/src/immutable.js.flow
+++ b/src/immutable.js.flow
@@ -79,8 +79,6 @@ import type { Actions } from './actions.types'
 type StructureMap = Object
 
 declare export var ReduxFormContext: React.Context<Context>
-declare export var withReduxForm: <P>(React.ComponentType<P>) =>
-  React.ComponentType<P & ReactContext>
 
 export type FormProps = _FormProps
 export type FormNameProps = _FormNameProps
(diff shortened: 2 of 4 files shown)


iamandrewluca:
CI errors are triggered by flow not finding react types. With this PR merged no errors are thrown.

https://github.com/erikras/redux-form/pull/4526
https://flow.org/en/docs/react/context/

jedwards1211:
@renatoagds you're saying you want to get rid of the HoC and just have people use `useContext(ReduxFormContext)` or `ReduxFormContext.Consumer`, right?

jedwards1211:
@renatoagds okay I got rid of the `withReduxForm` export

iamandrewluca:
@renatoagds you are right. Also if we move to react hooks we can get rid of it tottaly. And who wants to use it will do something like this.
```
import { useForm } from 'redux-form'

export const withReduxForm(options) {
	return Component => props => {
		const form = useForm(options)
		return <Component {...props} _reduxForm={form} />
	}
}
```

renatoagds:
@iamandrewluca @jedwards1211 sorry for the delay, we need one more thing before release this, we should update our docs and add this new export with usage examples, can one of you do that?

renatoagds:
@iamandrewluca today @erikras is responsible to generate the new docs.

renatoagds:

PR #4450, opened 2019-04-26
Update README.md

gtournie:
Add redux-form-validators code sandboxes

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ You can play around with `redux-form` in these sandbox versions of the Examples.
 - [Remote Submit](https://codesandbox.io/s/ElYvJR21K)
 - [Normalizing](https://codesandbox.io/s/L8KWERjDw)
 - [Immutable JS](https://codesandbox.io/s/ZVGJQBJMw)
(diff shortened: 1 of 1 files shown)


gtournie:
@gustavohenke : just trying to help here... I've seen many many tutorials/examples on the web where people create their validators from scratch (in other words, reinvent the wheel...). You support immutable.js (which is on your README), why not supporting a lib for validations? I created this lib because I couldn't find any other that would do the same job. Why not adding an "external tools" section and mention the most used/useful redux-form plugins...? Or maybe there is another place where we could put that kind of stuff...?

gustavohenke:
Redux Form builds on top of Immutable, and there are libraries that build on top of Redux Form (the case of redux-form-validators). Promoting one of these over the others is where the unfair lives.

Nonetheless, it's already listed on this file:
https://github.com/erikras/redux-form/blob/master/tools.md

Maybe it needs more visibility.

erikras:
@gustavohenke @gtournie contacted me directly and requested that his tool could be advertised in the RF docs. I'm always happy to promote companion helper libs, but I agree that the README is probably the wrong place. I was thinking more on a page [like this](https://redux-form.com/8.2.0/examples/syncvalidation/) ([actual file here](https://github.com/erikras/redux-form/blob/master/examples/syncValidation/src/SyncValidation.md)), where it could be hinted that there is a useful library that could be handy for writing validation.

Issue #4299, opened 2018-12-12
Error: formValues() must be used inside a React tree decorated with reduxForm()


someden:
**Bug report**

Demo: https://stackblitz.com/edit/redux-form8?file=Form.js
But demo doesn't work because of https://github.com/erikras/redux-form/issues/4293

**Steps to reproduce**
```bash
$ npx create-react-app form
$ cd form
$ yarn add redux redux-form react-redux
# Copy code from demo index.js and Form.js
$ yarn start
```

So, about bug:
```javascript
import { reduxForm, formValues } from 'redux-form';
/* ... */
export default reduxForm({ form: 'Form' })(formValues('name')(Form));
```
After update to version 8.0.1 using `formValues` decorator results in an error:
```
Uncaught Error: formValues() must be used inside a React tree decorated with reduxForm()
```

dagstuan:
The issue still exists in 8.0.4, working (broken) example: https://codesandbox.io/s/21m68rvozn

erikras:
Confirmed. Will investigate tomorrow.

tyscorp:
It looks like the consumer is being called before the provider, even though the provider is at a higher level in the tree. I have no idea how/why.

erikras:
It's got to be something to do with using `Context.Provider` rather than just a HOC to create the context. Hmm... You'll have to put the component connected to `formValues()` deeper in the tree.

Until I can figure it out, here's a little render-prop workaround for you.

[![Edit Redux Form - Simple Example](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/8k766888mj)

tyscorp:
Here is a version of that using recompose's [toRenderProps](https://github.com/acdlite/recompose/blob/master/docs/API.md#torenderprops) and [fromRenderProps](https://github.com/acdlite/recompose/blob/master/docs/API.md#fromrenderprops) that allows the same usage as before.

This is very strange. I wonder if it's a bug in React?

tyscorp:
@erikras Removing the `withReduxForm` from [this line](https://github.com/erikras/redux-form/blob/755b670864c52d544d6e63c54d33844ed3772b1c/src/createReduxForm.js#L1152) fixes the issue.

erikras:
> This is very strange. I wonder if it's a bug in React?

I assume there's no way that @gaearon can have the time to check his GH mentions, but...?

cajoy:
[...]

PR #4464, opened 2019-05-24
createReduxForm: Don't put WrappedComponent in a withReduxForm

asazernik:
withReduxForm is for internal redux-form components that need to
interact directly with the guts of the form state, not for access by
user components.

Fixes (at least in some cases?) #4299.

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -144,7 +144,6 @@ const describeReduxForm = (name, structure, combineReducers, setup) => {
     it('should provide the correct props', () => {
       const props = propChecker({})
       expect(Object.keys(props).sort()).toEqual([
-        '_reduxForm',
         'anyTouched',
         'array',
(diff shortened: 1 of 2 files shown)


renatoagds:
@asazernik code climate pointed a problem, could you fix it? (https://codeclimate.com/github/erikras/redux-form/pull/4464)

asazernik:
@renatoagds whoop. Removed the usage, forgot to remove the import. Fixing.

renatoagds:
@asazernik Looking more close to your PR, your changes will disconnect all forms from ReduxFormContext. Since `createReduxForm` return `ReduxForm` (https://github.com/asazernik/redux-form/blob/master/src/createReduxForm.js#L1175). If we disconnect them, we will break the Forms. I don't know why all tests are passing. We should test this scenario. Anyway, sorry to not get this before. We should go ahead with the fix proposed here: https://github.com/erikras/redux-form/pull/4448.

PR #4468, opened 2019-05-26
Better Config typing

asazernik:
I'm working on putting less extraneous cruft into ReduxFormContext to avoid unnecessary re-renders. (These were causing bugs in my project.) Since this is a bit of a scary refactor, wanted to make sure (internal-use-only) Flow types were up to snuff first.


src/createReduxForm.js
```diff
@@ -170,115 +170,120 @@ type OnChangeFunction = (
...
-  onSubmitFail?: OnSubmitFail,
-  onSubmitSuccess?: OnSubmitSuccess,
-  propNamespace?: string,
-  submitAsSideEffect?: boolean,
-  validate?: ValidateFunction,
-  warn?: ValidateFunction,
-  touchOnBlur?: boolean,
-  touchOnChange?: boolean,
-  persistentSubmitErrors?: boolean,
-  registeredFields: any
```
renatoagds: You forgot to put `registeredFields` back. It's missing in `Config`.

renatoagds:
@asazernik do you have plans to work in the fix?

src/createReduxForm.js
asazernik: It's in Props (in that big literal that comes after `OptionalConfig` in the intersection type chain). As far as I understand it isn't something that a caller passes into `@reduxForm`. Are you seeing any false positives from the typechecker?

asazernik:
Agh! Missed the previous comment. Replied inline.

src/createReduxForm.js
asazernik: (Pushed up a new version with comments describing each type's purpose.)

renatoagds:
Published in [`v8.2.6`](https://github.com/erikras/redux-form/releases/tag/v8.2.6).

PR #4466, opened 2019-05-24
Revert "disable loose mode for @babel/plugin-transform-spread"

renatoagds:
Reverts erikras/redux-form#4337 since it cause a problem in FieldArray.
Closes #4439 #4456 #4457

Issue #4439, opened 2019-04-16
FieldArray broken in v 8.2


danielpowell4:
### Are you submitting a **bug report** or a **feature request**?
bug report

### What is the current behavior?
Pushing a new field into a field array is broken

1. Visit https://redux-form.com/8.2.0/examples/fieldarrays/
1. Click "Add Member"
1. See `TypeError: Invalid attempt to spread non-iterable instance` in console

### What is the expected behavior?
A new member field should be pushed into the field array

### Sandbox Link
https://codesandbox.io/s/nrzq9mzp34
https://nrzq9mzp34.codesandbox.io/

### What's your environment?
Redux-form v8.2 (not 8.1)
At least Chrome 73, Safari 12.1, Firefox 65

### Other information
<img width="604" alt="Screen Shot 2019-04-15 at 10 51 50 PM" src="https://user-images.githubusercontent.com/15863213/56182811-15cbeb80-5fd1-11e9-9404-4045e4337d83.png">

Recent _potentially_ related commit https://github.com/erikras/redux-form/commit/fc3b48d0557c4bd0dac27cc5f6c22d984fdf1860 I say potentially as I see it was added in 8.2 and has the word FieldArray in it. Haven't wrapped my head around it all

nikhilag:
This is a major issue and I can confirm that it's not working 8.0.0 onwards. Must be something more fundamental due to v8 change.

danielpowell4:
@nikhilag I have a simple codesandbox working on 8.1

https://codesandbox.io/s/30xx4588lm

Fairly certain was something with the commit I referenced above

taylorqj:
Experiencing the same error on `8.1`.

SharudAgarwal:
Anyone have a workaround for now?

tomasztomys:
+1

rovnyart:
same issue here :(

darkestblue91:
Today I have to validate if one FieldArray has elements, using this (as the documentation says):
` 
if (!values.options || !values.options.length) {
errors.options = { _error: 'You have to provide at least one option.' };
}
`
I note that if I remove that condition on the validator, everything works well, so maybe that's a start to detect where is the bug or at least get a workaround.

kurtfurbush:
+1

dmason30:
@erikras Any chance you could take a look?

[...]

PR #4470, opened 2019-05-28
Run tests using same babel config as built product

esetnik:
Run tests using same babel config as built product. This was created in response to https://github.com/erikras/redux-form/issues/4439#issuecomment-496594691. The goal here is prove that ImmutableJS is now broken as of v8.2.1 via a set of tests which fail after #4337 was reverted by https://github.com/erikras/redux-form/pull/4466.

During investigation, I was going to write an immutable test that fails but instead I found that the babel configuration was incorrectly running with `{ targets: { node: '8' } }`. The test suite should run using the same target as the built product to eliminate scenarios where babel either fixes or introduces new issues in the built product that do not fail during testing. One can see that this branch / PR is failing the test suite due to ImmutableJS errors.

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -9,8 +9,7 @@ module.exports = {
       '@babel/preset-env',
       {
         modules: isEs ? false : 'commonjs',
-        loose,
-        ...(test ? { targets: { node: '8' } } : {})
+        loose
       }
     ],
     '@babel/preset-react',


Asday:
>Tests:       3 failed, 1798 passed, 1801 total


That's looking like some fantastic work, thank you @esetnik.

(Not that I understand it, yet).

esetnik:
@asday you will see in my original issue #4313 that I did a deep investigation of the issue with ImmutableJS + Babel in this project. All this PR does is make the test run against the same code that is output by Babel using the config specified in @babel/preset-env.

erikras:
Published in [`v8.2.2`](https://github.com/erikras/redux-form/releases/tag/v8.2.2).

PR #4475, opened 2019-05-30
Revert "Run tests using same babel config as built product"

erikras:
Reverts erikras/redux-form#4470. This broke v8.2.2 and needs more examination.


renatoagds:
Published in [`v8.2.3`](https://github.com/erikras/redux-form/releases/tag/v8.2.3).

PR #4492, opened 2019-06-19
fix: re-bind actionCreators to changed props.form

opichals:
The `mapDispatchToProps` was only called once as it was returning a `() => computedActions` cached function. But in order to reflect a `form` property updates when one re-uses the decorated `reduxForm()(DetailComponent)` instance the bindings must be re-bound to the changed `form` prop value.


opichals:
Updated with a test for the change.

renatoagds:
Published in [`v8.2.2`](https://github.com/erikras/redux-form/releases/tag/v8.2.5).

PR #4498, opened 2019-06-27
feat(reducer/clear-fields): set to initial values if present, otherwise cleanup

### LINKED ISSUE #4101: clearFields doesn't clear field value

Bug.

If a form is initialized from state, then a call to clearFields doesn't do anything.

clearFields should clear field values specified in parameters.

### Sandbox Link

https://codesandbox.io/s/8lq6w36xkl

### What's your environment?

redux-form 7.4.2 (but it's working for 7.3.0 https://codesandbox.io/s/42mjmk0mxw)

hugojanruiter:
where clearFields did not work as expected with initial values
Basically I have copied the logic from resetSection for resetting to initial values

Closes #4101

### DIFF
--- a/docs/api/ActionCreators.md
+++ b/docs/api/ActionCreators.md
@@ -74,10 +74,10 @@ actions such as `CHANGE` or `BLUR`, the specific field.
 
 ### `clearFields(form:String, keepTouched: boolean, persistentSubmitErrors: boolean, ...fields:String)`
 
-> Cleans fields values for all the fields passed in.
+> Cleans fields values for all the fields passed in. Will reset to initialValue for each field if has any.
 
 > If the `keepTouched` parameter is `true`, the values of currently touched
-> fields will be retained. 
+> fields will be retained.
 
 > If the `persistentSubmitErrors` parameter is `true`,
 > the values of currently submit errors fields will be retained
--- a/docs/api/Props.md
+++ b/docs/api/Props.md
@@ -127,9 +127,9 @@ class SimpleForm extends React.Component {
 
 ### `clearFields(keepTouched: boolean, persistentSubmitErrors: boolean, ...fields:String)`
 
-Cleans fields values for all the fields passed in. Will reset to initialValue for each field if has any.  
-If the keepTouched parameter is true, the values of currently touched fields will be retained.  
-If the persistentSubmitErrors parameter is true, the values of currently submit errors fields will be retained
+> Cleans fields values for all the fields passed in. Will reset to initialValue for each field if has any.  
+> If the keepTouched parameter is true, the values of currently touched fields will be retained.  
+> If the persistentSubmitErrors parameter is true, the values of currently submit errors fields will be retained
 
 ### `destroy() : Function`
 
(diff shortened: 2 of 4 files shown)


hugojanruiter:
Sure thing. But is the new test case reflecting the desired behaviour? I have assumed that we would like to reset it to initial value, but I later thought that with clearing, we really want to clear the field and ignore inital values.

renatoagds:
@hugojanruiter I think the tests should follow what you said, to clear them instead back to initial value. Do you have plans to work on these fixes?

tnrich:
>@renatoagds, sure thing. But is the new test case reflecting the desired behaviour? I have assumed that we would like to reset it to initial value, but I later thought that with clearing, we really want to clear the field and ignore inital values.

Just to add my 2 cents, it seems like _both_ are valid use cases. I think we should add an option to  clearFields to allow the user to reset it to the initial value. That would be a breaking change as it would change the function signature to something like this: 

```
clearFields(form:String, keepTouched: boolean, keepInitial: boolean, persistentSubmitErrors: boolean, ...fields:String)

//If the keepInitial parameter is true, the field will be reset to its initial value or undefined if no initial value has been set. 
```



If we don't want to break the existing api like that, we could also add another function resetFields which is just like clearFields but keeps initialValues:
```
resetFields(form:String, keepTouched: boolean, persistentSubmitErrors: boolean, ...fields:String)
```

Thoughts?

iamandrewluca:
I opt for what @tnrich said
* `clearFields` removes fields valus entirely
* add `resetFields` that will reset fields to `initial`

PS: `props.clearFields` is missing from docs

renatoagds:
@tnrich @iamandrewluca we're have a reset function by [`action creator`](https://redux-form.com/8.2.2/docs/api/actioncreators.md/#-code-reset-form-string-code-) and [`props`](https://redux-form.com/8.2.2/docs/api/props.md/#-code-reset-function-code-)

> PS: props.clearFields is missing from docs

@iamandrewluca `clearFields` isn't a prop, it's a [`action creator`](https://redux-form.com/8.2.2/docs/api/actioncreators.md/#-code-clearfields-form-string-keeptouched-boolean-persistentsubmiterrors-boolean-fields-string-code-)

iamandrewluca:
I mean, there is a `Form.props.clearFields` already binded, and is missing from the docs, like other binded actions

tnrich:
@renatoagds in the docs there is indeed a reset function:

```
reset(form:String)
Resets the values in the form back to the values passed in with the most recent initialize action.

resetSection(form:String, ...sections:String)
```

however this reset function can't take a list of fields like clearFields can: 

```
clearFields(form:String, keepTouched: boolean, persistentSubmitErrors: boolean, ...fields:String)
```

I think either we need to add a way for reset to to take a list of fields or we need to add a new resetFields function. Or like I said originally we could add a way for clearFields() to also optionally clear or reset depending on a flag but that would have to break the existing api.

renatoagds:
@tnrich Agreed with you, but I prefer following the way to add a `resetFields` that's only responsible for that. But, this shouldn't infer in this Pull Request.

tnrich:
@renatoagds I've added a new issue asking for a resetFields() function -- https://github.com/redux-form/redux-form/issues/4543

renatoagds:

PR #4499, opened 2019-06-28
Use lastSyncWarnings in updateSyncWarningsIfNeeded

### LINKED ISSUE #4488: Warnings with initial values not calculated correctly

Bug report

When providing initial values to a form only the errors are calculated for those values, the warnings are initially calculated as if no initial values were given. Only after the first change to any form value the warnings are displayed correctly.
Example:
1. Create a form with warning validation which requires a value in a certain field.
2. Initialize the form with an initial value in that field.
3. When rendering the form the field receives a "Required"-warning even though it is correctly filled.
4. (Type anything into any field - the "Required"-warning disappears.)

Error and warn validation are expected to work the same. With initial values given both validation levels should be calculated based on those values.

### Sandbox Link

[https://codesandbox.io/s/redux-form-initial-warning-bug-rrmw3](https://codesandbox.io/s/redux-form-initial-warning-bug-rrmw3
)

### What's your environment?

    "react": "16.8.6",
    "react-dom": "16.8.6",
    "react-redux": "7.0.2",
    "redux": "4.0.1",
    "redux-form": "8.2.0"

(not OS/browser related)

### Other information

The behavior seems related to the function `updateSyncWarningsIfNeeded` in `createReduxForm`
When changing the line:
`const noWarnings = (!syncWarnings || !Object.keys(syncWarnings).length) && !warning`
to:
`const noWarnings = (!lastSyncWarnings || !Object.keys(lastSyncWarnings).length) && !warning`
everything works as expected.
That's also consistent to the logic in `updateSyncErrorsIfNeeded` (which would explain why that problem doesn't show with errors but only with warnings).
The problem only occurs if warn-validation with empty values returns warnings, but warn-validation with initial values returns (should return) none.

renatoagds:
`updateSyncWarningsIfNeeded` was using `syncWarnings` from `this.props`, but using it, in `componentWillReceiveProps` we won't have the value that we expected (since it's in `nextProps`), instead that, we use `lastSyncWarnings` that point for `nextProps`.

Closes #4488

### DIFF
--- a/src/__tests__/reduxForm.spec.js
+++ b/src/__tests__/reduxForm.spec.js
@@ -3874,6 +3874,38 @@ const describeReduxForm = (name, structure, combineReducers, setup) => {
       expect(propsAtNthRender(renderInput, 1).meta.warning).toEqual(warning)
     })
 
+    it.only('should not warn with initialValues', () => {
+      const store = makeStore({})
+      const formRender = jest.fn()
+      const renderInput = jest.fn(props => <input {...props.input} />)
+
+      class Form extends Component {
+        render() {
+          formRender(this.props)
+          return (
+            <form>
+              <Field name="foo" component={renderInput} type="text" />
+            </form>
+          )
+        }
+      }
+      const Decorated = reduxForm({
+        form: 'testForm',
+        initialValues: { foo: 'test' },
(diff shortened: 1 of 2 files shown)


renatoagds:
Published in [`v8.2.5`](https://github.com/erikras/redux-form/releases/tag/v8.2.5).

PR #4507, opened 2019-07-19
fix(reducer/change): allow `change` to handle `undefined` payload

### LINKED ISSUE #3946: Calling Field's input.onChange(undefined) does not set field value to undefined.

tnrich:
Closes #3946

### DIFF
--- a/src/__tests__/helpers/reducer.change.js
+++ b/src/__tests__/helpers/reducer.change.js
@@ -186,7 +186,6 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
     expect(state).toEqualMap({
       foo: {
         anyTouched: true,
-        values: {
         fields: {
           myField: {
             touched: true
--- a/src/__tests__/helpers/reducer.change.js
+++ b/src/__tests__/helpers/reducer.change.js
@@ -187,7 +187,6 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
       foo: {
         anyTouched: true,
         values: {
-        },
         fields: {
           myField: {
             touched: true
--- a/src/__tests__/helpers/reducer.change.js
+++ b/src/__tests__/helpers/reducer.change.js
@@ -172,7 +172,7 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
     })
   })
 
-  it('should not modify a value if called with undefined', () => {
+  it('should modify a value if called with undefined', () => {
     const state = reducer(
       fromJS({
         foo: {
@@ -187,7 +187,6 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
       foo: {
         anyTouched: true,
         values: {
-          myField: 'initialValue'
         },
         fields: {
           myField: {
--- a/src/createReducer.js
+++ b/src/createReducer.js
@@ -303,7 +303,7 @@ function createReducer<M, L>(structure: Structure<M, L>) {
     ) {
       let result = state
       const initial = getIn(result, `initial.${field}`)
-      if (payload === '') {
+      if ((initial === undefined && payload === '') || payload === undefined) {
         result = deleteInWithCleanUp(result, `values.${field}`)
       } else if (isFunction(payload)) {
         const fieldCurrentValue = getIn(state, `values.${field}`)
--- a/src/createReducer.js
+++ b/src/createReducer.js
@@ -303,7 +303,7 @@ function createReducer<M, L>(structure: Structure<M, L>) {
     ) {
       let result = state
       const initial = getIn(result, `initial.${field}`)
-      if (initial === undefined && payload === '') {
+      if (payload === '') {
         result = deleteInWithCleanUp(result, `values.${field}`)
       } else if (isFunction(payload)) {
         const fieldCurrentValue = getIn(state, `values.${field}`)
@@ -312,7 +312,7 @@ function createReducer<M, L>(structure: Structure<M, L>) {
           `values.${field}`,
           payload(fieldCurrentValue, state.values)
         )
-      } else if (payload !== undefined) {
+      } else {
         result = setIn(result, `values.${field}`, payload)
       }
       result = deleteInWithCleanUp(result, `asyncErrors.${field}`)


iamandrewluca:
Tests fail.

src/createReducer.js
```diff
@@ -303,7 +303,7 @@ function createReducer<M, L>(structure: Structure<M, L>) {
     ) {
       let result = state
       const initial = getIn(result, `initial.${field}`)
-      if (initial === undefined && payload === '') {
+      if (payload === '') {
```
iamandrewluca: to not make this a breaking change for thoose who use `''` to reset value use bellow suggestion
```suggestion
      if ((initial === undefined && payload === '') || payload === undefined) {
```

iamandrewluca: this will break only 2 tests, fix them to allow `undefined` value
```
  ● reducer.immutable › reducer.immutable.change › should allow setting an initialized field to ''
  ● reducer.immutable › reducer.immutable.change › should not modify a value if called with undefined
```

iamandrewluca: this should be `request change`

iamandrewluca: to not make this a breaking change for thoose who use `''` to reset value use bellow suggestion
```suggestion
      if ((initial === undefined && payload === '') || payload === undefined) {
```

this will break only 2 tests, fix them to allow `undefined` value

```
  ● reducer.immutable › reducer.immutable.change › should allow setting an initialized field to ''
  ● reducer.immutable › reducer.immutable.change › should not modify a value if called with undefined
```

iamandrewluca: I think only `reset` should reset to `initialValue`, using `change` should allow any value to be changed

iamandrewluca: https://github.com/erikras/redux-form/blob/c89ee4473bc497f2740bc742df9e30b7263419d7/src/__tests__/helpers/reducer.change.js#L189
remove `values` and tests should work.

iamandrewluca: https://github.com/erikras/redux-form/blob/c89ee4473bc497f2740bc742df9e30b7263419d7/src/__tests__/helpers/reducer.change.js#L189
remove `values` and tests should work

iamandrewluca: to not make this a breaking change for thoose who use `''` to reset value use bellow suggestion
```suggestion
      if ((initial === undefined && payload === '') || payload === undefined) {
```

this will break only 2 tests, fix them to allow `undefined` value

```
  ● reducer.immutable › reducer.immutable.change › should allow setting an initialized field to ''
  ● reducer.immutable › reducer.immutable.change › should not modify a value if called with undefined
```

I think only `reset` should reset to `initialValue`, using `change` should allow any value to be changed. But this will be a breaking change

https://github.com/erikras/redux-form/blob/c89ee4473bc497f2740bc742df9e30b7263419d7/src/__tests__/helpers/reducer.change.js#L189
remove `values` and tests should work

tnrich:
@iamandrewluca I tried to make the changes but I'm not sure if I got them all. Could you let me know if I need to make anymore?

tnrich:
Also, @iamandrewluca it looks like the tests are failing for an unrelated error: 

```
    console.error node_modules/jsdom/lib/jsdom/virtual-console.js:29
      Error: Uncaught [Error: Fields must be inside a component decorated with reduxForm()]
```

Maybe I need to merge in the latest commits on master?

src/__tests__/helpers/reducer.change.js
```diff
@@ -187,7 +187,6 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
       foo: {
         anyTouched: true,
         values: {
```
iamandrewluca: in reducer is used function `deleteInWithCleanUp` to cleanup, from it tests I see that if an object has no children it also removes the parent. In this case you should also remove `values` object
```suggestion
```

src/__tests__/helpers/reducer.change.js
```diff
@@ -187,7 +187,6 @@ const describeChange = (reducer, expect, { fromJS }) => () => {
...
-          myField: 'initialValue'
         },
```
iamandrewluca:
```suggestion
```

iamandrewluca:
@tnrich those are just `console.error` from code, they are expected. I think we should run tests in silent mode to have a more cleaner output.

tnrich:
Okay awesome I applied your suggestions. Thanks! Is it all ready to go now?

iamandrewluca:
@tnrich yeap, we just need to wait for an project collaborator :)

tnrich:
Perfect thanks so much! I hope it'll go through!

iamandrewluca:
Would be great to have title more descriptive, may catch collaborators eyes
```
feat(reducer/change): allow `change` to handle `undefined` value
```
and include in body
```
Closes #3946
```

iamandrewluca:
Also I would opt to deprecate using this `initial === undefined && payload === ''` in `change`.  
This allows only to clean field when initial value is undefined.
User should take care to what value is changing.

tnrich:
> Also I would opt to deprecate using this initial === undefined && payload === '' in change.
This allows only to clean field when initial value is undefined.
User should take care to what value is changing.

How should I go about doing that? @iamandrewluca

iamandrewluca:
Wi'll think of that in a future refactor.

renatoagds:
Published in [`v8.2.6`](https://github.com/erikras/redux-form/releases/tag/v8.2.6).

PR #4514, opened 2019-07-29
wip: test(ci): run tests in silence, keep output clear

iamandrewluca:
~~Using `--silent` flag, `console.{log,error,warn,...}` will not be outputted.~~  
Failled tests will still be showed.  
Making CI logs cleaner helps new comers to understanda better CI logs.

Notes:
https://gist.github.com/gaearon/adf9d5500e11a4e7b2c6f7ebf994fe56
https://github.com/facebook/react/issues/11098#issuecomment-370614347
https://github.com/esphen/jest-prop-type-error
https://github.com/facebook/prop-types/issues/28
https://medium.com/shark-bytes/type-checking-with-prop-types-in-jest-e0cd0dc92d5

### DIFF
--- a/src/__tests__/Field.spec.js
+++ b/src/__tests__/Field.spec.js
@@ -12,6 +12,7 @@ import plain from '../structure/plain'
 import plainExpectations from '../structure/plain/__tests__/expectations'
 import immutable from '../structure/immutable'
 import immutableExpectations from '../structure/immutable/__tests__/expectations'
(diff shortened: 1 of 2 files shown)


renatoagds:
IMO doing this, we're masking warnings and problems that our code has. If we have warnings from React, we're doing something wrong, we should fix them and not silent them.

iamandrewluca:
@renatoagds Agree, then I'll refactor this PR into fixing those warnings and errors.

PR #4556, opened 2019-10-23
chore: rename unsafe lifecycle methods to silence console warnings for v5.4.0

kelly-tock:
I would like to publish a version for v5 that renames the lifecycle methods. all tests passing, and ran the code mod similar to:
https://github.com/redux-form/redux-form/pull/4520/files

- had to update react for this to work in devDeps
- import TestUtils from react-dom as a result.


iamandrewluca:
@kelly-tock you are using `react@^16` ? why not better upgrade to latest `redux-form`, we can't keep all versions up to date with react, and this would be a breaking change.

@renatoagds @erikras how this should be handled?

kelly-tock:
Our code base quite frankly is too large to do that ATM. We would like a patch to give us some time for when react 17 comes out

kelly-tock:
added minimum peer dependencies, would that help? this is a weird scenario where the library is changing to be compatible with a newer version of a library. at least this way it would show that folks would need to be on `>= react 16.9` for a 5.4.1(or whatever semver makes sense).

renatoagds:
@kelly-tock This won't help since we'll sting bumping `react` and `react-dom` in `dependencies` to `16`, if someone else still using react 15, this will break their applications, mainly because isn't everyone that use pinned version, and the default install for `npm` is `^`, that updates for `patch` and `minor` releases and like @iamandrewluca said this is a major release. Unfortunately that's not so much we can do. I suggest you fork the repo and release your own version with this update.

kelly-tock:
pretty crazy that  npm and semver doesn't support this use case. we can't all upgrade all the time for every package. :( exactly what I was going to have to do.

iamandrewluca:
No language, or package manager does support this. This is how semver works. we have it for a reason 🙂

PR #4557, opened 2019-10-24
prevent shipping eval sourcemaps in UMD bundle

henryqdineen:
This probably doesn't effect many users but we are bundling the UMD build of this library and I noticed that the `dist/redux-form.js` was very large and included the `eval()` sourcemaps. From looking at other popular open source `react` libraries it looks like shipping `eval()` sourcemaps is not common. Thanks!


iamandrewluca:
@henryqdineen `'none'` is an invalid `devtool` option. We don't include any source maps at all.
`devtool` is `false` by default.
https://webpack.js.org/configuration/devtool/#devtool

Mostly is something wrong in your production config, that generates the sourcemaps

henryqdineen:
@iamandrewluca do you see the `eval()` sourcemaps here: https://unpkg.com/browse/redux-form@8.3.2/dist/redux-form.js

iamandrewluca:
@henryqdineen you are right, I will change it to `source-map` and they will be generated in separate file. `none` is not a valid option. Strange why `false` is not set by default.

iamandrewluca:
@henryqdineen Thanks!

henryqdineen:
Thanks! Sorry about the `'none'` confusion. 🙂

PR #4596, opened 2020-01-10
fix(field): use onChange return as preventDefault only in RN

### LINKED ISSUE #4391: react-native if onChange callback is async ConnectedField won't trigger the change action

bug report

If onChange prop is async function ConnectedField does not trigger the change action. I belive this is because async function always returns a Promise instance so this condition is always true
https://github.com/erikras/redux-form/blob/84359b3fb5be9f62e030657100ea1ddede4c5ef7/src/ConnectedField.js#L132

Maybe we should check for Promises here and wait resolve them and then use their result as value for defaultPrevented variable?

### Sandbox Link
https://codesandbox.io/s/l7vp6w6nyz

App.js
Line: 16

### What's your environment?
react-native, both iOS and Android

### LINKED ISSUE #4401: `onChange` prop with a returned value breaks custom fields
### Bug Report
If you define a custom `Field` component that calls `onChange` with the value of the field, providing an `onChange` prop to the field will prevent the `@@redux-form/CHANGE` event from firing, IF that `onChange` function returns an explicit value. 

I believe this may be related to https://github.com/erikras/redux-form/issues/3446

I've provided a [minimal Sandbox](https://codesandbox.io/s/n9kk00qw9j) to reproduce. Say I've created a custom field component called `Number` that is supposed to take the value of the input when it changes and update some other field.
```
const Number = ({ input }) => (
  <input
    type="number"
    {...input}
    onChange={({ target }) => input.onChange(target.value)}
  />
);

...

<Field
  name="number"
  component={Number}
  onChange={(val) => change("my-form", "some_other_field", val + 1)}
/>
```

This will prevent `@@redux-form/CHANGE` from firing because it returns a value, rendering the field effectively dead. However, if the `onChange` prop in `Field` returns undefined, the whole thing works just fine:
```
<Field
  name="number"
  component={Number}
  onChange={(val) => {
    change("my-form", "some_other_field", val + 1);
    return undefined;
  }}
/>
```

I would expect a function provided as an `onChange` prop to `Field` to take the value provided and perform whatever operation with it, regardless of the return value of that function. There's a workaround to this, but it's very weird implicit behavior that I now need to document for future developers who might want to use these components.

In my case, I would like to be able to provide custom field components that fire `onChange` with their current value and then be able to perform further operations with that particular value. 

### Sandbox Link
https://codesandbox.io/s/n9kk00qw9j

### What's your environment?
```
"dependencies": {
    "react": "16.4.0",
    "react-dom": "16.4.0",
    "react-redux": "6.0.1",
    "redux": "3.5.2",
    "redux-form": "8.1.0"
  }
```

### Other information
I believe I've provided the most minimal case for this bug. Please let me know if there is any other information I can provide.

timhwang21:
Fixes regression introduced in 8.1.0.

Issue #4309 adds a way to cancel dispatches from RN. However, the method
added also impacts React Web. This leads to a regression in React Web
due to application of RN behavior.

The current change locks the change from #4309 to RN, and maintains
existing behavior otherwise.

Related:
- #4391 outlines a case where this regression may cause problems. While
  the issue mentions RN, this can easily occur outside RN as well.
- #4401 describes a similar issue and a user-end fix.

Closes #4391 
Closes #4401

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -126,7 +126,12 @@ const createConnectedField = (structure: Structure<*, *>) => {
             name
           )
         } else {
-          defaultPrevented = onChange(event, newValue, previousValue, name)
+          const onChangeResult = onChange(event, newValue, previousValue, name)
+          // Return value of change handler affecting preventDefault is RN
+          // specific behavior.
+          if (isReactNative) {
+            defaultPrevented = onChangeResult
+          }
         }
       }
       if (!defaultPrevented) {


iamandrewluca:
@timhwang21 Thanks for the fix! I'll take a look tommorow!

renatoagds:

PR #4604, opened 2020-02-05
fix: send an event-like object into Field onChange when no event (#4548)

### LINKED ISSUE #4548: event.preventDefault is not a function in onChange(event, value, previousValue, name)
after upgrade `react-redux` to version `7.1.1` and `redux-form` to version `8.2.6`...
i found that preventDefault object is missing from the event params...

Here is my code...

```
<Field
     component={ReduxFormRadioGroup}
     formItemProps={{ className: 'operation-hours__option-container' }}
     label='Hours of Operation'
     name='operationHoursCategory'
     onChange={this.handleCategoryChange}
/>

handleDisabledDayChange = (event, value, previousValue, name) => {
        console.log(event, value, previousValue, name)
        event.preventDefault()
        ...
        .......
}
```

in `"redux-form": "^7.2.0"` and `"react-redux": "^5.0.7"`:
console.log(event) can see the `preventDefault: ƒ preventDefault()`

in `redux-form": "^8.2.6"` and `"react-redux": "^7.1.1"`:
console.log(event) do not have `preventDefault: ƒ preventDefault()`

this in result `event.preventDefault()` is not a function and break the application!

dereksandkamp:
The docs mention that it is possible for an input to call `input.onChange` and pass either an event or a new value as the argument.

They also mention that it's possible to pass `onChange` to a `<Field />` component, and that handler should receive an event as its first argument, with a `preventDefault` method on it which will prevent the change from being dispatched.

It's not explained in the docs that there's any connection between these two APIs. However, since [this change](https://github.com/redux-form/redux-form/pull/4170/files), the consumer only receives an event object in `Field.onChange` if the underlying input has called `input.onChange` with an event.

Since many people (including myself) have been relying on the ability to BOTH pass a value to `input.onChange` and call `event.preventDefault()` within the `Field.onChange` handler, I think we should either make that possible or update the docs to explain the restriction. This PR makes it possible.

Closes #4548

### DIFF
--- a/src/ConnectedField.js
+++ b/src/ConnectedField.js
@@ -72,16 +72,16 @@ const createConnectedField = (structure: Structure<*, *>) => {
       return !!(
         this.props.children ||
         nextProps.children ||
-        (nextPropsKeys.length !== thisPropsKeys.length ||
-          nextPropsKeys.some(prop => {
-            if (~(nextProps.immutableProps || []).indexOf(prop)) {
-              return this.props[prop] !== nextProps[prop]
-            }
-            return (
-              !~propsToNotUpdateFor.indexOf(prop) &&
-              !deepEqual(this.props[prop], nextProps[prop])
-            )
-          }))
+        nextPropsKeys.length !== thisPropsKeys.length ||
+        nextPropsKeys.some(prop => {
+          if (~(nextProps.immutableProps || []).indexOf(prop)) {
+            return this.props[prop] !== nextProps[prop]
+          }
+          return (
+            !~propsToNotUpdateFor.indexOf(prop) &&
+            !deepEqual(this.props[prop], nextProps[prop])
+          )
+        })
       )
     }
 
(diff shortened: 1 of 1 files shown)


machinek56:
Hi guys! Any updates on this PR?
@dereksandkamp @iamandrewluca

iamandrewluca:
I'll try to find some time this week to review this. From what I saw it seems this may be a breaking change.
Implicated in some urgent COVID-19 projects and don't have to much time to allocate for `redux-form`

dereksandkamp:
Thanks! For what is worth, from my perspective the breaking change already happened and this is a fixing change 😆

machinek56:
@dereksandkamp, for now i wrapped your PR into [own package](https://www.npmjs.com/package/@machinek56/redux-form) as we extremely need it.

duyenho:
👋 @iamandrewluca and @dereksandkamp 
cc: @machinek56 

Hope everyone is safe during COVID-19.
Would love to get this one in as well 🙂

PR #4637, opened 2020-03-19
perf: improve actionTypes exports

### LINKED ISSUE #4618: Tree shaking not working properly with webpack
### Subject of the issue

If you only import `import { actionTypes } from 'redux-form'` you'd expect your end bundle to only contain the `actionTypes` file but the way the export are written in `index.js` breaks the tree-shaking of webpack and the end chunks also contains the actions.

In webpack this can lead to breaking the chunk completly when used in a dll. See related issue in webpack https://github.com/webpack/webpack/issues/10475

To fix it, the export of `actionTypes` that is like this:
```js
// index.js
import * as _actionTypes from './actionTypes'
export const actionTypes = _actionTypes
```

should be like this:

```js
// index.js
export { default as actionTypes } from './actionTypes';
```

This just requires to add a default export in `./actionTypes`.

### Your environment

- **OS**: macOS
- **Packages**: see repo below
- **Env**: node 12/npm

### Steps to reproduce

You can checkout the repo https://github.com/skimi/redux-form-tree-shaking-broken

### Expected behaviour

When importing { actionTypes } in a module end bundle should only contain the actionTypes module.

### Actual behaviour

Other modules from `redux-form` get pulled in, the tree-shaking is not working

### Other information

Webpack issue: https://github.com/webpack/webpack/issues/10475

iamandrewluca:
Closes #4618


renatoagds:
Published in [`v8.3.2`](https://github.com/redux-form/redux-form/releases/tag/v8.3.2).

PR #4648, opened 2020-04-08
fix: update lodash for redux-form@7 (CVE-2019-10744)

darcusfenix:
this upgrade was fixed in 8x.
https://github.com/redux-form/redux-form/commit/d0053a0e99dce46c2d4eb6059866407e863c4100

but what about 7x? 
i cannot upgrade to 8x in this moment


iamandrewluca:
Thanks!

PR #4661, opened 2020-04-13
refactor: upgrade to exact-by-default object type syntax

### LINKED ISSUE #4539: Mark flow types as inexact with the new `...` syntax
Feature request

The flow typings for objects in redux-form are currently all inexact (i.e. they're defined as `{}` instead of `{| |}`). Example: https://github.com/erikras/redux-form/blob/master/src/FieldProps.types.js.flow 

Flow is gradually moving towards "exact by default" behaviour. The new syntax for inexact objects is `{ foo: number, bar: string, ... }` See https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf for more info on this.

Flow types for objects in redux-form should have `...` to signify that they're inexact. This helps codebases transition to "exact by default" behaviour. Without `...`, Flow considers those objects exact which usually results in a fair number of errors.

### What's your environment?
master branch

### Other information
Adding `...` might require updating Flow and Babel to the relatively recent versions to support this syntax.

iamandrewluca:
https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf
https://medium.com/flow-type/how-to-upgrade-to-exact-by-default-object-type-syntax-7aa44b4d08ab

Closes #4539

### DIFF
--- a/src/Form.js
+++ b/src/Form.js
@@ -4,9 +4,7 @@ import PropTypes from 'prop-types'
 import { withReduxForm } from './ReduxFormContext'
 import type { ReactContext } from './types'
 
-export type Props = {
-  onSubmit: Function
-}
+export type Props = { onSubmit: Function, ... }
 
 type PropsWithContext = ReactContext & Props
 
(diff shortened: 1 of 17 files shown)


iamandrewluca:
Won't do for now. This will imply a breaking change.

Issue #4435, opened 2019-04-12
Bug: async validation showing up in error reporting service


tstirrat15:
### What is the current behavior?
On our app, we use [Sentry](https://github.com/getsentry/sentry-javascript) to catch errors, which includes unhandled promise rejections. We recently upgraded to the most recent release of that package (5.0.6), and we've started seeing Sentry reporting the error objects thrown in `asyncValidate` functions as unhandled promise rejections. An example:

This code:
```javascript
export const asyncRegistrationValidation = async ({ phone, email }, dispatch, { isLoggedIn }) => {
  const response = await apiFetch(
    '/auth/validate/registration/',
    {
      method: 'POST',
      body: JSON.stringify({
        // Don't validate emails of logged-in users
        email: !isLoggedIn ? email : undefined,
        phone,
      }),
    },
  );
  const body = await response.json();
  if (response.status === 400) {
    const error = {};
    if (body.email) {
      error.email = (
        <EmailErrorMessage
          errorMessage={body.email[0]}
        />
      );
    }
    if (body.phone) {
      error.phone = body.phone[0];
    }
    throw error;
  }
};
```
Which matches the convention in the [asyncvalidation](https://redux-form.com/8.2.0/examples/asyncvalidation/) docs.

This error rejection is happening on submit, which points to this code: https://github.com/erikras/redux-form/blob/b46bd11727664d92a20400316095213c93d7587f/src/handleSubmit.js#L114

Note that this block ends with a `Promise.reject`: https://github.com/erikras/redux-form/blob/b46bd11727664d92a20400316095213c93d7587f/src/handleSubmit.js#L128

And is not caught by the calling code: https://github.com/erikras/redux-form/blob/fc3b48d0557c4bd0dac27cc5f6c22d984fdf1860/src/createReduxForm.js#L814

### What is the expected behavior?
Somewhere in this chain, the promise rejection should be caught and handled. Am I missing something?

### Sandbox Link
This issue should theoretically show up in https://codesandbox.io/s/nKlYo387. I'm not sure why it doesn't, unless there's some mechanism by which Sentry is listening to something that the console isn't.
### What's your environment?
OS: Ubuntu 18.10
Browser: Chrome 73.0.3683.103
redux-form: 8.2.0
@sentry/browser: 5.0.6

jamesmillerburgess:
We are also encountering lots of these error reports.

The uncaught promise shows up for me when I use the form in the docs in a certain way:
[...]

PR #4675, opened 2020-06-02
fix: don't reject asyncErrors

jamesmillerburgess:
Closes #GH-4435

This line was causing unhandled promise rejections any time `asyncValidate` threw an error during a submit.

We have tested this fork and it made the issue go away, without affecting the errors appearing correctly in our forms.

I also noticed that the tests in `handleSubmit.spec.js` didn't have `expect.assertions` statements, which wasn't a problem for most due to a workaround with throwing inside `then`.

However, this is still a brittle mechanism, so I added `expect.assertions` to make it more robust. I'm happy to remove that if it shouldn't be in the scope of the PR.


Imjamper:
Hi there, any update on this? When this can be merged?

PR #4681, opened 2020-07-01
Removed scarf analytics

erikras:
[Reasoning here](https://twitter.com/erikras/status/1278071436754575360).


opichals:
@erikras Thank you for dropping this.

There is other useful stuff done in `8.3.6` which is hindered by the scarf addition. Could you please create a release including this change?

PR #4684, opened 2020-07-16
Add possibility to change function passed to handleSubmit

### LINKED ISSUE #4682: handleSubmit fires cached version when form is submited by dispatch
### Subject of the issue
It looks like submit by `<button type='submit' />` and `dispatch(submit('form'))` behaves differently. While submiting by button is ok, submiting using dispatch fires cached function (first function that is passed to `handleSubmit` function)

### Your environment

| Software              | Version(s) |
| --------------------- | ---------- |
| `redux-form`          | 8.3.6 
| `redux`               | 4.0.5
| `react-redux`         | 7.2.0
| `react`               | 16.12.0
| `react-dom`           | 16.12.0

### Steps to reproduce

https://codesandbox.io/s/ecstatic-leftpad-54u9k
Fill any data in form, click `Submit` - it should log in console your data. Now try `Submit by dispatch` - it logs `undefined`.

### Expected behaviour

Submit by button and dispatch should fire the same function passed to handleSubmit.

### Actual behaviour

Submit by dispatch fires cached function.

ideffix:
Closes #4682

When submiting form remotely by dispatching `submit` action and during `Form` lifetime and method provided to handleSubmit has changed - it isn't fired - only first function provided to `Form` is fired. After invastigation I've noticed that for remote submitting `submit` function from `createReduxForm.js` calls method `innerOnSubmit()` and it is set with following function: `registerInnerOnSubmit: innerOnSubmit => (this.innerOnSubmit = innerOnSubmit)`. 
In `Form.js` I can see that it only register method when component will mount ` this.props._reduxForm.registerInnerOnSubmit(this.props.onSubmit)` so even if you change method provided to handleSubmit during `Form` life, it wont be fired `onSubmit`. 
My fix adds possibility to change that method during `Form` life.

### DIFF
--- a/src/Form.js
+++ b/src/Form.js
@@ -14,14 +14,16 @@ class Form extends Component<PropsWithContext> {
   constructor(props: PropsWithContext) {
     super(props)
     if (!props._reduxForm) {
-      throw new Error(
-        'Form must be inside a component decorated with reduxForm()'
-      )
+      throw new Error('Form must be inside a component decorated with reduxForm()')
     }
   }
 
   UNSAFE_componentWillMount() {
     this.props._reduxForm.registerInnerOnSubmit(this.props.onSubmit)
+  }
+
+  UNSAFE_componentWillUpdate(nextProps: Props) {
+    this.props._reduxForm.registerInnerOnSubmit(nextProps.onSubmit)
   }
 
   render() {
--- a/src/Form.js
+++ b/src/Form.js
@@ -18,11 +18,11 @@ class Form extends Component<PropsWithContext> {
     }
   }
 
-  UNSAFE_componentWillMount() {
+  componentDidMount() {
     this.props._reduxForm.registerInnerOnSubmit(this.props.onSubmit)
   }
 
-  UNSAFE_componentWillUpdate(nextProps: Props) {
+  componentDidUpdate(nextProps: Props) {
     this.props._reduxForm.registerInnerOnSubmit(nextProps.onSubmit)
   }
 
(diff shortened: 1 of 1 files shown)


ideffix:
Hi @iamandrewluca, can we do something with it?

iamandrewluca:
We don't want to add any legacy things, this will make development of library event harder if we want to move on with next versions of react.

> Note:
> 
> These methods are considered legacy and you should avoid them in new code:
>
> UNSAFE_componentWillUpdate()
> UNSAFE_componentWillReceiveProps()
> — https://reactjs.org/docs/react-component.html

ideffix:
@iamandrewluca I've changed unsafe methods of this component

src/Form.js
```diff
@@ -14,16 +14,18 @@ class Form extends Component<PropsWithContext> {
...
     }
   }
 
-  UNSAFE_componentWillMount() {
+  componentDidMount() {
     this.props._reduxForm.registerInnerOnSubmit(this.props.onSubmit)
   }
 
+  componentDidUpdate(nextProps: Props) {
+    this.props._reduxForm.registerInnerOnSubmit(nextProps.onSubmit)
```
iamandrewluca, line 26: I think we should add a check here and register again only if onSubmit has changed

ideffix: Done

iamandrewluca:
Thankss @ideffix, I'll find some time in the evening to make a new release.

ideffix:
Hi @iamandrewluca, I know its been a while since this PR was merged to master, but can we have a realease version with it? I kinda need it now :D

erikras:
Published in [`v8.3.7`](https://github.com/redux-form/redux-form/releases/tag/v8.3.7).

iamandrewluca:
@ideffix FYI

ideffix:
Gr8, thanks guys! :1st_place_medal:

Issue #4412, opened 2019-03-26
module is not defined


wmertens:
When building with webpack 5, this line gives a `ReferenceError` for `module`: https://github.com/erikras/redux-form/blob/f209539ea65e1b46534e701c635df31c100c3259/src/util/isHotReloading.js#L3

I'm not sure how to fix this with the typing stuff intact, normally I would just put a `if (typeof module === 'undefined') return false` above that line.

wmertens:
Update: Simply removing the castModule and using `module` directly makes it work properly. So it's a matter of the TypeScript workaround making it not work with Webpack 5.

sbaechler:
Another fix is prefixing module with `globalThis`. I don't know Flow enough to create a merge request that passes the type system. But using patch-package I could simply change this line to get my app working with Vite.

     var castModule = globalThis.module

Another way I got it working was to define 

    module: undefined

in vite.config. However this might cause side-effects.

PR #4723, opened 2021-04-09
Avoid assuming that module exists

sholladay:
This PR fixes a significant regression in `redux-form` v8.1.0 (and above), where the following error is thrown in the browser on any page using `redux-form`:

```
Uncaught ReferenceError: module is not defined
```

I fixed it by ensuring that the `module` reference is guarded by `typeof`.

Note that I also updated package.json to use an npm `prepare` script instead of `prepublishOnly`, because that allows me to use my GitHub fork as a dependency in my app. I think it would be a good idea to use `prepare`, but I'm happy to revert that commit if you'd prefer, as it's unrelated to the bugfix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "build:umd:min": "cross-env NODE_ENV=production webpack src/index.js -o dist/redux-form.min.js",
     "clean": "rimraf dist lib es",
     "lint": "eslint src",
-    "prepublishOnly": "npm run lint && npm run test:cov && npm run test:flow && npm run clean && npm run build",
+    "prepare": "npm run lint && npm run test:cov && npm run test:flow && npm run clean && npm run build",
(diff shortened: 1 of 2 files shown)


shyamkumaryadav:
#4412

onur-saf:
when will you publish it on npm ?

shyamkumaryadav:
https://github.com/shyamkumaryadav/redux-form/pkgs/npm/redux-form if you need urgent try this

enrique-ramirez:
Hi @erikras. do you think we could release this change to the npm registry? I tried installing @shyamkumaryadav version to no avail, and this is the one thing preventing me to migrate my app to Vite instead of Webpack.

nphmuller:
Also pushed a fork with the fix to npm, as a temporary workaround: https://www.npmjs.com/package/@nph.muller/redux-form

erikras:

Issue #4642, opened 2020-03-31
initialValues available after update


motoedie:
### Subject of the issue

Form is rendered twice and initialValues are rendered with update. Validation is started immediately without initialValues.

Form is rerendered when initialValues are saved to redux. It is unnecessary update. Calling validate() function without initialValues isn't correct neither. This didn't happen in version 7.4.2 because initialValues were available immediately.

### Your environment

- **OS**: macOS 10.15.3
- **Packages**: redux-form 8.3.0, react 16.12.0, redux 3.6.0, react-redux 7.2.0
- **Env**: Chrome 80

### Steps to reproduce

1. Go to https://codesandbox.io/s/redux-form-simple-example-94gdw
2. App crashed in validate() function because initialValues aren't available yet.
3. If you check `initialized` prop, application works https://codesandbox.io/s/redux-form-simple-example-31jpk

### Expected behaviour

initialValues should be available in first render and also in validate() function.

### Actual behaviour

initialValues are available after component update. validate() function is called without initialValues.

erikras:
Published fix in [`v8.3.10`](https://github.com/redux-form/redux-form/releases/tag/v8.3.10).

Issue #4628, opened 2020-03-07
Form failing required validation even when set with initialValues


joshuaohana:
### Subject of the issue

I have a React Redux Form which has a few required field validators and has its initial values set with initialValues prop

The values initialize fine, but when I try to save the form it errors saying the field is required (even though there's a value in there). If I simple just CLICK into the field then save again everything works fine!

I have tried every way I can find using initialize/reset/destroy/change/blur/etc to manually touch or set the field all to no avail

```
reduxForm({
    form: 'editCase',
    touchOnChange: true,
    touchOnBlur: true
}),

useEffect(() => {
    if (initialValues && initialValues.treatmentDate) {
        // changeFormValue('treatmentDate', initialValues.treatmentDate)
        blurFormValue('treatmentDate', '20001111')
    }
}, [initialValues])

```
and a whole slew of different options as above

### Your environment

- **OS**: macOS 10.14.4
- **Packages**: React 16.8.6 React-Redux 7.1.0 Redux 4.0.4 Redux-Form 8.2.4
- **Env**: NPM 12.13.0 Chrome 80

### Expected behaviour

When values get initialized into required fields, the fields should be marked as dirty and allow submission.

### Actual behaviour

Form is invalid for required field even though there is a valid value in that field.

erikras:
Published fix in [`v8.3.10`](https://github.com/redux-form/redux-form/releases/tag/v8.3.10).

EvgenyArtemov:
@erikras how it was eventually solved? I faced with the same issue, updated to version with fix and it didn't help.

erikras:
Ok. Reopened.

PR #4728, opened 2021-04-29
fix: initialValues missing in validation issue #4628 #4642

shanshanzhu:
During intialization, the value is passed to validation function even if intialvalues are set. (#4628 #4642)
Without the fix, the `props.validate` function is called three times, twice with the `props.value`, and the third time with the `initialValues`, this creates 2 problems:
1 - unnecessary call to validate function (less performant)
2 - when we upgrade to v8 from v7 / v6, the validate function in previous version is invoked with `initialValues`, hence a lot of code is breaking b/c the assumption that `initialValues` are present for the `validate` function on component load time.

Another thing this diff does is to remove the `UNSAFE_componentWillMount`, which will be deprecated in future React versions. This is recommended on react doc https://reactjs.org/docs/react-component.html#unsafe_componentwillmount
`we recommend using the constructor() instead for initializing state.`

In this fix, instead of validating at two places on mount, here we need to validate once after component is fully mounted and when there is not another redux life cycle event updating the value from `initIfNeeded`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3389,12 +3389,6 @@
       "integrity": "sha1-iYUI2iIm84DfkEcoRWhJwVAaSw0=",
       "dev": true
     },
-    "asap": {
-      "version": "2.0.6",
-      "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
-      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=",
-      "dev": true
-    },
     "asn1": {
       "version": "0.2.4",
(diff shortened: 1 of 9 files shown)


shanshanzhu:
Is this project still accepting PR for bug fixes? I cannot set assignee or reviewers to the PR so ping your guys for attention.
@erikras @iamandrewluca

shanshanzhu:
@iamandrewluca Thanks for adding as reviewer.
I updated the description with why I add this fix, and let me know if there is any question.

vladkosinov:
@erikras @iamandrewluca is the a chance this PR will be merge?

shanshanzhu:
@iamandrewluca  Any update on this PR?

erikras:

PR #4722, opened 2021-04-08
Fix peer dependency for react 17.0.x

mariuz:
Related tweet 

https://twitter.com/wesbos/status/1359173446740172810

Fixes Could not resolve dependency:
peer react@"^16.4.2" from redux-form@8.3.7

https://github.com/redux-form/redux-form/issues/4721

and 

https://github.com/redux-form/redux-form/issues/4710

PR #4749, opened 2021-10-14
8.3.8

Blaapje:
In order to take PR #4722 live, we need to bump the version number.


HarshaDW:
Do we have an update? Blocks from upgrading to react 17

erikras:
Released in [`v8.3.8`](https://github.com/redux-form/redux-form/releases/tag/v8.3.8).

PR #4762, opened 2022-05-04
Fix peer dependencies to include future versions

lynxtaa:
Fixes problems with NPM peer dependencies, including future versions since this package is unmaintained.

Fixes NPM errors when installing react@18 and react-redux@8

Fixes https://github.com/redux-form/redux-form/issues/4763

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,10 +105,10 @@
     "webpack-cli": "^3.3.11"
   },
   "peerDependencies": {
-    "react": "^16.4.2 || ^17.0.0",
-    "react-redux": "^6.0.1 || ^7.0.0",
-    "redux": "^3.7.2 || ^4.0.0",
-    "immutable": "^3.8.2 || ^4.0.0"
+    "react": ">=16.4.2",
+    "react-redux": ">=6.0.1",
+    "redux": ">=4.0.0",
+    "immutable": ">=3.8.2"
   },
   "peerDependenciesMeta": {
     "immutable": {


lynxtaa:
@erikras please merge this PR and release a new version 🙏

OmerHerera:
👋 Any idea when this will be merged to the main branch and release 
Thanks

Themonkey180:
Can we get this merged?
The build issue was only in stable node.
node: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by node)

made-in-nz:
It's pretty clear this library is no longer maintained.  Even it's successor library, final-form/react-final-form is no longer actively maintained.  No disrespect to @erikras, I don't think anyone could keep up with the volume of noise these popular libraries generate!  IMO this library is better than final-form & others such as formik & react-hook-form so it's a pity.

ValeriyaKaretnikova:
Would be great if this PR is merged! Please 🙏

hvician:
I wonder when will this be fixed, based on this tread [travis-ci.community](https://travis-ci.community/t/the-command-npm-config-set-spin-false-failed-and-exited-with-1-during/12909/6) the version/distribution of Ubuntu Linux that Travis CI runs on needs to be upgraded,` { dist: focal } `prop needs to be changed and no changes to travis.yml are required so this is more of a dev ops issue. I wish I had the privs to do the upgrade. The final-form or Formik libraries will never be a replacement of a redux-form as bigger apps need to have access to the form state and form API from various components even if the form component is not mounted.

jmgarciasanti:
@erikras please merge this PR because React 17 dependency prevents the evolution to expo SDK 46 which is using React 18
Thanks!

sbaechler:
But does Redux Form work with React 18? I guess it won't support concurrent mode.

jmgarciasanti:
@sbaechler in my case I'm "using" React 18 due to Expo upgrade to SDK 46, but I'm not using React 18 new features like concurrent mode. Maybe dependencies can be upgraded in reduc-foem to include React 18, with a disclaimer of not supporting concurrent mode at this time.

lynxtaa:
I guess it's not relevant anymore since https://github.com/redux-form/redux-form/pull/4771 is merged

Issue #4759, opened 2022-03-12
error UNSAFE_componentWillReceiveProps and UNSAFE_componentWillMount


tranlammedia:
export default reduxForm({form: FORM_NAME})(withStyles(styles)(TaskForm))

--> error Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code. See https://reactjs.org/link/unsafe-component-lifecycles for details.

PR #4768, opened 2022-07-14
Fix/4759/replace unsafe methods

KostiantynPopovych:
Closes #GH-4759

Referring to [Strict Mode](https://reactjs.org/docs/strict-mode.html) - remove UNSAFE_componentWillMount & UNSAFE_componentWillReceiveProps,  also update some devDependencies - react, react-dom, react-test-renderer 18.2+ versions.


src/ConnectedFields.js
```diff
@@ -46,7 +53,12 @@ export default function createConnectedFields(structure: Structure<any, any>) {
         this.onBlurFns[name] = event => this.handleBlur(name, event)
       })
 
-    UNSAFE_componentWillReceiveProps(nextProps: Props) {
+    static getDerivedStateFromProps(nextProps: Props, state: State) {
+      state.prepareEventHandlerIfNecessary(nextProps)
+      return null
+    }
+
+    prepareEventHandlerIfNecessary(nextProps: Props) {
```
sbaechler, line 61: What is the reason for putting this method into the component state? 

And what is `getDerivedStateFromProps` doing if it always returns null?

```suggestion
    prepareEventHandlerIfNecessary = (nextProps: Props) => {
```

KostiantynPopovych: Hey @sbaechler, I'm so sorry for such a long reply.

- the reason for putting methods into the state object is that using this approach, we will be able to call methods like this in the `getDerivedStateFromProps` method (it is `static`, so we don't have access to `this`);
- in this implementation, `getDerivedStateFromProps` only calls the methods in the proper phase. Returning `null` from this method is required when we don't want to make any changes to the `state` object;

I have upgraded my initial solution by adding a check if we need to perform the operation in the `getDerivedStateFromProps` method. Previously, the `componentWillReceiveProps` has been called only during the `updating` phase, but the `getDerivedStateFromProps` is called during the `mounting` and `updating` phases, so we don't really want to call our method inside the `getDerivedStateFromProps` during the `mounting` phase because it will change the previous working behavior (even if the tests are passing).

[The diagram](https://qph.cf2.quoracdn.net/main-qimg-bb3d4526cda108b56ee47ee806fbd8cc) shows the methods execution flow.

KostiantynPopovych:
Closing due to side-effects in the `getDerivedStateFromProps` method. `UNSAFE_componentWillReceiveProps` should be replaced with `componentDidUpdate`, but it also will affect a lot of logic in other class methods.

Issue #4774, opened 2022-12-06
Support react-redux 8.x.x


GFGill:
### Subject of the feature

react-redux current version is 8.0.6, but the peerDependency does not  allow it

### Problem

current peerDependency only allows:
"react-redux": "^6.0.1 || ^7.0.0",

### Expected behavior

Suggesting an update to the peerDependency so that redux-form can work with the current version of react-redux

elartix:
Npm package still contain
`"react-redux": "^6.0.1 || ^7.0.0",`

[https://www.npmjs.com/package/redux-form/v/8.3.9?activeTab=explore](https://www.npmjs.com/package/redux-form/v/8.3.9?activeTab=explore)

Aerendir:
Any news on this? The change was merged, but the version wasn't released...

Aerendir:
It seems this package is abandoned.

On the home page of its documentation, there is a clear statement saying:

> Attention: Do not begin a project with Redux Form. The author of Redux Form has written a superior library, [React Final Form](https://final-form.org/react/). If you have already started using Redux Form, there is a [migration guide](https://final-form.org/docs/react-final-form/migration/redux-form).

<img width="915" alt="Screenshot 2023-03-03 alle 18 10 32" src="https://user-images.githubusercontent.com/1968622/222783867-9ff404b2-fe63-49b4-b84d-19242e1516b3.png">

Ref.: https://redux-form.com/8.3.0/

So, it seems we have to move to React Final Form.

yanndinendal:
This seems to be fixed in `redux-form@8.3.10` ([release notes](https://github.com/redux-form/redux-form/releases)), which now allows `react-redux@^6.0.1 || ^7.0.0 || ^8.0.0`, but since then a new major `react-redux@9` was released and is not allowed by `react-redux`'s peerDeps.

vicodinvic1:
ok, now we need support for v9 please :)

PR #4775, opened 2022-12-06
Update package.json

GFGill:
* react-redux 8.0 Support

Updated peer dependency for react-redux, to work with 8.0 version.


GFGill:
Fixes issue #4774

maguileran:
It seems to be a problem between Node 19 which is the stable version but not well-supported by the linux dist (xenial)

Maybe you need to declare a specific version, node 18 for example or change the linux dist to one that supports node 19

erikras:

